PR #126, opened 2014-11-04
Enable Travis CI to test specs against Ruby Sass

nschonni:
Note this shows that some of the specs currently fail on the Ruby Sass stable.
Eventually it would probably be good to create a build matrix that would run on the various implementations.


xzyfer:
This has since been addressed. Thanks.

PR #218, opened 2015-01-02
add test for escaped #{}

smikes:
The contents of `expected_output.css` were generated with sass @ 3.4.9 ; I believe this behavior has not changed in the sass 3.0 line but I have not verified this.

In a string, the sequence `#{` introduces an interpolation except when it is escaped with backslash: `\#{`.  Ruby sass implements this, but libsass does not.  See https://github.com/sass/libsass/pull/798 for bugfix to libsass.


xzyfer:
Hi @smikes thanks for this. Since this feature is currently unimplemented in Libsass can you please move this file into the `libsass-todo-tests` folder?

smikes:
@xzyfer Sure, but when the https://github.com/sass/libsass/pull/798 is landed it will need to be moved out of TODO so is it better to just hold this until the other goes, or do you want me to make the move?

xzyfer:
That's ok, please move it. 

It's good practice to always make todo tests. It's hard to predict when we'll be able to merge  sass/libsass#798.

smikes:
will do

smikes:
OK, tests moved to libsass-todo-tests.

Bonus to doing this is I can add two additional tests, which fail even with the PR noted above.  Naively looking for a preceding backslash is actually wrong (even when implemented correctly) because the *actual* requirement for the `#{}` to be unescaped is that it is preceded by an **even** number of backslashes; an **odd** number of backslashes is escaped.

smikes:
Closing as irrelevant, since related PR #798 was closed.

PR #234, opened 2015-01-10
Speed up the Ruby Sass tests by running them in-process.

chriseppstein:
I know Ruby Sass is slow, but there's no need to make using it painful. This makes the Sass spec tests run in about 5 seconds on my machine instead of taking more than 4 minutes. This is accomplished by running the sass spec tests in the sass-spec process instead of starting a new ruby process for each test case.

This refactor also makes it easier to support new engines in the future by providing a clean object-oriented interface instead of making assumptions about the CLI that is being invoked.

### DIFF
--- a/lib/sass_spec/cli.rb
+++ b/lib/sass_spec/cli.rb
@@ -1,10 +1,11 @@
+require_relative 'engine_adapter'
 
 module SassSpec::CLI
   require 'optparse'
 
   def self.parse
     options = {
-      sass_executable: "sass",
+      engine_adapter: SassEngineAdapter.new("sass"),
       spec_directory: "spec",
       tap: false,
       skip: false,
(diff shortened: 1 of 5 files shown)


chriseppstein:
@xzyfer @akhleung can I merge?

akhleung:
If you've tested it, I have no objections.

xzyfer:
As long as the `--unexpected` flag still works as expected :+1: from m me.

chriseppstein:
What does `--unexpected` do?

xzyfer:
It runs all tests including todos and notifies you if a todo test now passes.

xzyfer:
We use it to determine if a patch has unexpectedly caused previously failing specs to pass. It's probably only useful for Libsass(?). I've just had a look at the PR and it doesn't appear to affect it though.

PR #244, opened 2015-01-12
Added spec for LibSass issue #613

HugoGiraudel:
Added:

* each
* for
* function
* if
* mixin
* nested
* root
* rulset
* while

Hope it's enough.

### DIFF
--- a/spec/scope/clash/expected_output.css
+++ b/spec/scope/clash/expected_output.css
@@ -1,0 +1,10 @@
+.foo {
+  content: 42;
+  content: 43;
+  content: 45;
+  content: 42;
+  content: 4;
+  content: 45;
+  content: 3;
+  content: 5;
+  content: 45; }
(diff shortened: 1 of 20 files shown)


xzyfer:
Thanks for this @HugoGiraudel . The one bug thing missing from all your cases is testing that variables in different scopes don't clash (which is the current missing feature) i.e.

```scss
$foo: 42;
$bar: 43;
$baz: 45;

foo {
  foo: $foo;
  bar: $bar;
  baz: $baz;
  
  $bar: 4; // this is a different $bar than `$bar !global`

  foo: $foo;
  bar: $bar;
  baz: $baz;

  @if true {
    $foo: 3; // this is a different $foo than `$foo !global`
    $bar: 5; // this is a different $bar than `$bar !global`

    foo: $foo;
    bar: $bar;
    baz: $baz;
  }
}
```

HugoGiraudel:
Added.

mgreter:
Hey @HugoGiraudel I have added your commits in here into https://github.com/sass/sass-spec/pull/282 and regenerated the expected output files to include the new output styles. IMO this PR can be closed and the solution should be merged soon in libsass: https://github.com/sass/libsass/pull/986. Maybe you want to give it a try? Thanks!

PR #249, opened 2015-02-01
Check saturation/desaturation

anlutro:
Ensure that saturation of fully saturated colours or desaturation of fully desaturated colours has no effect.

See https://github.com/sass/libsass/issues/864 https://github.com/sass/libsass/pull/865


xzyfer:
Thanks @anlutro

PR #250, opened 2015-02-02
Add more saturation/desaturation tests

anlutro:
Included some not-so-edge cases. Expected values checked with ruby sass 3.4 and 3.2. Related to https://github.com/sass/libsass/pull/865

Note the odd behaviour of `saturate()` on grey colours.


xzyfer:
Thank again!

PR #262, opened 2015-02-16
Feature/selectors

Snugug:
Selector tests for various ways `&` can be used as requested by @xzyfer in sass/libsass#548. Does not add tests for the [String Functions](http://sass-lang.com/documentation/Sass/Script/Functions.html#selector_functions) but could if desired. Reading the README, it appears as if these tests should go under the `libsass-open-issues` but I don't see that folder and thought that, considering there are lots of things that can be done with selectors, having a selector folder wasn't so bad. If you'd like me to move it I'm happy to.


xzyfer:
Thanks heaps for this @Snugug!

The correct place for these specs would be https://github.com/sass/sass-spec/tree/master/spec/libsass-todo-tests, this marks the tests as todos and won't be run as part of our usual development and CI. The structure you have is great, just move it into `spec/libsass-todo-tests`.

The selector specs are really useful. The cases variable assignments and function/mixin arguments really interest me as we currently have _no_ coverage for them.

xzyfer:
The major suggestion I would make is to please drop the `--` and `__` in favour of just creating a sub directory.

Snugug:
I can do both of those things; wasn't sure if nested folders were cool. I'll do both

xzyfer:
Great! Thanks so much!

Snugug:
Done.

xzyfer:
Perfect @Snugug thanks heaps!

PR #267, opened 2015-02-22
[WIP] Add spec tests for interpolate refactoring

mgreter:
I created a lot of spec tests for my interpolation refactoring work.Most of them are auto generated, but they helped me big time to get the implementation right with ruby sass. IMO it adds a few cases which have not been handled before (like escape sequences etc). @xzyfer what do you think?

https://github.com/mgreter/libsass/commits/refactor/output-preparation


mgreter:
Closing in favor of https://github.com/sass/sass-spec/pull/270

PR #268, opened 2015-02-24
add spec for libsass issue 893

callmecavs:
adds a spec for sass/libsass#893


callmecavs:
@mgreter was this merged in error? my issue was closed as not reproducible, the expected output in this pull request is incorrect, it should be (according to Sassmeister):

```css
.row {
  margin: $-20px; }
```

mgreter:
Never mind, already fixed it with https://github.com/sass/sass-spec/commit/3ffcc64b1c658e425fb5c9e97385d174a250e52c !

callmecavs:
:+1:

Issue #247, opened 2015-01-19
Support the indented syntax


xzyfer:
Looks like Libsass has issues with indented syntax that aren't in the scss syntax https://github.com/sass/libsass/issues/845. As is stands we follow the normal procedure of writing a spec because sass-spec.rb can only handle one format per run.

This is likely related to https://github.com/sass/sass-spec/issues/82, and https://github.com/sass/sass-spec/issues/46.

/cc @chriseppstein

saper:
Closed via #503.

Issue #82, opened 2014-10-10
Test organization?


mirisuzanne:
@xzyfer mentioned the need for this on sass/libsass#521 — and I thought it deserved an issue over here. The current organization of sass-spec tests is quite difficult to sort through, based partially on what feature is being tested and partially on where support already exists.

I think the only organization that makes sense long-term is feature-based. Is there something stopping us from moving completely in that direction? Is there anything we can use as a "full list of features" to work against? Any inherent structure in the code, or somewhere in the RubySass project?

chriseppstein:
I volunteer to reorganize these tests into features much like how @HugoGiraudel suggested. Some tests may need to be re-written, etc. The goal here is to have a clear approach to see how completely a engine implements a particular feature set and also going forward to know which tests are for new features that are being added to Sass.

Do you guys want me to go crazy or do you want to approve the work etc?

xzyfer:
I think with a mix of feature and language (http://apps.workflower.fi/vocabs/css/en) based naming we can really clean up the specs. 

However my feeling atm is that the current state of sass-spec makes having an educated and productive discussion on the organisation of specs, and how sass spec fits into our workflow virtually impossible.

> Some tests may need to be re-written

This will almost certainly be the case. Many specs are snippets of code from the project that had the issue and aren't granular assertions of functionality.

> The goal here is to have a clear approach to see how completely a engine implements a particular feature set and also going forward to know which tests are for new features that are being added to Sass.

IMHO this plays into https://github.com/sass/sass/issues/1466. Ideally, for Libsass, new Sass features begin with a sass spec PR as a minimum viable RFC.

> Do you guys want me to just dig in or do you want to approve the work first?

Please feel free to tackle this on a branch and PR the results with a TLDR of the spec layout, and the workflow you envision.

Since we're currently working rapidly on 3.2.0, and 3.1.0 is about to hit the masses via node-sass@2.0.0 having sass-spec changing under our feet (read: on master) is going to be problematic.

HugoGiraudel:
As I said to @chriseppstein last night, I am fully okay to help with the new test organisation. I would probably suggest we tackle things this way:

1. Move all tests to a `spec/todo/` folder or whatever.
1. For each test, try to find it a generic parent suite collection (`strings`, `mixins`, `variables`, `scope`, `at-root`...) then try to find it a meaningful name.
[...]

PR #270, opened 2015-03-01
Test all output styles

mgreter:
Related to https://github.com/sass/libsass/pull/910 and https://github.com/sass/sassc/pull/97

- Adds expected files for all output styles
- Changes test runner to test all output styles
- Added flag files to skip or clean certain tests


xzyfer:
Could you please update the `nuke` code to produce all these new spec files?

mgreter:
By the way, quite a few todo tests emit this warning in ruby sass:
```
The subject selector operator "!" is deprecated and will be removed in a future release.
This operator has been replaced by ":has()" in the CSS spec.
```

I wonder if we want to update those tests and if we even want to implement the deprecated syntax?

xzyfer:
We should continue to support those specs until Ruby Sass starts to throw an error. Other wise we're not a drop in replacement.

xzyfer:
Also fixes https://github.com/sass/sass-spec/issues/247

mgreter:
Added PR to my own branch for https://github.com/sass/libsass/issues/922
https://github.com/mgreter/sass-spec/pull/1

mgreter:
IMO we should get a decision if we want to merge this PR before releasing libsass 3.2!
Since it is a pretty big change I would like to have some feedback before merging it on my own!
//CC @chriseppstein @hcatlin @akhleung @xzyfer @am11

xzyfer:
IMO this is definitely worth doing, especially if all these specs are currently passing on 3.2!

As long the utility methods (nuke, unexpected-pass) function correctly then this gets a :+1:  from me.

The one possible concern could the significantly increased running time of Ruby Sass if they were to switch to sass-spec sometime in the future?

It's worth noting it's likely that sass-spec will undergo significant changes with @chriseppstein's work as per https://github.com/sass/sass-spec/issues/82#issuecomment-69998072.

am11:
:+1:
@mgreter, (at least on http://sassmeister.com/) it seems like ruby sass has default output style set to expanded. Should we do the same here (and then all the down-streams will follow)? 

This (breaking) change will be the perfect candidate for forthcoming major version of node-sass.

//cc @jedfoster.

mgreter:
IMO the nuke option etc. should still work, since it only adds more tests with different output options and expected output path. But I actually never did any ruby before these changes to the spec runner. So I really would like to get a O.K. from a someone that knows ruby.

The running time to run the spec test might be an issue, but IMO it's worth it. My perl-libsass spec runner needs 15s-20s to do all 4044 checks (including todo test). Regenerating all 4044 spec test via ruby sass takes about 3 minutes, while exhausting all my [4 cores] [1] (i7-3770).

It will probably also increase the coverage by one or two percent (I would guess)!

[1]: https://github.com/sass/perl-libsass/blob/latest/t/99_sass_specs.t#L149

PR #272, opened 2015-03-10
Add specs for libsass issue 930

anlutro:
Hopefully the directory structure here is OK. Since the issue I submitted has two separate failing cases, I named the spec directories `issue_930_case1` and `issue_930_case2`. Let me know if this should be made any different.

Libsass issue: https://github.com/sass/libsass/issues/930


xzyfer:
Thanks for this @anlutro. The directory structure is a bit off though. You can use this as a reference _(replacing libsass-closed-issues with libsass-todo-issues)_

https://github.com/sass/sass-spec/blob/master/spec/libsass-closed-issues/issue_185/mixin/input.scss

Key points:
- /issue_<number>/
- use well name sub dirs if additional cases are required

anlutro:
Pushed an amended commit. Does that look OK?

xzyfer:
Thanks, looks good. It'd also be good to add a simple control case

```scss
.foo {
  &.bar {
    color: #F00;
  }
}
```

anlutro:
Added that and another similar test case where `.bar` is done via string interpolation.

xzyfer:
Perfect!

PR #278, opened 2015-03-16
Add tests for libsass issue 942

anlutro:
https://github.com/sass/libsass/issues/942

Note that there's a big difference between Sass 3.2 and 3.4 behaviour in this case. 3.2 will error on the multi-line string variable, and incorrectly compile the content into `"foo\ bar"`.


xzyfer:
Thanks @anlutro

PR #279, opened 2015-03-16
Add specs for issue 947.

mechanicalduck:
Add specs for issue 947.

### DIFF
--- a/spec/libsass-todo-issues/issue_947/expected.compact.css
+++ b/spec/libsass-todo-issues/issue_947/expected.compact.css
@@ -1,0 +1,2 @@
+@keyframes test { 10% { color: red; } }
+
--- a/spec/libsass-todo-issues/issue_947/expected.compressed.css
+++ b/spec/libsass-todo-issues/issue_947/expected.compressed.css
@@ -1,0 +1,2 @@
+@keyframes test{10%{color:red}}
+
--- a/spec/libsass-todo-issues/issue_947/expected.expanded.css
+++ b/spec/libsass-todo-issues/issue_947/expected.expanded.css
@@ -1,0 +1,6 @@
+@keyframes test {
+  10% {
+    color: red;
+  }
+}
+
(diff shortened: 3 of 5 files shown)


xzyfer:
Thanks @mechanicalduck

spec/libsass-todo-issues/issue_947/expected.compact.css
```diff
@@ -0,0 +1,2 @@
+@keyframes test { 10% { color: red; } }
+
```
am11: @xzyfer, would those extra line feeds at the end of file cause issue with test scripts? Normally, if it is CRLF then we add extra line feed so git squashes CRLF+CRLF+EOF (or LF+CRLF+EOF) to a single line feed (LF or CRLF) followed by EOF. If we don't have double CRLF, then git diff on GitHub shows a little red abnormal line termination marker (like [this](https://github.com/am11/node-sass/commit/ae70faeeb710f9a2a120f25de94e8c12d8eed414#diff-67dc932082c9767347dbd766ad547b5bR10)). But in case of double LF, git will retain double linefeeds. IMO, we should avoid this from happening.

xzyfer: I think this is fine. I force my editor to add a blank line at the end of every file. A lot of specs have this. Also the specs pass on Ruby Sass so I assume all is well.

```
➜  sass-spec git:(master) ruby sass-spec.rb -c sass -s spec/libsass-todo-issues/issue_963
Recursively searching under directory 'spec/libsass-todo-issues/issue_963' for test files to test 'sass' with.
Sass 3.4.9 (Selective Steve)
Run options: --seed 65106

# Running:

....................................................................................................................................................................

Finished in 57.447308s, 2.8548 runs/s, 8.5644 assertions/s.

164 runs, 492 assertions, 0 failures, 0 errors, 0 skips
```

am11: Yup we have `.trim()` in node-sass spec runner too. Would be nice if we enforce one rule for all specs. :sunglasses:

xzyfer: I guess the trick is getting our of people's way as much as possible. Trying to manage newlines, indentation etc is barrier to people contributing. Especially when we can automatically fix styling ourselves with some simple tooling. I'm fine to just yolo it for now until we're at parity with Ruby Sass.

PR #291, opened 2015-03-30
Activate passing spec tests

mgreter:
Activates issue tests
- libsass-closed-issues/issue_947
- libsass-closed-issues/issue_950
- libsass-closed-issues/issue_1007
- spec/extend-tests/090_test_comma_extendee
- extend-tests\091_test_redundant_selector_elimination

Activates passing is_superselector function tests

- has_isnt_superselector_of_contained_selector
- host_context_isnt_superselector_of_contained_selector
- host_isnt_superselector_of_contained_selector

Removed duplicate spec test
- extends\190_test_semi_placeholder_selector


xzyfer:
I'd argue against removing specs from the `extends` folder. For a couple reasons:

- it's good for use to group this functionality
- it's good to group specs by feature sets (which we're now doing more of yay!)
- (I believe) these are ported over from Ruby Sass so it'd be good not to diverge for the time being (IMO)

xzyfer:
I would instead opt for removing the test that duplicates `extends\190_test_semi_placeholder_selector`.

mgreter:
That test is the only test in folder `extends`, the other ones live in `extend-tests` and there already exists an exact copy of `190_test_semi_placeholder_selector` ...

xzyfer:
Ok cool! :+1:

PR #303, opened 2015-04-05
Add first indented syntax test

mgreter:
We actually can test `sass` files by using a simple import in `input.scss`.


xzyfer:
Do this not mean we cannot test both `input.scss` and `input.sass` in the same spec?
Seems like the better approach would be to just update the runner to handle `.sass` files.

PR #325, opened 2015-04-11
Add specs for variable concatenation

ausi:
I’m not sure if this is the right place for this. I found differences between RubySass and LibSass regarding concatenation of variables. They behave differently and both not right IMO.

So RubySass and LibSass currently don’t pass the specs I added in this pull request, should I rename *input.scss* to *input.disabled.scss* because of this?

Should I create separate issues describing the failing parts in the RubySass and LibSass repositories?


xzyfer:
Thanks @ausi. 

>So RubySass and LibSass currently don’t pass the specs I added in this pull request

Ruby Sass is the reference implementation and thus however it behaves is the _correct_ way.
Libsass is intended to be a drop in replacement for Ruby Sass and as such will do as it does.

>I’m not sure if this is the right place for this

This repo exists to assert the behaviour of Ruby Sass.

If you believe Ruby Sass is behaving incorrectly please open an issue there https://github.com/sass/sass/issue.

ausi:
I updated [my branch](https://github.com/ausi/sass-spec/tree/spec/variable-concatenation) to match the correct RubySass behavior according to https://github.com/sass/sass/issues/1704#issuecomment-94626433 and created the issue sass/libsass#1158.

Should I now create a new pull request with the corrected tests?

PR #326, opened 2015-04-14
Remove invalid specs for non-standard compact function

xzyfer:
This PR removes or updates specs that incorrect reference the non-standard `compact` function.


hcatlin:
This seems reasonable to merge in now.

mgreter:
IMO that probably means they will pass now, so no longer todo IMO :wink: But should be tested first.

xzyfer:
You're probably right haha

PR #327, opened 2015-04-14
Add sass-spec for issue 1098

atdrago:
This PR adds specs for sass/libsass#1098

### DIFF
--- a/spec/libsass-todo-issues/issue_1098/expected.compact.css
+++ b/spec/libsass-todo-issues/issue_1098/expected.compact.css
@@ -1,0 +1,1 @@
+div { opacity: 1\9; }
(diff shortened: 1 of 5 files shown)


xzyfer:
Thanks @atdrago. I've noticed the indentation looks really off on the expanded and nested output. Could you please take a look?

The best to generated the expected output is to use the following command in the sass-spec directory

```
bundle install
ruby sass-spec.rb -c sass spec/libsass-todo-issues/issue_1098 --nuke
```

xzyfer:
I've found some extra erroneous cases. I've created a https://github.com/sass/sass-spec/pull/331 to cover them.

PR #331, opened 2015-04-16
Add specs for issue 1098

xzyfer:
This PR adds specs for regressions in the IE property hack https://github.com/sass/libsass/issues/1098.

### DIFF
--- a/spec/libsass-todo-issues/issue_1098/expected.expanded.css
+++ b/spec/libsass-todo-issues/issue_1098/expected.expanded.css
@@ -1,0 +1,5 @@
+div {
+  opacity: 1\9;
+  width: 500px\9;
+  color: #f00\9\0\;;
+}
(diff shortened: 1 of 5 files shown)


spec/libsass-todo-issues/issue_1098/expected.expanded.css
```diff
@@ -0,0 +1,5 @@
+div {
+  opacity: 1\9;
+  width: 500px\9;
+  color: #f00\9\0\;;
```
xzyfer: I believe this double `;;` may be an issue with Sass. /cc @mgreter @chriseppstein

mgreter: IMO that looks correct as one `;` is escaped, as in the input. IMO the delimiter is optional for the last declaration!

xzyfer: You're correct. My mistake.

PR #363, opened 2015-05-04
Activate passing spec tests

mgreter:
- Activate test specs for new math unit operations (https://github.com/sass/libsass/pull/1049)
- Remove duplicated todo test (already activated but copy still in todo folder)


xzyfer:
> Remove duplicated todo test (already activated but copy still in todo folder)

This may be related to @hcatlin's recent merge.

xzyfer:
@hcatlin since there's a decent amount of activity here now is it worth you working of a fork and PR rather than directly on master?

hcatlin:
In my defense, that was actually truly human error that any of that got merged in at all. It was on a local branch locally, but my use of a new workflow started merging things in that were not intentional. Hence, the original revert attempts. I didn't want to -f push and instead tried to repair, badly.

xzyfer:
I wasn't having a dig at you, apologies if it came across that way. We all
know mistakes happen.

I simply meant it would be good for @mgreter and myself to be aware of
changes to sass-spec and PRs are the best way to achieve that.

This is especially true for when things change or break, like the recent
confusion around unexpected pass and multiple output modes.

mgreter:
Superseded by https://github.com/sass/sass-spec/pull/367

PR #373, opened 2015-05-10
Add spec test for libsass issue 1187

mgreter:
- https://github.com/sass/libsass/issues/1187


xzyfer:
Tracking error message mismatch in https://github.com/sass/libsass/issues/1561

Issue #361, opened 2015-05-01
Broken Builds ;(


hcatlin:
Apparently we have a lot of broken and builds against Sass 3.4.13

https://travis-ci.org/sass/sass-spec/builds/60883789

Can I nuke and fix these?

@mgreter it seems like your test of using non-utf-8 characters broke Ruby Sass... spec/libsass/length

mgreter:
Are you telling ruby sass to use utf8 encoding? IMHO I had to use `sass -E UTF-8` to have it behave correctly, but thought that was a windows only issue ...

xzyfer:
I've looked into this further. Aside the form UTF-8 spec mentioned above, there are two failing test.

**spec/libsass/scale**

Looks to be a rounding error with colours.

```
-  color: #c85ae6;
+  color: #c85ae5;
```

**spec/libsass-closed-issues/issue_614**

Regarding `@support` output. This has been acknowledged as a bug in Sass https://github.com/sass/sass/issues/1512#issuecomment-95376786.

mgreter:
For the rounding error see https://github.com/sass/sass/issues/1690

xzyfer:
Using `-E` doesn't fix this on OSX

```
$ sass -E UTF-8 spec/libsass/length/input.scss
Error: Invalid UTF-8 character "\xE9"
        on line 7 of spec/libsass/length/input.scss
  Use --trace for backtrace.
```

mgreter:
Can you confirm it works on windows?
Otherwise I could give it another try in a few hours ...

xzyfer:
I think this might be related to some Windows UTF-8 issue. Once I open and saved the file OSX it worked file.

Here's the diff https://github.com/sass/sass-spec/pull/377

xzyfer:
I can't test on Windows atm.

mgreter:
A hex diff would help me most (too see which encoding we actually had) ...

xzyfer:
https://www.diffchecker.com/h6nzbjaq

mgreter:
OK, so the left text is "old-school" ascii "latin-1" encoded. Libsass tries to parse it as utf8, which uses the upper half of the ASCII table for it's encoding (leaving the lower half as is). We could come up with something that first checks if the file not only "looks" like utf8, but is also valid utf8 and fallback to "latin1" ascii table. At least it is clear now what was causing this issue.

xzyfer:
[...]

PR #377, opened 2015-05-11
Fix weird UTF-8 characters in the length spec

xzyfer:
This PR fixes a spec that causes the Ruby Sass process to crash due to weird UTF-8 characters.

Originally mentioned in https://github.com/sass/sass-spec/issues/361.


xzyfer:
The command line diff was more interesting.

screen shot 2015-05-11 at 11 59 11 pm

PR #380, opened 2015-05-13
Respect the --ignore-todo flag

xzyfer:
This PR fixes a big with sass-spec which causes it to falsely report todo specs as passing. This has resulted in some recent Libsass regressions, and incorrectly closed bugs.

/cc @mgreter

### DIFF
--- a/lib/sass_spec/test.rb
+++ b/lib/sass_spec/test.rb
@@ -33,10 +33,10 @@ def run_spec_test(test_case, options = {})
     end
   end
 
-  if !test_case.todo?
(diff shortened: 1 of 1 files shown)


xzyfer:
Ah man this is bad. Looks a decent chunk of the recently closed issues weren't in fact passing. Here are the ones I've identified.

sass/libsass#674
sass/libsass#941
sass/libsass#948
sass/libsass#1043
sass/libsass#1096
sass/libsass#1115
sass/libsass#1162
sass/libsass#1169
sass/libsass#1170
sass/libsass#1171

PR #398, opened 2015-05-28
[RFR] Added spec for LibSass issue 1236

HugoGiraudel:
https://github.com/sass/libsass/issues/1236


xzyfer:
Issue a duplicate of https://github.com/sass/libsass/issues/1210 which already has a spec.

Thanks @HugoGiraudel

PR #399, opened 2015-05-28
Add specs for issue 1248 (@extend specificity problems)

mzgol:
Refs sass/libsass#1248


mzgol:
I fixed the mistake that made Travis fail. This should be ready to land.

PR #403, opened 2015-06-05
Add specs for issue 1210

danielguillan:
PR add specs for sass/libsass#1210


xzyfer:
This is fantastic, thank you @danielguillan

PR #421, opened 2015-06-30
Revert "Add missing todo spec tests for issue 1219"

xzyfer:
Reverts sass/sass-spec#410

https://github.com/sass/libsass/issues/1219 is testing an error condition, as such this spec should not pass by design.


mgreter:
This should again be reverted, the issue #1219 does produce valid output via ruby sass!

PR #434, opened 2015-07-08
Remove unnecessary space after comma in compressed output

mgreter:
Change expected output to match libsass output, which is more correct in this case than ruby sass output (probably/somewhat related to https://github.com/sass/sass/issues/1296, discussed in https://github.com/sass/libsass/pull/1249).


xzyfer:
I've given this some though and I don't think we should proceed down this path. The goal of Sass Spec is to be a portable representation of the Sass language spec. As long as Ruby Sass is the reference implementation Sass Spec should reflect it's _quirks_.

There are two Sass bugs at play here
- extra semicolons in compressed output (https://github.com/sass/sass/issues/1772)
- extra spaces in at-rule directives in compressed output (https://github.com/sass/sass/issues/1296)

I think we should address this as follows.
- patch the semicolon issue upstream
- emulate the extra space issue in libsass 

It doesn't appear Ruby Sass will fix the extra space issue in the near future. The extra semicolon issue on the other hand should be straight forward to fix.

xzyfer:
I have proposed a fix for the extra semi-colon issue in https://github.com/sass/sass/pull/1773

saper:
Sass 3.4.18.4ef8e31 (Selective Steve) has changed that and now passes the modified `spec/libsass-todo-issues/issue_1263` test

PR #467, opened 2015-08-06
add specs for issue 1415

mahtd:
Add specs for sass/libsass#1415


xzyfer:
Thanks for this @mahtd. Could you please update the spec to use this as the input

```
$parent: &;

@if $parent {
  .foo {
    foo: bar;
  }
}
```

This is this minimal code required to reproduce the issue. Having a little code as possible is important as it allows to narrow down the internally issue much faster.

mahtd:
Thanks, @xzyfer, that is much better. I updated to the second test case that you mentioned in the issue

xzyfer:
Thanks for that @mahtd. I've also noticed you haven't generated the expected output. This is required for the test to work. You can take a look at a sample PR in https://github.com/sass/sass-spec/pull/459

xzyfer:
Hi @mahtd I've superseded this PR with https://github.com/sass/sass-spec/pull/468 since we had a patch ready to go. Thanks again for the PR!

PR #469, opened 2015-08-10
Activate specs for issue 1432

xzyfer:
This PR activates specs for sass/libsass#1432


saper:
We have a difference in error messages here:

-"Error: $selector: null is not a string for `simple-selectors'"
+"Error: $selector: null is not a valid selector: it must be a string,"

xzyfer:
Tracking difference in https://github.com/sass/libsass/issues/1563

PR #470, opened 2015-08-12
add specs for issue 1434

mahtd:
Specs for sass/libsass#1434


xzyfer:
@mahtd perfect :ok_hand:

PR #276, opened 2015-03-15
Add spec for issue 945

am11:
This PR adds an error spec for issue described at sass/libsass#945.

PR #472, opened 2015-08-15
Activate specs for issue 1443.

am11:
This PR activates specs for sass/libsass#1443.


xzyfer:
Thanks @am11. Although tests for error conditions don't get merged, they're important.

Don't stress that the build is failing. It'll always fail if an error is expected. https://github.com/sass/sass-spec/pulls?q=is%3Aopen+is%3Apr+label%3Aerror-test

am11:
> Although tests for error conditions don't get merged, they're important.

[Yup I remember](https://github.com/sass/sass-spec/pull/276) :smile:

xzyfer:
Oh sorry, as I briefly alluded to in https://github.com/sass/libsass/issues/1443#issuecomment-130619597. This is not actually an error condition for Ruby Sass. The numbers given to `rgb/rgba` are out of range they are clamped. So `$a:2` is clamped to `$a:1`.

This tests as is stands is invalid.

xzyfer:
The wider issue that we have poor covereage over the colour function. Greater coverage would have caught that we were incorrectly throwing an error here, which in turn causes sass/libsass#1443.

I understand that sass/libsass#1443 is a symptom of a larger issue, however I'm sure how to test it correctly.

am11:
You are right. Just tested with Ruby Sass and it actually considers `$alpha>1` as `$alpha=1`.

Should I prepare another PR in libsass to not throw?

xzyfer:
>Should I prepare another PR in libsass to not throw?

If you've got the time it'd be great. You'd also need to implement the clamping other wise it'd introduce new bugs :)

xzyfer:
I'll set aside some time for building out some thorough colour specs

am11:
that's the way to make weekend more colourful! :)
I'll take a stab at it. 
IMO, if this is the last blocking issue for 3.3 and is a regression, then https://github.com/sass/libsass/pull/1446 can be merged to fix the regression. Color values out of range throwing behavior was there before memory-manager was introduced.

PR #488, opened 2015-09-04
Activate specs for issue 1405

xzyfer:
This PR activates specs for sass/libsass#1405


xzyfer:
I've also update the spec to better match the improvements made in https://github.com/sass/libsass/pull/1471

Issue #136, opened 2014-11-16
Cannot test error reporting


xzyfer:
As it stands it's not possible to assert an error condition. i.e `error on duplicate keys in maps `  https://github.com/sass/libsass/issues/628

There are cases where a spec like the following should be possible https://github.com/sass/sass-spec/pull/133/files.

Maybe we can redirect error output in the assertion some how. Thoughts? 

/cc @lunelson @hcatlin

chriseppstein:
Sass has an output format for errors that results in CSS output. IMO, Errors should be compared using that output. We also need to track warnings. For this, I think there should be a file `output.stderr`. If not present then any stderr output would be a failure, and if present, the stderr output would have to match.

chriseppstein:
I don't think an unexpected output is required. Simply making an output file that is empty should suffice.

HugoGiraudel:
>  Errors should be compared using that output

This.

mgreter:
@chriseppstein I used pretty much the same approach for my local test runner. But IMHO there could a use case for having `output.css` and `output.err` present. Basically it just means to test stdout and stderr and nothing forbids it to output on both (like i.e. emitting warnings). I made both files optional which simply means that I expect an empty stream. There is no difference in having the file empty or not existing (altough output.css could be used to trigger the test in the first place).

am11:
Perhaps it would be a good idea to replace Ruby minitest gem with something like [C++ minitest](http://www.codeproject.com/Articles/9757/Minimal-unit-test-library-for-C) to have more native control over memory and std streams to tests other aspects of libsass, beyond Sass code-generation. This way we can also write independent tests for source-map and external plugins in future. One thing we would probably have to write is the coverage part, for which we can hook native lcov or gcov.

saper:
Before we rewrite a whole thing, I just had a look at the ruby code with my basic understanding of the language and found the following:

* a [test case](https://github.com/sass/sass-spec/blob/6d38353b04e22a044eaab6f6e9bebaeeb5c5f7e4/lib/sass_spec/test_case.rb#L45) reports standard output, cleaned standard output, standard error and compile error and the status code resulting from the compilation
* a [runner](https://github.com/sass/sass-spec/blob/6d38353b04e22a044eaab6f6e9bebaeeb5c5f7e4/lib/sass_spec/runner.rb#L44-L45) just loads few files that describe the expected values:
  * "expected standard output"
  *"expected clean standard output" (optional)
  * applicable output style
[...]

PR #494, opened 2015-09-19
Test for error output

saper:
Enable testing for error output
and sass compiler failures.
 
If files "error" and "status" are existing
in the test directory, do expect errors
and test against them.

Errors returned by the Sass engines are
now treated as test failures, not errors.

Also create error and status files with --nuke.
An error will be reported in this case anyway.

**Important** Now the tests cannot be ran in parallel,
because we need to capture standard error reliably.

Should fix: https://github.com/sass/sass-spec/issues/136 https://github.com/sass/sass-spec/pull/133

### DIFF
--- a/lib/sass_spec/cli.rb
+++ b/lib/sass_spec/cli.rb
@@ -69,7 +69,7 @@ Make sure the command you provide prints to stdout.
       end
 
       opts.on("-s", "--skip", "Skip tests that fail to exit successfully") do
-        options[:skip] = true
+        # Note: --skip is no longer necessary as this is now the default behavior, since we test for errors
       end
 
       opts.on("--nuke", "Write a new expected_output for every test from whichever engine we are using") do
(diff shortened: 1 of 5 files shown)


saper:
As the Travis shows, this change work only with adding compiler via the `-c` option (can be sassc or Ruby sass CLI). Unfortunately, calling Ruby sass internally presents the incompatible interface (this needs to be fixed, too).

saper:
So after resolving #495, #497, #499 the tests finally pass...

saper:
This should be enough for rudimentary error testing. Should we need more, we should be aware of the following limitations:

- [X] status returned from external process is not numeric ( #495 ) - solved via #496 
- [ ] status returned from sass cli, sassc, and Sass ruby module have no consistent values ( #495 ) - worked around via https://github.com/saper/sass-spec/commit/6495e47696399671d0a297bad36cee3fa7615d87
- [ ] no consistent form to report exceptions between engines ( #497 ) - worked around by picking the first line only https://github.com/saper/sass-spec/commit/66bdbdda154dd75028f2684faf3f8e764e103cda
- [ ] race condition trying to capture Sass ruby module std err output ( #499 ) - worked around by turning off parallel testing https://github.com/saper/sass-spec/commit/22392459d12b6ea23e75be86389fdab26ce0711a

Issue #495, opened 2015-09-19
EngineAdapters: inconsistent status value returned


saper:
When running a subprocess with `-c`, for example `sassc` or Ruby `sass` CLI, the value returned in the `status` variable is a [Process::Status](http://ruby-doc.org/core-1.9.3/Process/Status.html), that contains the PID, an exit code and other information. 
Ruby `sass` CLI returns with an exit code 65 when failed with a `SyntaxError` (in fact any error ocuring during Sass processing, also coming from an `@error` directive). `sassc` fails with an exit code `1` in that case.

When using Ruby Sass as a ruby module, we are getting a numeric output: 0 for OK, 1 for a Sass `SyntaxError` or 2 for some other exception.

This inconsistency prevents us from using the status code in the test framework consistently. For example, when being written to a file **Process::Status** will produce `pid 123123 exit 1` string which is not a simple numeric exit code one can compare against.

saper:
#496 solves the simpler part (now the status is consistently returned integer)

We would need to agree on status code values returned so have the same set of vales accross engines (like 1 - script not found, 2 - test syntax error, 3 ...., etc.).

xzyfer:
I suspect it'll be easier to Sass spec to have a set in internal constants for types of errors. We then infer which type of error a spec run into by the error output.

chriseppstein:
Ultimately, we're just checking for zero or non-zero status even though the returned status is written to the status_code file.

PR #496, opened 2015-09-19
ExecutableEngineAdapter should return numeric status

saper:
ExecutableEngineAdapter and SassEngineAdapter
should consistently return numeric exit code
from the compiler.

Fixes https://github.com/sass/sass-spec/issues/495 (partially)


xzyfer:
LGTM

PR #503, opened 2015-09-20
Support tests with indented syntax

saper:
Now input.scss and input.sass are
reconignized as valid input names.

Fixes: https://github.com/sass/sass-spec/issues/247

### DIFF
--- a/spec/sass/basic/input.scss
+++ b/spec/sass/basic/input.scss
@@ -1,1 +1,0 @@
-@import "input.sass";
--- a/lib/sass_spec/cli.rb
+++ b/lib/sass_spec/cli.rb
@@ -17,7 +17,7 @@ module SassSpec::CLI
 
       # Constants
       output_styles: ["nested", "compressed", "expanded", "compact"],
-      input_file: 'input.scss',
+      input_files: ["input.scss", "input.sass"],
       nested_output_file: 'expected_output',
       compressed_output_file: 'expected.compressed',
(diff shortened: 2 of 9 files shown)


saper:
[Before](https://travis-ci.org/sass/sass-spec/builds/81262682):

    5087 runs, 15261 assertions, 0 failures, 0 errors, 0 skips

[After](https://travis-ci.org/sass/sass-spec/builds/81304527)

    5091 runs, 15273 assertions, 0 failures, 0 errors, 0 skips

It seems that `spec/libsass-closed-issues/issue_1546/input.sass` kicked in

spec/sass/basic/input.scss
```diff
@@ -1 +0,0 @@
-@import "input.sass";
```
xzyfer: What's happened here?

xzyfer: I read the commit message. Please don't change this. It tests an important feature of Sass, being able to mix syntaxes in a single run.

saper:
No, I don't change this :) `basic` test checks for using plain SASS file right away (only `index.sass` there) and `imported` test checks for what `basic` was doing - it invokes `index.scss` that imports `imported.sass`. What was `basic/index.sass` is now `imported/imported.sass`.

One thing is testing for mixing syntaxes, another thing is being forced to use SCSS because SASS could not be directly imported before :) So we have two tests now.

xzyfer:
I see. I misread the change.

Issue #333, opened 2015-04-20
The --unexpected-pass should take into account output styles


xzyfer:
The `--unexpected-pass` should consider a spec as unexpectedly passing, if and only if, all the output styles for that spec also pass.

saper:
Every unit test should independent. Currently every output format variant constitutes a separate case. This means that even if we only check for an error or status code ( #494 ) we currently run 4 tests. Are there cases in real life when this matter? Because to implement this all output formats would need to became assertions in one test, which complicates things a bit.

xzyfer:
Currently determining if a test has magically started passing is PITA.

IMHO the underlying sass spec data model is broken which makes these hard. We treat the entire spec as a single suite with 1000s of unit tests. Instead each spec folder should a suit and it's expected outputs as unit tests. If a single unit test fails that suite should fail. If a suite fails it's parent has failed etc..

saper:
Note to self: sample problematic issues

spec/libsass-todo-tests/libsass/propsets
spec/libsass-todo-issues/issue_1026

saper:
We no longer cater for different output styles.

chriseppstein:
`--unexpected-pass` no longer exists. Use `--probe-todo` instead.

PR #508, opened 2015-09-21
Introduce expected_output.todo markers

saper:
If expected*.todo file is present (alongside of the .css file) this marks
this test as "todo", which can be ignored with
--ignore-todo.

Therefore test which have some output formats passing, but not the other
ones can have a more fine-grained "todo" granularity. Such tests should not
be in the folders that contain "todo", since then a whole test may be
skipped.

Should fix: https://github.com/sass/sass-spec/issues/333


xzyfer:
From Slack

>I don't like [this] approach
>it's already too hard to find how many todo test we have
>.todo hides that we have failing test
>in the same way that I had no idea we have "todo" specs that were in a /todo/ sub folder of other specs
>https://github.com/sass/sass-spec/pull/491/files#diff-c80a744e742eb3c2218e188eb849657b (edited)
>I don't think we can hack a fix a for #333. We should do it right or not at all.

mgreter:
Guess I have missed this one. We already have [pretty much the same feature] [1], which I've added some time ago when I implemented the three missing output modes. I guess the `--ignore-todo` thing might be handy, so not sure if we want replace the `.skip` flag file. There are currently 5  skipped output tests (4 compressed, 1 compact).

I think is still nicer to have 3 spec tests actually passing, than needing to exclude the whole spec test. One of these would i.e. be the bourbon test, which does not pass 100% exactly with compressed mode.

[1]: https://github.com/sass/sass-spec/blob/master/lib/sass_spec/runner.rb#L66

saper:
yes, I was aware of the `skip` feature and I think it is useful. 

Although from your comment I reckon that it was there for cases where particular output formats behave differently and had to be excluded. With selective execution of formats now we can actually just drop the output files in that case ( and prepare another "todo" case for the failing format only.

This feature is different from `skip` in a way that it is conditional on `--ignore-todo`, our favourite libsass testing option.
So tests sub-results can be excluded individually only for libsass, but till get executed by Ruby. `skip` is skipped by both.

But since we no longer need to generate all output formats (with #610 hopefully to come) maybe this feature is no longer needed.

chriseppstein:
No longer needed.

PR #529, opened 2015-10-05
Add specs for issue 1578

xzyfer:
This PR add specs for sass/libsass#1578

### DIFF
--- a/spec/libsass-todo-issues/issue_1578/expected.compact.css
+++ b/spec/libsass-todo-issues/issue_1578/expected.compact.css
@@ -1,0 +1,1 @@
+foo { bar: baz; }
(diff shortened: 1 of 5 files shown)


xzyfer:
@saper looks like a possible bug in sass-spec.

saper:
Strange. When I generated the test (with `-c path/to/ruby/sass`) it didn't produce any warning and there was no `error` file. 

I was using Sass 3.4.14.2e0f33b (Selective Steve)

How did you get that warning?

saper:
https://github.com/xzyfer/sass-spec/pull/1

xzyfer:
Thanks @saper applied your patch in 51c7469.

I think the root cause of why this happened is because the --nuke flag is not idempotent anymore https://github.com/sass/sass-spec/issues/533

PR #554, opened 2015-10-22
Generated missing output styles for warn directive test

mgreter:


saper:
I have left out other styles on purpose, it is not this what we are testing there and this counts as 1 and not 4 tests. I generally use `--generate nested` for cases where an error is expected for example.

PR #555, opened 2015-10-24
Regenerated color function tests

mgreter:
Generated with Sass 3.4.16 (Selective Steve)

libsass will need some adjustements which I will post in a few ticks.
This will actually solve some issues when compiling as 32bit ...

Needs https://github.com/sass/libsass/pull/1621


mgreter:
BTW. I've tested it with ruby sass x64 and x86 (2.2.3) and sass 3.4.19. Both give consistent results!

xzyfer:
This is a known issue. Ruby Sass changed it's colour rounding in 3.4.15 which is why we haven't updated the Sass version of sass-spec.

xzyfer:
Done in https://github.com/sass/sass-spec/pull/590

PR #579, opened 2015-10-29
Improve spec.sh

asottile:
I used this modified script to create #578

### DIFF
--- a/spec.sh
+++ b/spec.sh
@@ -1,4 +1,21 @@
 #!/bin/bash
+set -ex
+
+EDITOR="${EDITOR-subl}"
+
+
+function runspec() {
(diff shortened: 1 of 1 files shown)


xzyfer:
Thanks for this but we won't be accepting these changes. This is largely a personal script which I should probably remove.

The changes here actually harm my workflow with this script. Fwiw this is the intended workflow.

``` 
./spec.sh create todo 1234
./spec.sh nuke todo 1234
git add spec
git commit --amend
git push
```

saper:
@asottile you an use an unfortunately undocumented `--generate` option to the `sass-spec.rb`, for example `--generate all` or `--generate nested`

PR #582, opened 2015-11-02
Normalise path to spec in error message to `/sass/sass-spec/...`

xzyfer:
The diffs caused by difference in paths when nuking test is slowing
down my figuring out a Sass 3.4.19 upgrade path. We should probably
do something like default in sass-spec but this was the fastest
solution to my current problem.


xzyfer:
>We should probably do something like default in sass-spec

Turns out we do. https://github.com/sass/sass-spec/pull/583 probably makes this moot.

PR #598, opened 2015-11-06
Update specs to ruby sass 3.4.16

mgreter:
Also normalizes all paths to prepare comparison of full stderr ourput.
I use the following regex to clean the stderr output:
```
s/(?:\r\n)/\n/g;
s/^(?:\n|\r)+//;
s/\Q$cwd_url\E/\/sass/g;
s/\Q$cwd_path\E/\/sass/g;
s/libsass\-[a-z]+\-issue/libsass\-issue/g;
s/[a-z]+\/sass-spec\//\/sass\/sass-spec\//g;
```


saper:
But I hope paths do not affect testing? We are comparing only the first line of the output (I hope)

mgreter:
Yep, ruby spec runner only tests first line ... I'm a bit further now with my perl spec runner, but of course quite a few error specs do not pass if I compare the full stderr output. So this is in preparation to improve it gradually.

saper:
Good, I think I will just implement skipping over the backtrace only; the only major problem now is that I can't fully test the case that generates two warnings.

PR #608, opened 2015-11-13
Add test for libsass issue #1706

saper:
https://github.com/sass/libsass/issues/1706


saper:
Do we want to test for the deprecated feature like this? Leaving open

xzyfer:
We do because it need to get to 3.4.20 parity before this feature can be deprecated.

PR #626, opened 2015-11-25
Modify hex arithmetic test to cover more cases

mgreter:
We currently do not pass this test 100%
```
div {
  foo: 10 - #a4F;
  bar: 10 / #a4F;
}
```


xzyfer:
I'm confused was this merged or closed?

Issue #627, opened 2015-11-25
Errors may not be equal for all output styles


mgreter:
As I suspected, there are some errors which differ for varying output styles. Today the check I've added in the perl-libsass runner (I mentioned that in a slack chat AFAIR) catched a first sample:

```scss
foo {
  bar: (rgba(16,16,16,0.5) + #AAA);
}
```

results
```
Error: Alpha channels must be equal: rgba(16,16,16,0.5) + #AAA // compressed
Error: Alpha channels must be equal: rgba(16, 16, 16, 0.5) + #AAA // all others
```

//CC @saper

mgreter:
@chriseppstein should errors, warnings and inspect at all respect the output styles and precision settings? This seems not 100% consistent with ruby sass all the times (so far I thought it is ignored).

saper:
@mgreter can you just prepare one directory with 4 subdirectories, 4 identical input.scss files in every one of them, and only one expected*css file per directory? You can use `sass-spec -g format` to generate them one by one.

mgreter:
Not sure if I understand correctly, but how does the spec runner know which output style it has to produce? Or do you just want the files, because I already pushed them to a [local branch] [1].  I actually would propose a structure like this, since I already support this in perl-libsass :)
![sass-error-specs](https://cloud.githubusercontent.com/assets/282293/11411526/f7999d4e-93d0-11e5-8541-e4c58036bfd3.png)
If the optional files don't exists, I expect the default exit code and nothing on stderr ...

[1]: https://github.com/mgreter/sass-spec/tree/test/value-op-errors/spec/errors/subtract/alpha-channels-todo

xzyfer:
Personally I would like to see this addressed upstream. I've opened an issue to discuss this https://github.com/sass/sass/issues/1916

mgreter:
IMHO we would still need different tests if the error reporting is according to output style settings (and ie. precision). Currently we take it for "granted" that errors will be reported the same for all settings. If that would change, we still would need support to test stderr for different options. So IMHO my proposal is still valid, and I'd rather test too much then too little.

saper:
This is my proposed solution to the problem: https://github.com/sass/sass-spec/pull/629

quite simple imho and no changes to sass-spec needed

after putting `input.scss` in the respective directories I did:

````
[...]

PR #629, opened 2015-11-29
Test alpha channel math

saper:
Ruby Sass generates different error message
in the compressed output fromat, therefore
we introduce a separate sub-test for this format only.

Related-to: https://github.com/sass/sass/issues/1916
Related-to: https://github.com/sass/sass-spec/issues/627


saper:
A simpler approach with two tests only: #631

xzyfer:
Closing in favour of #631

Issue #533, opened 2015-10-06
Nuke is not idempotent


xzyfer:
Discovered in https://github.com/sass/sass-spec/pull/529.

If a spec is nuke'd multiple times, and one run produces an error, the error file is never cleaned up.
I expect the same applies in the inverse. If nuking an error spec multiple times and one mistaken produces output, this output is not cleaned up (untested).

saper:
Yes, I know that and this will be fixed soon. I have the same problem with `--generate`.

chriseppstein:
I fixed this as I was working on things. See: https://github.com/sass/sass-spec/blob/master/lib/sass_spec/test.rb#L398-L400.

Also, nuke is gone, use `-g` now.

PR #630, opened 2015-11-29
Make --nuke idempotent

saper:
With this change:

   ruby sass-spec.rb -n -g all <testcase>
      Deletes all test output files and generates new ones
      for all output formats

   ruby sass-spec.rb -g nested <testcase>

      Generates test output files for the nested format only,
      all other formats are left unchanged.

   ruby sass-spec.rb -n -g compressed <testcase>

      Deletes all test output files and generates new one
      ony for the compressed format.

   ruby sass-spec.rb -n <testcase>

      Regenerates test output files for the formats
      already defined for the test.

Fixes: https://github.com/sass/sass-spec/issues/533

### DIFF
--- a/lib/sass_spec/runner.rb
+++ b/lib/sass_spec/runner.rb
@@ -42,10 +42,22 @@ class SassSpec::Runner
         folder = File.dirname(filename)
         expected_stderr_file_path = File.join(folder, "error")
         expected_status_file_path = File.join(folder, "status")
+        if @options[:nuke]
+          [expected_status_file_path, expected_stderr_file_path].each do |f|
+             if File.file?(f)
+               File.unlink(f)
+             end
+          end
+        end
         @options[:output_styles].each do |output_style|
           output_file_name = @options["#{output_style}_output_file".to_sym]
           expected_stdout_file_path = File.join(folder, output_file_name + ".css")
           clean_file_name = File.join(folder, output_file_name + ".clean")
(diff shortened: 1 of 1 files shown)


xzyfer:
I still don't understand the use case for generate.

Nuke should remove all output/error files and regenerate them from the input.scss.

saper:
`-g` is much more gentle - you can create a test case with just one or two output formats. I use it for cases which return the same error for all output formats (so it makes no sense to run 4 tests instead of one) and to generate cases like #629 where we need to test different output formats separately. `--nuke` is more like all or nothing.

lib/sass_spec/runner.rb
```diff
@@ -42,10 +42,27 @@ def _get_cases
...
+        end
         @options[:output_styles].each do |output_style|
           output_file_name = @options["#{output_style}_output_file".to_sym]
           expected_stdout_file_path = File.join(folder, output_file_name + ".css")
           clean_file_name = File.join(folder, output_file_name + ".clean")
+          if @options[:generate].size > 0
+             print 'Generate!'
+          end
+          print 'Size: '
+          print @options[:generate].size
```
mgreter: I guess these are left-overs?

saper:
Thanks @mgreter, good catch!

chriseppstein:
this is no longer needed.

PR #631, opened 2015-11-29
Test alpha channel math

saper:
Test only for one format (normal nested format)
because we are checking only for the error message
and no normal output is produced anyway.

This also helps to avoid a situation
where Ruby Sass produces slightly different
error messages for the compressed output format.

Related-to: https://github.com/sass/sass/issues/1916
Related-to: https://github.com/sass/sass-spec/issues/627


xzyfer:
Please move this to the `libsass-todo-tests` folder and not create any more `-todo` specs. They're greatly hamper discoverability.

saper:
Sorry, it is just a subset of @mgreter tests and I didn't really intend that for merging. But I have fixed this here!

saper:
More extensive approach: #629

xzyfer:
This seems like the correct approach to me. I don't see the value in testing error messages for other output modes.

PR #642, opened 2015-12-12
Nuked with 3.4.20 (minus interpolation deprecation warning)

mgreter:
Will push a PR for libsass to pass all 3.4.20 specs minus interpolation deprecation warnings.


mgreter:
CI for this PR showed a behavior I also saw locally. Ruby Sass doesn't seem to produce very random numbers. We pretty often get duplicates on the first 1000 runs (as done by the test). I've added a more tolerant version that allows up to 25 duplicates.

mgreter:
Continued in https://github.com/sass/sass-spec/pull/669

PR #643, opened 2015-12-26
Add tests for issue #1801

saper:
Testing for https://github.com/sass/libsass/issues/1801

Since the infinite loops keep sassc busy,
this should probably merged after fixing this.
Alternatively those tests should be run
with a limited stack size (not implemented).


saper:
Travis fails because there is an embedded full path in the error message and we compare all of it.

xzyfer:
Thanks. We'll need to work around the path issue otherwise this spec will cause CI to always fail.

saper:
Is `simple-loop-error` passing now with the truncated error message...?

saper:
Ah, #674 fixed it probably..?

mgreter:
One still made some problems. Need to check later, moved it back to todo folder.
I'm in the middle of sorting it all out ...

saper:
Will check it... It this PR I have removed the path since it was in the first line of the output. with #674 in place it should work, I just need to regenerate the case and #674 should truncate it.

mgreter:
I'm pretty positive that I have regenerated it, but one failed nonetheless in CI (it worked locally) ...

PR #655, opened 2015-12-28
Revert "Fix path resolution regressions for absolute spec directory path"

xzyfer:
Reverts sass/sass-spec#653

This patch broke the runner on Windows. This requires further investigation to fix the regression with absolute spec directory path.

/cc @am11


am11:
Thanks for catching this. I will take a look too. :)

PR #660, opened 2015-12-28
code: Fixes abosolute path issue

am11:
A regression was introduced in #640 which broken spec file discovery
with when `spec_directory` was an absolute path.

//cc @xzyfer


xzyfer:
Works locally :+1: Thank you

PR #664, opened 2015-12-30
spec: Modifies 1096 input with CRLF

am11:
This is to test sass/libsass#1817.

//cc @xzyfer


am11:
Now how to test it with AppVeyor without sass/libsass#1817. :D

xzyfer:
@am11 if you add the sass-spec PR URL to the libsass PR body CI will build against that branch of sass-spec.

PR #669, opened 2016-01-05
Nuka cola 3.4.21

mgreter:
Continuation from https://github.com/sass/sass-spec/pull/642 since 3.4.20 still has some difficult to implement differencies with numbers and compressed output.


saper:
If we (hopefully) merge https://github.com/sass/sass-spec/pull/630 first, `--nuke` will regenerate only existing output formats and will not re-generate additional 3 dummy unit tests for cases where we only test for an error output.

mgreter:
FYI I nuke with perl-libsass and this should already be the case there (perl-libsass can nuke the whole thing in parallel, which gives quite a nice boost on a deca or octa core). Or did you see anything in this PR that should be different?

saper:
There are additional tests added like `spec/libsass-todo-tests/errors/addition/alpha-channels/expected.expanded.css`. I should have also noticed ` spec/libsass-todo-tests/errors/addition/alpha-channels/expected.expanded.css.error` which does not come from `sass-spec.rb` obviously. I think those are superfluous and just contribute to a longer running time.

mgreter:
@saper good catch, thought I had removed them manually. Was aware that perl-libsass nuker did not yet clean these, but I've added that now. Will first update and push current spec tests with 3.4.16 in a speparate PR. I also added some path cleaning for the error specs:
```perl
$str =~ s/\Q$cwd_url\E/\/sass/g;
$str =~ s/\Q$cwd_path\E/\/sass/g;
$str =~ s/\/libsass\-[a-z]+\-test\//\//g;
$str =~ s/\/libsass\-[a-z]+\-issue/\/libsass\-issue/g;
$str =~ s/[a-z]+\/sass-spec\//\/sass\/sass-spec\//g;
```
This makes it possible to activate todo specs without the need to update them. It should also make it possible to test the full output on stderr and not only the first line at some later point.

mgreter:
Some deprecation warnings seem to have gone:
```
DEPRECATION WARNING on line 1, column 1:
The subject selector operator "!" is deprecated and will be removed in a future release.
This operator has been replaced by ":has()" in the CSS spec.
For example: x:has(.bar)
```

```
DEPRECATION WARNING: Passing 3, a non-string value, to unquote()
will be an error in future versions of Sass.
        on line 2 of /sass/sass-spec/spec/libsass-issues/issue_1291/input.scss, in `spec1'
        from line 16 of /sass/sass-spec/spec/libsass-issues/issue_1291/input.scss
```

```
DEPRECATION WARNING on line 4, column 13 of /sass/sass-spec/spec/libsass-issues/issue_1096/input.scss:
Unescaped multiline strings are deprecated and will be removed in a future version of Sass.
To include a newline in a string, use "\a" or "\a " as in CSS.
```

We are probably safe to remove them too, so documenting the cases here for now.

mgreter:
I had to deactivate two test since they had minor regressions in CI. It only happened on linux with BUILD=shared. A single `\r` got replaced with ` `. I don't know why this happens for that config, but I'm pretty sure it has to do with linefeed conversion (binmode not set on filehandles).

PR #692, opened 2016-01-12
Improve spec test runner

mgreter:
Had to make some changes to get everything running correctly on windows.
Also fixed a small issue that is also addressed in https://github.com/sass/sass-spec/pull/630 (//CC @saper)

It changes the string handling internally to not bother about the encoding. We need to open/read files with `:binmode => true` on windows, as it's the only way I know to disable the linefeed conversion, which triggered a wrongly failed spec test. And we really don't care about the encoding as long as the binary representations are exactly equal.

Also re-introduced parallelize_me conditionally, since it should be ok when invoking external commands.

Also cleaned up the cleanup functions a bit. Hope CI doesn't mind!

### DIFF
--- a/lib/sass_spec/engine_adapter.rb
+++ b/lib/sass_spec/engine_adapter.rb
@@ -1,5 +1,3 @@
-require "open3"
-
 class EngineAdapter
   def describe
     not_implemented
@@ -39,13 +37,15 @@ class ExecutableEngineAdapater < EngineAdapter
   end
 
   def version
-    stdout, stderr, status = Open3.capture3("#{@command} -v")
+    require 'open3'
+    stdout, stderr, status = Open3.capture3("#{@command} -v", :binmode => true)
     stdout.to_s
   end
 
 
   def compile(sass_filename, style)
-    stdout, stderr, status = Open3.capture3("#{@command} -t #{style} #{sass_filename}")
+    require 'open3'
+    stdout, stderr, status = Open3.capture3("#{@command} -t #{style} #{sass_filename}", :binmode => true)
     [stdout, stderr, status.exitstatus]
   end
 end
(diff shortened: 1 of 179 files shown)


lib/sass_spec/engine_adapter.rb
```diff
@@ -66,20 +66,34 @@ def version
 
   def compile(sass_filename, style)
     require 'sass'
+    # restore old fh
+    stderr = StringIO.new
+    # restore previous default encoding
+    encoding = Encoding.default_external
+    # Does not work as expected when tests are run in parallel
+    # It runs tests in threads, so stderr is shared across them
+    old_stderr, $stderr = $stderr, stderr
```
saper: $stderr is global, how can make sure the output of parallel tests does not mix here?

lib/sass_spec/test.rb
```diff
@@ -1,9 +1,16 @@
...
 
+parallelize = false
+
 def run_spec_test(test_case, options = {})
   if options[:skip_todo] && test_case.todo?
     skip "Skipped todo"
   end
+  # When running executables we can safely parallelize
+  # Executables must become their own process environment
+  if options[:engine_adapter].instance_of? (ExecutableEngineAdapater)
```
saper: I'd rather introduce method/property here, instanceof is evil

lib/sass_spec/engine_adapter.rb
mgreter: stderr is bound to the process. Since we IMO cannot change the way that ruby sass prints these messages to stderr. Only way I see to do it portable is to use real processes and not threads.

lib/sass_spec/test.rb
mgreter: Why is it evil? Seems to be standard OO idiom?

saper: Instances should have a paralellizable? method that determines if it can be parallelized or not.

That's also the reason why our use of RTTI is bad, but I digress...

mgreter: Will remove the commit again, had a thinking error. It would only work if we redirect stderr after the process is started. This still has a race condition between redirection of stderr and open3.

mgreter:
Will merge this once CI is green. Next stop 3.4.21 :smiley:

PR #713, opened 2016-01-18
Add spec test for error reporting with unicode involved

mgreter:
https://github.com/sass/libsass/issues/1856


mgreter:
Looks like spec runner has some problems with unicode on it's own. Well perl-libsass spec runner also showed an issue with handling unicode.  https://github.com/sass/sass-spec/issues/714

mgreter:
Looks like that did the job! Thanks @am11 :+1:

PR #729, opened 2016-03-09
Add an option to not exit after running tests.

chriseppstein:
This is needed to integrate sass-spec with ruby sass's build.


xzyfer:
lgtm

saper:
That exit shouldn't be there in the first place. Returning error code via exit from sass-spec.rb would not freak the sass suite, would it?

PR #730, opened 2016-03-12
Ability to consume sass-spec as a rubygem.

chriseppstein:
Ruby Sass will consume sass-spec as a rubygem via this git repo. No need to release it to rubygems.org. This structure is just really helpful for doing parallel development and downloading sass-spec for CI and in development.


xzyfer:
:+1:

PR #743, opened 2016-03-17
Add specs for sass/libsass#1944

danielguillan:
sass/libsass#1944


xzyfer:
Thanks @danielguillan :tada:

PR #744, opened 2016-03-18
Ruby sass integration

chriseppstein:
I ran into numerous issues getting ruby sass's tests working against sass-spec. This PR fixes all of them. The most notable is that there's a file path length limit in rubygems that gets enforced when consuming sass spec. This is mostly fine, but on travis, the installed paths are unusually long. In order to prevent failures from happening again, I've added validation of the length of filenames to check if they are too long. I had to rename a bunch of files and then I renamed a few others that didn't require it just to keep what felt like a naming convention consistent.


xzyfer:
Testing this against LibSass master in https://github.com/sass/libsass/pull/1957

xzyfer:
LibSass is happy :+1:

PR #745, opened 2016-03-28
Language versions

chriseppstein:
This PR introduces a proper metadata system to sass-spec (as opposed to encoding it within naming of directories). Metadata can be set on any directory in the spec directory and that data will be inherited down the file tree. Metadata is stored in the `options.yml` file and I suspect that it will expand to control other aspects of how those tests are ran and managed.

There's a whole new suite of command line tools available for managing the test metadata available using `sass-spec.rb annotate -h`

This PR gets the basic system in place and add support for controlling which test cases are selected according to the desired language version. This is the minimum we need to start developing new specs for sass 3.5 and 4.0 and running different tests against different sass language versions.

Once we have alpha releases for sass 3.5 and 4.0 we can start running all tests for sass-spec in CI, for now, CI will only run sass 3.4 specs.


xzyfer:
LGTM. I've got a PR updating LibSass' CI to run against 3.4.

PR #746, opened 2016-03-29
Refactor warnings

chriseppstein:
This brings warnings in line with current ruby sass warnings.

It also removes specific exclusions for not erroring for some ruby warnings. Doing so did introduce some new expected errors, but not many. I'd like to land this on top of the `language_versions` branch and add specific metadata about expecting a failure for warnings vs failing for output so that libsass can have TODOs around deprecation warnings while still avoiding output regressions.

### DIFF
--- a/lib/sass_spec/test.rb
+++ b/lib/sass_spec/test.rb
@@ -489,6 +489,7 @@ def change_options(options_file, options)
 
 end
 
+
 GEMFILE_PREFIX_LENGTH = 68
 # When running sass-spec as a gem from github very long filenames
 # can cause installation issues. This checks that the paths in use will work.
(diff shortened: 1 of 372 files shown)


lib/sass_spec/test.rb
```diff
@@ -138,6 +67,34 @@ def run_spec_test(test_case, options = {})
...
+  error.each_line do |line|
+    if consume_next_line
+      next if line.strip == ""
+      error_message += line
+      break
+    end
+    if (line =~ /DEPRECATION WARNING/)
+      if line.rstrip.end_with?(":")
+        error_message = line.rstrip + "\n"
+        consume_next_line = true
```
chriseppstein: This lets us verify deprecation messages match up better than just the first line can accomplish.

chriseppstein:
280 skips now :(

spec/libsass-closed-issues/issue_1916/error
```diff
@@ -1,5 +1,5 @@
 Error: ".btn:hover" failed to @extend ".z-depth-half-1".
        The selector ".z-depth-half-1" was not found.
        Use "@extend .z-depth-half-1 !optional" if the extend should be able to fail.
-        on line 13 of /home/saper/sw/sass-spec/spec/libsass-todo-issues/issue_1916/input.scss
```
chriseppstein: @saper ಠ_ಠ

PR #760, opened 2016-04-05
Activate one spec test for libsass issue 1651

mgreter:
Activate to ensure we don't break existing behavior!
The `without` test (original issue) is still in todo folder!


mgreter:

PR #775, opened 2016-04-21
Interactive mode

chriseppstein:
This PR introduces a new `--interactive` option for running sass spec.

Whenever a test would fail, the runner stops and asks you what you'd like to do. A number of reparations are made available in a contextual way. We can add more common fixes to interactive mode as we see a need for them.

Common options:

1. update the expected output to match the actual output
2. migrate the test. This makes a copy of the test, updates the copy to expect the new results and sets the new test to run starting in the current version. The old test is then set to end in the version before the current version.
3. show details (source, expected, actual, etc)
4. let the failure happen
5. exit the test suite
6. mark the test as "todo" for the current implementation.

There is a new class `SassSpec::Interactor` that creates a simple interface for declaring prompts, options, and what to do when options are selected. The existing examples in code should be sufficient to understand how to use it.


xzyfer:
My Ruby is pretty meh but I agree with the intent :+1:

chriseppstein:
rebased and landed on master at 41a137e0012a01e55e8ecbbd8936f060626a16fd

Issue #600, opened 2015-11-07
Run todo tests by default for CI


mgreter:
IMO it would be nice to also check the todo tests on CI. We should not alter the overall error state for these, but it should report the spec tests that are unexpectedly passing. We sometimes add todo tests for segfaults, which should not be included here, since it would abort the whole test runner. Therefore I propose to add tests for segfaults to another folder, that would be excluded from this proposed test run.

xzyfer:
Sass spec doesn't exists to serve LibSass, it's for all implementors of the Sass language. We should endeavour to remove LibSass specific logic not add more.

Testing for segfaults and such is something we should probably do in LibSass via a different method. I know @am11 has thoughts on this.

am11:
Copy/paste from slack chat:

> We can test Unix Signals (including  `sigsegv` ) with Criterion. Here is a sample: https://github.com/Snaipe/Criterion/blob/ec6ea15/samples/signal.c. :smile:

xzyfer:
I think this is something we should move into LibSass. As well as testing the C-API.

mgreter:
@xzyfer I mean todo specs that make the libsass master segfault (which we have had in the past). It's pretty inconvenient when I try to find todo specs that are passing with master automatically. I always have to move tests that segfault first and then make sure I don't accidentally commit it. I know it's not ideal, but we don't live in an ideal world. So I still would like to add spec tests that currently segfault for master to a `libsass-todo-segfaults` folder (instead of `libsass-todo-issues`). IMO this shouldn't change anything for ruby spec runner, it would just allow me and others (including the CI test runner) to differ between segfaulting todo tests and ones I can automatically check.

xzyfer:
@mgreter I understand now. I personally have not run into this. It seems worth doing if someone has the time and desire.

saper:
@mgreter one question: if you have segfault tests in, say, `libsass-todo-issues`, what happens if you run `sass-spec.rb` against them without `--ignore-todo` ? You probably don't need to move them away?

Do I understand it correctly that you basically need to run `sass-spec.rb --unexpected-pass` ?

mgreter:
@saper don't really understand the relation to the issues you linked. But yes, the problem exists when we have segfaulting todo tests and run `--unexpected-pass` (in `perl-libsass` I unset `skip-todo-tests`).

[...]

PR #836, opened 2016-04-22
Add mode to probe for passing todo tests

mgreter:
Fixes https://github.com/sass/sass-spec/issues/600

### DIFF
--- a/lib/sass_spec/runner.rb
+++ b/lib/sass_spec/runner.rb
@@ -42,7 +42,18 @@ class SassSpec::Runner
       Minitest.reporter = Minitap::TapY
     end
 
-    Minitest.run(minioptions)
+    result = Minitest.run(minioptions)
+
+    if @options[:probe_todo]
+      puts "Reporting successfully probed todo tests"
+      test_cases.each do |test_case|
+        if test_case.todo? && test_case.result?
+          puts test_case.folder
+        end
+      end
+    end
+
+    result
   end
 
   def language_version
--- a/lib/sass_spec/test.rb
+++ b/lib/sass_spec/test.rb
@@ -16,7 +16,7 @@ def run_spec_test(test_case, options = {})
 
   if test_case.overwrite?
     overwrite_test!(test_case)
-    return
+    return true
   end
 
   return unless handle_missing_output!(test_case)
@@ -30,6 +30,8 @@ def run_spec_test(test_case, options = {})
     return unless handle_expected_error_message!(test_case, options)
     return unless handle_unexpected_error_message!(test_case, options)
   end
+
+  return true
 end
 
 # if the test case is interactive it will do the interaction and return
@@ -51,6 +53,7 @@ def handle_expected_error_message!(test_case, options)
   expected_error_msg = _extract_error_message(test_case.expected_error, options)
 
   if expected_error_msg != error_msg
+    return false if test_case.probe_todo?
     interact(test_case, :fail) do |i|
       i.prompt(error_msg.nil? ? "An error message was expected but wasn't produced." :
                                 "Error output doesn't match what was expected.")
@@ -106,13 +109,14 @@ end
 
 def handle_unexpected_error_message!(test_case, options)
   return true if test_case.verify_stderr?
-  
+
   _output, _clean_output, error, _status = test_case.output
 
   error_msg = _extract_error_message(error, options)
 
   return true if error_msg.nil? || error_msg.length == 0
 
+  return false if test_case.probe_todo?
   interact(test_case, :fail) do |i|
     i.prompt "Unexpected output to stderr"
 
@@ -160,6 +164,7 @@ def handle_output_difference!(test_case, options)
     return false
   end
 
+  return false if test_case.probe_todo?
   interact(test_case, :fail) do |i|
     i.prompt "output does not match expectation"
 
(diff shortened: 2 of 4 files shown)


mgreter:
Any objections @chriseppstein @xzyfer ?

See https://github.com/sass/libsass/pull/2035#issuecomment-213621537 for the benefits!

lib/sass_spec/test.rb
```diff
@@ -51,6 +53,8 @@ def handle_expected_error_message!(test_case, options)
   expected_error_msg = _extract_error_message(test_case.expected_error, options)
 
   if expected_error_msg != error_msg
+    return false if options[:probe_todo] && test_case.todo?
+    return false if options[:probe_todo] && test_case.warning_todo?
```
chriseppstein: I don't understand why you want this to pass when failing. I thought the goal was to find tests that were passing.

lib/sass_spec/test.rb
```diff
@@ -106,13 +110,15 @@ def handle_expected_error_message!(test_case, options)
...
-  
+
   _output, _clean_output, error, _status = test_case.output
 
   error_msg = _extract_error_message(error, options)
 
   return true if error_msg.nil? || error_msg.length == 0
 
+  return false if options[:probe_todo] && test_case.todo?
+  return false if options[:probe_todo] && test_case.warning_todo?
```
chriseppstein: same comment as above.

lib/sass_spec/test.rb
```diff
@@ -253,6 +261,7 @@ def handle_unexpected_pass!(test_case, options)
       return false
     end
 
+    return false if options[:probe_todo] && test_case.todo?
```
chriseppstein: I think we should define TestCase#probe_todo? and it should evaluate to `options[:probe_todo] && (todo? || warning_todo?)`

chriseppstein:
Actually, thinking about this, It seems to me that if you're running todo tests, we should just print this out afterwards. I don't see a need for a new option.

lib/sass_spec/test.rb
mgreter: It returns false, so it indicates a failure IMO. It just does not trigger the assertion, so that flag can be stored to `result?`. In the end I want the same behavior as without `--run-todo`, but to report which todo tests would actually have passed.

mgreter:
`--run-todo` seems to report todo tests that still fail, and I don't want to see that in CI.

mgreter:
But I agree somehow that I would have wanted to `--run-todo` as `--probe-todo` does now ...

mgreter:
The `--run-todo` test still seems usefull with `--interactive`, but `--probe-todo` could also respect that flag. I don't really see a use case for `--run-todo` to report expected failures ...

lib/sass_spec/test.rb
mgreter: I agree, should probably also respect interactive.

mgreter:
@chriseppstein I have ammended the commit in regard to your feedback.

lib/sass_spec/runner.rb
```diff
@@ -42,7 +42,18 @@ def run
       Minitest.reporter = Minitap::TapY
     end
 
-    Minitest.run(minioptions)
+    result = Minitest.run(minioptions)
+
+    if @options[:probe_todo]
+      puts "Reporting successfully probed todo tests"
```
mgreter: Here I access the stored result to determine which todo tests did pass.
The ones that did not pass should be ignored as if they were skipped.

lib/sass_spec/test.rb
```diff
@@ -16,7 +16,7 @@ def run_spec_test(test_case, options = {})
 
   if test_case.overwrite?
     overwrite_test!(test_case)
-    return
+    return true
```
mgreter: Not sure if this should actually return true

lib/sass_spec/test.rb
```diff
@@ -30,6 +30,8 @@ def run_spec_test(test_case, options = {})
     return unless handle_expected_error_message!(test_case, options)
     return unless handle_unexpected_error_message!(test_case, options)
   end
+
+  return true
```
mgreter: indicates that the test case passed.
We could probably store result to test_case here

mgreter:
I'm going to abuse my powers here and will proceed to merge. This really speeds up my cycles with activating spec tests. It's also great to see subtle compiler differences. I really think this should be how `--run-todo` should work as I don't see a reason why we want to fail the runner if a todo task fails.

chriseppstein:
So, `--probe-todo` is:

1. print out of report of passing todo specs.
2. pass todo specs that are failing.

I think we should always print out a report of passing todo specs if they are ran.

and instead of marking todo specs that fail as passing, I think they should still be marked as 'skip'

I have a few minutes so I'll make these changes to your branch and land them.

mgreter:
Yep, that's pretty much it, just be aware this is already on the master branch ;)

mgreter:
BTW. one outcome of it was that I was able to put this together: https://github.com/sass/libsass/issues/2039 - in the old workflow I had to create a special PR on sass-spec then one on libsass, to see on which compiler these were failing.

chriseppstein:
see: d1e615e7b67b45292e0e6492de5f18cd84c11925 and 841f57b3fe5d707b3764088e29bd6c5bf241e646

mgreter:
Pretty much anything is fine with me here :smile: Can you make a PR to sass-spec and then one in libsass to see how CI does behave?

mgreter:
One very minor nitpick is that the 1st line of the git commit is "too" long and wraps around.
I personaly "try" to stick to the "official" git rules (60 chars) and I also tend to start the message with a verb so the intention of the action is easier visible (i.e. add, remove, patch, fix, improve, etc). But as I said, just very minor nitpicking :wink:

PR #838, opened 2016-04-23
[CI] Test probe mode to report actual fails

mgreter:
Just to be sure


mgreter:
That seems to work as expected!

PR #840, opened 2016-04-23
Activate todo spec test for libsass issue 424

mgreter:
This is only failing with MSVC mode (https://github.com/sass/libsass/issues/2039)


mgreter:

PR #846, opened 2016-04-25
Remove :nuke option as it was redundant

chriseppstein:
`:generate` and `:nuke` ended up being identical after my recent work. This removes `:nuke`

### DIFF
--- a/lib/sass_spec/cli.rb
+++ b/lib/sass_spec/cli.rb
@@ -102,10 +102,6 @@ Make sure the command you provide prints to stdout.
         # Note: --skip is no longer necessary as this is now the default behavior, since we test for errors
       end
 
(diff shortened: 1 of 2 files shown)


mgreter:
LGTM :wink:

mgreter:
I guess this option is not used in any script we have in the sass/libsass ecosystem?

xzyfer:
:+1:

chriseppstein:
> I guess this option is not used in any script we have in the sass/libsass ecosystem!?

It's a development-time tool for building/updating specs, maybe they are, but I suspect all the people who care are CC'd on sass-spec PRs.

mgreter:
@chriseppstein my first thought was actually to everything under sass github org. Normally I do a recursive search on all repos in cases like these, but I'm pretty sure no CI (or other) script uses that option :)

Issue #707, opened 2016-01-17
Missing error specs for warn family parsing errors


mgreter:
From https://github.com/sass/libsass/issues/1865 as it fits better here.
We should already do the right thing but could use these specs:

Each of these produces a parsing error:
```scss
@warn;
@error;
@debug;
@warn("foo";
@error("foo";
@debug("foo";
```

PR #848, opened 2016-04-26
Add missing error specs for warn family parsing errors.

chriseppstein:
Addresses #707


chriseppstein:
@mgreter is this what you had in mind?

mgreter:
:+1:

PR #857, opened 2016-05-02
Activate specs for issue 1732

xzyfer:
This PR activates specs for sass/libsass#1732


xzyfer:
I couldn't enable one of the test because LibSass and Sass error at different places with different messages.

@chriseppstein is there a way to annotate a test as erring correctly with the wrong error message?

PR #859, opened 2016-05-03
Add initial tests for `set-nth` and maps

mgreter:
Accompanying https://github.com/sass/libsass/pull/2068 and https://github.com/sass/libsass/issues/1930


nex3:
Looks like there are already specs for this: https://github.com/sass/sass-spec/blob/master/spec/values/maps/set-nth/input.scss#L11

PR #861, opened 2016-05-03
Test all ruby versions

chriseppstein:
This patch makes it easy to test sass spec against all ruby versions in development and sets up CI to do this for every build.


chriseppstein:
This can't land until https://github.com/sass/sass/issues/2043 is addressed.

PR #893, opened 2016-08-23
Add specs for issue 2151

xzyfer:
This PR add specs for sass/libsass#2151


nex3:
Closing this because it relies on old-style interpolation parsing, and I don't think it's worth landing it *and* migrating.

PR #896, opened 2016-08-25
Test for equality of list

lokson:
This will test that lists are compared more strict.
Related to: https://github.com/sass/sass/issues/2121
and: https://github.com/sass/sass/pull/2124


nex3:
@lokson Can you update the expected output to be in expanded mode (or give me write access to this PR so I can do it)?

spec/sass_3_5/change-equality-of-lists/input.sass
```diff
@@ -0,0 +1,5 @@
+a
```
nschonni: This should probably be using SCSS syntax. The only indented syntax tests are under https://github.com/sass/sass-spec/tree/master/spec/sass although I get why that is confusing since this folder does start with `sass`

lokson:
My every PR in sass / sass-spec should have "allow edits from maintainers" now. Do whatever you want with that code, I'm ok if you throw it away. The only one I really cared was multi line.

nex3:
@lokson Does that mean you're no longer interested in making changes to these PRs?

lokson:
@next, I will do changes in those PRs if you ask but I need help with sass-spec part. Related test files were always provided. Some of them may need to be moved. Folders inside sass-spec were always unclear for me and I have no idea how it all fits with dart-sass.

nex3:
The folder structure is kind of unclear to everyone :stuck_out_tongue_closed_eyes:—there have been a number of different layout conventions over time that have never really been normalized.

In this case, bracket lists are an entirely new feature to Sass 3.5, so their specs are defined in `spec/sass_3_5/bracket_lists`. I think the best way to add equality specs is to add lines to that file that verify the output of the following cases, the the style of the existing lines:

* `[foo bar] == [foo bar]`
* `[foo bar] == [foo, bar]`
* `[foo bar] == (foo bar)`
* `[] == []`
* `[] == ()`

xzyfer:
@nex3 I updated the specs as suggested. Please a take a look and merge if it's good to go. I'll have a LibSass PR ready in the morning.

PR #897, opened 2016-08-25
Test error when passing empty block

maciej-ka:
Tests that an error is raised whenever an empty block is passed to a
mixin.

Related to: https://github.com/sass/sass/issues/2106
and: https://github.com/sass/sass/pull/2112


nschonni:
Think this needs an options.yml to mark it as todo for ruby sass, but now libsass
```yml
---
:todo:
  - sass
```
That should fix the build failure

nex3:
@lokson Can you add a TODO for dart-sass?

nex3:
@nschonni I don't think this should have a Ruby Sass TODO; the fixes already exist in sass/sass#2112.

nschonni:
@nex3 maybe i'm confused with the flow. I thought it would be TODO till the sass patch actually landed, then it gets flipped on.

nex3:
Here's how I've been doing fixes for Dart Sass:

* Create a sass-spec PR with updated specs.
* Create a Dart Sass PR that links to the sass-spec PR. All major impls have special logic to run against the sass-spec PR if one is linked.
* If the Dart Sass PR passes Travis and the sass-spec PR passes all tests *except* the one for Dart Sass, land both PRs simultaneously.

This was we get full CI testing for changes without needing to worry about updating Gemfiles before landing and so forth. The downside is that sass-spec always looks busted on Travis, although theoretically there are a couple ways we could fix that.

xzyfer:
Fwiw this also my flow for LibSass. It makes for small, discrete units of
work across both repos.

PR #898, opened 2016-08-30
[WIP] NPM packagable version

nschonni:
This isn't meant to replace the Ruby version, but as a helper for https://github.com/sass/node-sass/issues/1686
 - [x] Port node-sass spec runner
 - [x] Teach it to read the options.yml
 - [x] Teach it the `todo` syntax
 - [x] Teach it about version ranges
 - [x] Fix `error` syntax
 - [x] Teach it the new `status` outputs
 - [x] Fix recursive "suites", i.e. nested folders
 - [ ] Group tests into suites by folders
 - [ ] Fix hard-coded `utf-8` assumption that fails some tests

This can always get split out to a separate repo with a submodule if preferred.

/cc @xzyfer these are the remaining failures right now https://gist.github.com/nschonni/5ec9cbc20b3f9747c84c9e9e3625cfda

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,5 @@ tmp
 .sass-cache
 .DS_Store
 Gemfile.*.lock
+node_modules/
+*.log
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,11 @@
 rvm:
   - "2.3.1"
+env:
+  - NODE_SASS=yes
+  - RUBY_SASS=yes
+matrix:
+  allow_failures:
+  - env: NODE_SASS=yes
 script: "./test_all_impls.sh"
 language: ruby
 sudo: false
--- a/package.json
+++ b/package.json
@@ -24,6 +24,7 @@
   },
   "dependencies": {
     "mocha": "^3.0.2",
+    "object-merge": "^2.5.1",
     "read-yaml": "^1.0.0"
   }
 }
--- a/package.json
+++ b/package.json
@@ -23,6 +23,7 @@
     "node-sass": "^3.8.0"
   },
   "dependencies": {
-    "mocha": "^3.0.2"
+    "mocha": "^3.0.2",
+    "read-yaml": "^1.0.0"
   }
 }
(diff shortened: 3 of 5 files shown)


xzyfer:
@nschonni I think some of those failures are due to specs that are valid for LibSass master but not the latest release. I've created a `3.3.6` tag at a point in time where `node-sass@v3.8.0` should pass sass-spec.

nschonni:
I restarted off the 3.3.6 tag here https://github.com/nschonni/sass-spec/tree/336-npm-version
May be better, but the errors jumped up because I know run the nested test folders https://gist.github.com/nschonni/d33dcf14e6b5a1aab74cd858db02f891
Most of the errors seem to be coming from `compressed` differences not removing trailing semicolons and some whitespaces.

nschonni:
The commits are a mess right now, but this passes if you pull the `sass-spec.js` and `package.json` to the 3.3.6 tagged content.

saper:
I think this is very cute and a very good idea. What about putting sass-spec.js in the node-sass repository instead to avoid circular dependency (sass-spec would depend on node-sass, which would depend on sass-spec for testing...)?

maybe it would generally a good idea to keep sass-spec module to contain only the test data, and the runner code (js and ruby modules) should be delegated to separate packages.

nschonni:
@saper I'm going to PR this to node-sass in the meantime. This wouldn't actually cause a circular depenedncy because node-sass is listed in the devDependencies (although peerDependencies is better, the npm3 changes make it less than ideal.
I do like idea of splitting up the runner, but it will probably take some work since libsass also uses the runner in its CI.

xzyfer:
@saper is correct, this belongs in node-sass. We already have this in node-sass as [test/spec.js](https://github.com/sass/node-sass/blob/master/test/spec.js). The issue sass/node-sass#1686 is referring to is that `test/spec.js` wasn't unaware of the recently added annotations. It was in need of this upgrade :)

nschonni:
If i end up getting the runner fixed in node-sass, I'll come back here and we can have a simple package.config to just publish the `spec` folder to NPM.

xzyfer:
Makes sense.

nschonni:
Closing since the spec-runner is PR'd against node-sass, and you've PR'd the package.json here

PR #904, opened 2016-09-03
Add SassC to the CI build setup

nschonni:
Uses the same build logic as the SassC CI setup, and adds an extra
build to the Travis matrix. Currently the Sass Language for SassC is
pinned to 3.4 to match libsass testing.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,7 @@ tmp
 .sass-cache
 .DS_Store
 Gemfile.*.lock
+
+sassc/
+
+libsass/
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,8 @@
 rvm:
   - "2.3.1"
+env:
+  - RUBY_SASS=yes
+  - SASSC_SASS=yes LANGUAGE_VERSION=3.4
 script: "./test_all_impls.sh"
 language: ruby
 sudo: false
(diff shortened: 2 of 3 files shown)


xzyfer:
This is going to an issue to LibSass development. If I understand correctly this PR would cause sass-spec to fail PRs for activating specs that aren't on LibSass master yet.

The LibSass development workflow is to create a sass-spec PR that activates a spec the LibSass PR is intending to fix. The LibSass CI will check out that sass-spec PR and test against that branch instead of master. This asserts that the LibSass PR does indeed fix the what it's intending too.

Example sass/libsass#2062 was intended to fix sass/libsass#1732. To assert it did that we created sass/sass-spec#857 which activates the relevant spec. This means that the LibSass PR will not pass CI if that spec (or any other active spec) fails.

nschonni:
Sorry, I had added the SassC stuff to `allowed failures`, but I must of removed it before the final push

nschonni:
I'll reopen this in a minute, just closing while a fixup and test with travis on my own Travis setup

nschonni:
OK, exanded the test matrix to include 3.4, 3.5, and 4.0 with SassC, but added all of them to the allowed failures in the matrix. Also moved the ci build stuff to a separate script, so the other one can continue to be used by the devs

PR #921, opened 2016-09-18
Add unix_newline option parsing

nschonni:
Honestly doesn't do much right now, but will allow testing better on Appveyor/Windows
Not really a Ruby dev, so the implementation might be bad


nex3:
Closing this out; I don't think we want to require Unix newlines everywhere, because we should be testing CR normalization anyway (I don't know if we *are* yet, but we should be).

PR #912, opened 2016-09-08
Add specs for issue 2082

xzyfer:
This PR add specs for sass/libsass#2082

PR #930, opened 2016-09-25
Activate specs for issue 2140

xzyfer:
This PR activates specs for sass/libsass#2140


xzyfer:
This spec is running into the same issue as #912. Commented out the offending lines as they were insignificant to the issue being tested.

Issue #950, opened 2016-10-17
Use the options.yml for TODO flagged tests


nschonni:
Instead of moving tests around from folder to folder for the todo (mostly libsass right now), what if we just create create the simple `options.yml` in the individual spec folder? Then the flag to remove the test would either update the `options.yml` file to remove the specific todo, or delete the file if it had no further values in the Yaml.

xzyfer:
We've done this for new features. This has come up before, and the LibSass team rallied to keep the current workflow because it gives us a way to gauge how much work we have to do at a glance. Another argument for keep the current libsass-todo-* folder has been that noone has the time to move the specs into the appropriate place. This would require someone to dedicate significant time to organising the specs into logical groups.

This is arguably less relevant in a post LibSass 3.4 world. All new 3.5 specs have been added as this issue suggests.

mgreter:
Not sure about this, as it obfuscates which issues are still open for libsass. IMO we must have at least some way to query the open todos before considering this. I really like that I can just browse those two folders to see which kind of issues are not solved yet. And I think it doesn't really make a difference if we move files or edit the options.yml, we need to make a commit either way (I would even argue that moving is easier and less error prone than editing options.yml).

nex3:
I'd also really like a way to query various stats about the number of tests with various sorts of configurations.

nschonni:
I'm thinking the report and other stats stuff might need to be split out of the [annotate](https://github.com/sass/sass-spec/blob/master/lib/sass_spec/annotate/cli.rb)

stof:
@nex3 @jathak I suggest closing this issue, as marked tests as TODO is handled by the options since years now.

PR #951, opened 2016-10-20
Use the options.yml file for the libsass issues

nschonni:
Took a rough shot at  #950, but without the full restructure. This just flattens the directory and uses the built in `annotate` functions in `sass-spec.rb` instead of moving the files around.


xzyfer:
My feeling is that this is jumping the gun on #950. I think need a more concrete plan before taking any action. Incremental changes could make things worse in the short term.

There are also likely to be some changes to sass-spec soon that could have impact on this.

nex3:
Closing this out; I think the plan is solid, but we can't execute on it until we have better reporting infrastructure and this pull request has bit rotted anyway.

PR #958, opened 2016-11-01
Add an :only_on: annotation.

### LINKED ISSUE #946: Add a way to declare a spec for a specific implementation
Different implementations may have different outputs for the same input, and all of those outputs may be valid. There's not a good way to represent this in sass-spec right now. You can _ignore_ specs for particular implementations, but that's not future-proof: in order to say "this is the output for libsass only", you'd need to explicitly list all other implementations.

When a spec fails in interactive mode, we could also provide an option that says "Migrate copy to pass on #{implementation}." This would add a copy of the spec that passes only on the current implementation, and ignore the base copy for that implementation.

nex3:
Closes #946

### DIFF
--- a/lib/sass_spec/test_case_metadata.rb
+++ b/lib/sass_spec/test_case_metadata.rb
@@ -35,7 +35,7 @@ module SassSpec
       existing_opts
     end
 
-    ACCUMULATED_OPTIONS = [:todo, :warning_todo, :ignore_for, :ignore_warning_for]
+    ACCUMULATED_OPTIONS = [:todo, :warning_todo, :ignore_for, :ignore_warning_for, :only, :warning_only_on]
 
     attr_reader :options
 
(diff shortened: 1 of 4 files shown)


lib/sass_spec/test_case_metadata.rb
```diff
@@ -35,7 +35,7 @@ def self.merge_options(existing_opts, new_opts)
       existing_opts
     end
 
-    ACCUMULATED_OPTIONS = [:todo, :warning_todo, :ignore_for, :ignore_warning_for]
+    ACCUMULATED_OPTIONS = [:todo, :warning_todo, :ignore_for, :ignore_warning_for, :only, :warning_only_on]
```
nschonni: Think `:only,` is supposed to be `:only_on,`

nex3: Good catch. Fixed in 36c5a29.

PR #962, opened 2016-11-02
Support TODOs in interactive mode.

nex3:


nschonni:
It would probably be good to update the README after you finish changing all these. Or maybe add a docs folder

nex3:
As far as I can tell, there isn't really any pre-existing documentation for anything around `--interactive` mode that I could update, and I don't really have time at the moment to write it from scratch.

nschonni:
This one would probalby be under https://github.com/sass/sass-spec#pending-todo-tests

nex3:
It already mentions it, even though it wasn't supported before:

> The `--interactive` mode will provide marking a test as pending for the current implementation as a remedy for many types of spec failures.

PR #964, opened 2016-11-08
Don't nest specs within other specs.

nex3:
Nested specs make it impossible to mark the parent spec as TODO or
ignore for an implementation without recursively marking its children as
well.


nex3:
@xzyfer @chriseppstein @nschonni Does this look reasonable to y'all?

nschonni:
Can't really tell what the structure is from the webview. I'll take a look when I get home.

nex3:
Basically, everywhere there was a spec like this:

```
spec/
  input.scss
  example_output.css
  nested/
    input.scss
    example_output.scss
```

I converted it to

```
spec/
  simple/
    input.scss
    example_output.scss
  nested/
    input.scss
    example_output.scss
```

There are a few other spec-specific tweaks, but that's the bulk of it.

nschonni:
OK, that makes sense. The only semi related thought I had was during the default switch, to just add the different outputs to the same folder.
```
spec/
  some-spec/
    input.scss
    expected_compact.scss
    expected_nested.scss
    expected_expanded.scss
    expected_expanded_dart.scss
```
A single options.yml could still controll what is enabled/disabled for a single folder

xzyfer:
:+1: the former is either legacy or bad form. We tend to convert them to the latter style as we stumble across them.

PR #997, opened 2016-12-11
Fix the expected output for str-slice().

nex3:


xzyfer:
I have a patch in LibSass (https://github.com/sass/libsass/pull/2241) please remove the todo for Libsass.

xzyfer:
I'm not sure I completely follow the issue. Can you please confirm what the output of the following should be?

```
a {
  test-11: str-slice('aöc', 0, -3);
  test-12: str-slice('aöc', 1, -3);

  test-16: str-slice('aöc', 0, -4);
  test-17: str-slice('aöc', 1, -4);

  test-18: str-slice('aöc', 2, -4);
  test-19: str-slice('aöc', 3, -4);
}
```

nex3:
@xzyfer `str-slice('aöc', 0, -3)` should both return `"a"`, because the `$end-at` parameter is inclusive and points to the third-from-last character, `a`. The same is true for `str-slice('aöc', 1, -3)`, since indices are 1-based and so the index 1 also points at `a`. The others should all return `""`.

xzyfer:
Sorry I'm not clear why these both return `a`.

```
test-11: str-slice('aöc', 0, -3);
test-16: str-slice('aöc', 0, -4);
```

See https://github.com/sass/sass-spec/pull/997/files#diff-2a1fce6f95db3f518285362f272d8e61R18

PR #999, opened 2016-12-12
Update specs for dp unit conversions.

nex3:
See sass/sass#2143

PR #1003, opened 2016-12-15
Mark dps conversion specs as todos for ruby-sass and dart-sass

xzyfer:
LibSass needs https://github.com/sass/sass-spec/pull/999 merged

### DIFF
--- a/spec/libsass/units/conversion/resolution/options.yml
+++ b/spec/libsass/units/conversion/resolution/options.yml
@@ -1,0 +1,5 @@
+---
+:start_version: '3.4'
+:todo:
+- dart-sass
+- ruby-sass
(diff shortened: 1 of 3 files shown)


xzyfer:
cc @nex3

spec/libsass/units/conversion/resolution/options.yml
```diff
@@ -0,0 +1,5 @@
+---
+:start_version: '3.4'
```
nex3: Why are you adding these start versions? I don't think it makes sense to have start versions for the oldest supported language version...

xzyfer: You're right

PR #1004, opened 2016-12-15
Fix and disable str-slice spec

xzyfer:
The spec was updated in #997 with incorrect output generated by a
big in Ruby Sass https://github.com/sass/sass/issues/2211.

This updates the expected output and disables the spec since all
implemenations are now failing.


xzyfer:
cc @nex3

PR #1006, opened 2016-12-15
Add todo spec test for libsass issue 2247

mgreter:
https://github.com/sass/libsass/issues/2247


mgreter:
mixed up pr number with issue nr

PR #1020, opened 2016-12-21
Fix multiline parent

maciej-ka:
https://github.com/sass/sass/pull/2157

### DIFF
--- a/spec/libsass-closed-issues/issue_1061/expected_output.css
+++ b/spec/libsass-closed-issues/issue_1061/expected_output.css
@@ -1,3 +1,3 @@
-a.div, a.span {
-  display: block;
-}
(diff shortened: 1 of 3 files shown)


nschonni:
Is this for a 3.4 or 3.5 change? I know this is for an upstream sass issue, so this should actually change the option.yml to exclude sass rather than breaking the runs for the other implementations. 👍 to adding a new test to cover you're functionality though

nex3:
This should probably be marked TODO for any implementations that intend to support nice selector newlines (which I believe is all of them at the moment).

nex3:
Closing because https://github.com/sass/sass/pull/2157 is on ice.

PR #1024, opened 2016-12-25
Add todo spec test for libsass issue 2243

mgreter:
https://github.com/sass/libsass/issues/2243


xzyfer:
Spec https://github.com/sass/sass-spec/pull/1022

PR #1039, opened 2016-12-30
Fix some missing 3.5 deprecation warning

xzyfer:
I'm working on getting LibSass running against 3.5 and notice some
specs for `call` with a string didn't have deprecation warnings.

I wasn't able to use sass-spec to generate the warning either. Looks
like there is a bug with generate. I haven't investigated further.


xzyfer:
I've logged #1040 to track the issue with generate.

PR #1044, opened 2016-12-31
Activate 3.5 call with string deprecation specs for LibSass

xzyfer:
Activates ~50 specs that were failing due to this missing deprecation notice. Most of these were deactivated in https://github.com/sass/sass-spec/pull/1042 so that we could get LibSass master building against 3.5 language specs.


xzyfer:
Superseded.

PR #1047, opened 2017-01-06
Activate some passing libsass spec tests

mgreter:


xzyfer:
Please don't activate these. I'm aware of them, but last I checked they were passing for the wrong reasons i.e. the wrong error message. I think more investigation is needed.

PR #1078, opened 2017-02-04
Activate bunch of passing libsass tests

mgreter:


xzyfer:
Update the LibSass 3.4 release with commits required to passes the new specs.

PR #1109, opened 2017-05-09
Add Ruby 2.4 to Travis.

junaruga:
Right now the Travis test is run on Travis default Ruby (=2.3.1), seeing below Travis test on latest master branch.
I want to add Ruby 2.4 to check the behavior.

https://travis-ci.org/sass/sass-spec/builds/230385467

```
$ ruby --version
ruby 2.3.1p112 (2016-04-26 revision 54768) [x86_64-linux]
```

Because I faced below error for `sass`. It is failed for only Ruby 2.4.1.
https://github.com/sass/sass/issues/2279

I want to know if it is issue of `sass-spec` or `sass`.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,13 +17,22 @@ matrix:
   include:
     - gemfile: Gemfile
       env: LANGUAGE_VERSION=3.4 IMPL=ruby-sass
+      rvm: 2.3.4
     - gemfile: Gemfile.sass_3_5
       env: LANGUAGE_VERSION=3.5 IMPL=ruby-sass
+      rvm: 2.3.4
     - gemfile: Gemfile.sass_4_0
       env: LANGUAGE_VERSION=4.0 IMPL=ruby-sass
+      rvm: 2.3.4
+    - gemfile: Gemfile.sass_4_0
+      env: LANGUAGE_VERSION=4.0 IMPL=ruby-sass
+      rvm: 2.4.1
     - env: LANGUAGE_VERSION=3.4 IMPL=libsass COMMAND="../sassc/bin/sassc" GITISH=3.4-stable
+      rvm: 2.3.4
     - env: LANGUAGE_VERSION=3.5 IMPL=libsass COMMAND="../sassc/bin/sassc" GITISH=master
+      rvm: 2.3.4
     - env: LANGUAGE_VERSION=4.0 IMPL=dart-sass
+      rvm: 2.3.4
 
 before_install:
   - if ./tools/skipped-for-impl.sh; then exit 0; fi


junaruga:
I got the error for `sass-spec` on Ruby 2.4.1.
What we have to fix is `sass-spec`, not `sass`.

https://travis-ci.org/sass/sass-spec/builds/230444996

```
  1) Failure:
SassSpec::Test#test__spec/libsass/charset [/home/travis/build/sass/sass-spec/lib/sass_spec/test.rb:269]:
expected did not match output.
--- expected
+++ actual
@@ -1,6 +1,6 @@
 # encoding: ASCII-8BIT
 "@charset \"UTF-8\";
 div {
-  content: \"\xC3\xB8\xC3\xA1\xC3\xA9\xC3\xAD\xC3\xB3\xC3\xBA\xC3\xBC\xC3\xB1\xC2\xBF\xC3\xA9\xC3\xA0\xC5\xA4\xC7\x85\xC7\x82\xC9\x8A\xC9\xB1\xCA\xAD\xCA\xAC\xD1\xAA\xD2\x88\xDD\x93\";
+  content: \"\xC3\x98\xC3\x81\xC3\x89\xC3\x8D\xC3\x93\xC3\x9A\xC3\x9C\xC3\x91\xC2\xBF\xC3\x89\xC3\x80\xC5\xA4\xC7\x84\xC7\x82\xC9\x8A\xE2\xB1\xAE\xCA\xAD\xCA\xAC\xD1\xAA\xD2\x88\xDD\x93\";
 }
 "
```

junaruga:
Memo: Maybe reading this file in the failed test.
https://github.com/sass/sass-spec/blob/master/spec/libsass/charset/expected_output.css

Here is the failed line.
https://github.com/sass/sass-spec/blob/master/lib/sass_spec/test.rb#L269

And when I did put debug print,

```
  def handle_output_difference!
...
+    puts "handle_output_difference expected: #{@test_case.expected.encoding}"
+    puts "handle_output_difference actual: #{clean_output.encoding}"
+
     assert_equal @test_case.expected, clean_output, "expected did not match output" <== failed here.
   end
```

The result is like this.

```
handle_output_difference expected: ASCII-8BIT
handle_output_difference actual: ASCII-8BIT
```

Ref. https://travis-ci.org/sass/sass-spec/jobs/230462442

junaruga:
Maybe `sass-spec` wants to read a file as `ASCII-8BIT` from
https://github.com/sass/sass-spec/blob/master/lib/sass_spec/test_case.rb#L161-L175
  #input, #expected method.

But `assert_equal` want to compare expected actual objects as "Encoding.default_external".
So, I suppose that the solution is no using `assert_equal`.

junaruga:
Ah sorry.
The problem is why `@test_case.expected` and `clean_output` are difference for only Ruby 2.4.1.
The condition to show `# encoding: ASCII-8BIT` is not related to the issue.

junaruga:
Maybe around below code.

https://github.com/sass/sass/blob/stable/lib/sass/script/functions.rb #to_upper_case
https://github.com/sass/sass/blob/stable/lib/sass/script/value/string.rb ::value

junaruga:
Maybe this issue ( https://github.com/sass/sass-spec/pull/1109#issuecomment-300243195 ) for Ruby 2.4 was fixed by https://github.com/sass/sass/issues/2279 .

junaruga:
OK all the test was passed.

junaruga:
@nex3 hello, can you merge this PR? Adding Ruby 2.4 to Travis is useful.

nex3:
I'm not really the primary maintainer for sass-spec... @nschonni or @xzyfer, can one of you take a look?

nschonni:
I think that it's better to keep the Ruby version matrix in the Sass repo (which you've already done 👍 ) since none of the other implementations rely on it for anything but the runner.

PS: I think it's better to use the MAJOR.MINOR instead of the MAJOR.MINOR.PATCH in the matrix since then we don't need to bump for those releases.

junaruga:
> I think that it's better to keep the Ruby version matrix in the Sass repo (which you've already done 👍 ) since none of the other implementations rely on Ruby for anything but the runner.

So, if you do not want to manage Ruby version in`sass-spec` repo, I am fine to close this ticket.

> PS: I think it's better to use the MAJOR.MINOR instead of the MAJOR.MINOR.PATCH in the matrix since then we don't need to bump for those releases.

This is about `sass` repo? Yeah, It may be better.
Though Travis always does not support latest MAJOR.MINOR version. They supports it later.
See https://github.com/travis-ci/travis-ci/issues/7077

junaruga:
Thanks for checking my PR!
I would close it.

PR #1132, opened 2017-06-02
Use precision 10 for specs by default.

nex3:
See sass/sass#1122


nex3:
@xzyfer It looks like changing the default precision is causing libsass to produce different colors for some reason. Can you take a look?

nschonni:
Think the upstream issue is https://github.com/sass/libsass/issues/2294

nex3:
@mgreter Changing the precision that's used for Sass arithmetic shouldn't affect the colors from color functions, but it does. I think this is a LibSass bug, and it's blocking my ability to land this pull request without marking tons of specs as TODO for LibSass. Can you look into fixing that?

nex3:
@mgreter ping?

nex3:
Since there hasn't been any response from the LibSass folks on this, I guess I'll just mark those specs as TODO.

PR #1133, opened 2017-06-02
Mark a reference combinator spec as ending at 3.4.

nex3:
Skip ruby-sass


xzyfer:
It's worth noting that remove support for reference combinators with adding support for the `>>>` combinator will prevent Sass working with Angular. More details in 

https://github.com/sass/libsass/issues/2378 and https://github.com/sass/sass/issues/2266

nex3:
According to [this Chrome platform status page](https://www.chromestatus.com/features/6750456638341120):

> Shadow-Piercing descendant combinator, '/deep/' (aka '>>>'), including '::shadow' pseudo elements, are being deprecated.

However, [the CSS Scoping 1 spec](https://drafts.csswg.org/css-scoping-1/#deep-combinator), which was updated today, still includes `>>>`. @tabatkins, can you provide some more context here? Is `>>>` deprecated or not? Is it supported in Chrome? Will it continue to be?

tabatkins:
Deprecated in style sheets; it's no longer allowed there per spec. Still allowed in querySelector()/etc.

nex3:
In that case, it sounds like Angular should no longer be using either `/deep/` or `>>>` in their style sheets, so we shouldn't feel obligated to support them.

tabatkins:
Yup, exactly.

luchillo17:
@nex3 Mind if i ask, if they remove such selectors for nested items, with what should they replace it?

tabatkins:
Whatever they're already using to make it work in browsers other than Chrome, presumably.

PR #1140, opened 2017-06-06
Add specific libsass error messages

mgreter:
Messages differ as catched by perl-libsass.
Not sure why this does not fail official spec runner!?

### DIFF
--- a/spec/sass_3_5/functions/join/error/positional/error-libsass
+++ b/spec/sass_3_5/functions/join/error/positional/error-libsass
@@ -1,0 +1,3 @@
+Error: wrong number of arguments (5 for 4) for `join'
+        on line 2 of /sass/spec/sass_3_5/functions/join/error/positional/input.scss
+  Use --trace for backtrace.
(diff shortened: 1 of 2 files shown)


xzyfer:
Tests don't fail because of they're configured to ignore warnings see


- [here](https://github.com/mgreter/sass-spec/blob/2dbe449933d1890252d1f108e04d50dc29402b24/spec/sass_3_5/functions/join/error/positional/options.yml)
- [here](https://github.com/mgreter/sass-spec/blob/2dbe449933d1890252d1f108e04d50dc29402b24/spec/sass_3_5/functions/join/error/positional_and_named/options.yml)

mgreter:
Message seems to be an error and not a warning?
IMO this should not be skipped by "warning_todo"!?

mgreter:
And sass-spec master seems broken :(

xzyfer:
I think warning_todo predates implementation specific error messages. Master is currently broken for Sass 4 IIRC. It's safe to ignore.

@nex3 can you please arrange for Sass 4 to be updated with the latest master?

mgreter:
OK, perl-libsass only ignored warnings on the "warning_todo" flag.
FWIW: `s/^(?:DEPRECATION )?WARNING(?:[^\n]+\n)*\n*//gm;`
These two messages were the only one failing.
So this PR is still valid, even if sass-spec does not need it yet.

xzyfer:
You should also delete the yaml files also since the ignores are no longer required?

nex3:
@xzyfer Updated sass master.

xzyfer:
@nex3 thanks!

PR #1154, opened 2017-07-06
chore: use newer Bash command substitution syntax

nschonni:
https://github.com/koalaman/shellcheck/wiki/SC2006


nschonni:
@xzyfer you'll need to merge since I don't have the commit bit

xzyfer:
Sorry for the delay I was playing email tag to fix the permission issue. You should now have the powers @nschonni

PR #1160, opened 2017-08-08
Add specs for issue 2462

asottile:
Refs sass/libsass#2462


xzyfer:
Ruby Sass 3.4 failures are unrelated.

@nex3 are you able to take a look at those failures when you have time?

nex3:
We should probably just delete 3.4 specs at this point, since it's no longer actively maintained...

xzyfer:
It makes sense that any 3.4 specs with deprecated behaviour be removed now

PR #1164, opened 2017-08-24
Add tests for LibSass issue 2472

mirisuzanne:
https://github.com/sass/libsass/issues/2472


xzyfer:
This spec appears to fail on all ruby was versions

mgreter:
Seem borked in other specs than this, experienced the same in perl-libsass spec runner since a few months.

mgreter:
Btw @mirisuzanne I guess this case would idealy need more than one spec test to cover all cases.

mirisuzanne:
@mgreter how would you like it broken out? Right now I have the two primary errors tested in that single file/folder. I could split those into two different tests? I think those are the only two fail-cases I've found - all my other experiments/hacks were successful. Are there other aspects of the problem that I'm missing?

mgreter:
@mirisuzanne if you're sure there only two cases, than that would be a good start. I would also include the work-around case, so it does not regress in the future. Thanks!

nex3:
@mirisuzanne are you planning on iterating on this?

mirisuzanne:
@nex3 I can take a look in the morning, and add the work-around case, if that's helpful.

nex3:
I don't know if the work-around is necessary, but getting this split into multiple test files would be helpful.

mirisuzanne:
@nex3 Hope I did that right!

mirisuzanne:
Oh, then I should really be using a different function in the first test – one that doesn't need to error on positional values. I could also capture the functions with `get-function` if that helps.

nex3:
@mirisuzanne Are you sure that other functions *did* error?

mirisuzanne:
@nex3 Yes. At the time, this was an error I hit when I used `call` on any internal Sass functions. Those two functions were only chosen as examples. I haven't tested more recently.

nex3:
@mirisuzanne Can you add a new test that would have triggered the argument-number error for a function other than `*-color()`?

mirisuzanne:
@nex3 At this point, I'm not sure I understand the proper structure for the tests/errors, and I don't have a way to check that it actually triggers the error (Not sure what version I'd have to test against) – but this follows the same structure without using `*-color()`:

```scss
// not enough arguments for `join' on line x at column y
@function my-join(
  $list,
  $args...
) {
  @return call('join', $list, $args...);
}

.test {
  too-few: my-join((a, b), (c, d));
}
```

And should output:

```scss
.test {
  too-few: a, b, c, d;
}
```

nex3:
Closing this out because I've lost all context on it at this point. Feel free to open a new PR for it if you feel it's important.

PR #1167, opened 2017-09-22
Remove specs that are only relevant for Sass 3.4

nex3:
See comments on #1160


xzyfer:
These CI failures have been happening intermittently recently. I don't know enough about ruby/bundler to resolve it.

PR #1206, opened 2018-01-25
Added Spec test for import using index if dir given

john681611:
Ref to: https://github.com/sass/sass/pull/2456

Not sure if i need to add more

CI will fail because it needs the referenced PR to be merged

Skip dart-sass
Skip ruby-sass


nex3:
I think we need more thorough specs here. I'd like to see:

* Importing `index.scss` (in `spec/directives/import/index/scss`).
* Importing `index.sass` (in `spec/directives/import/index/sass`).
* Importing `_index.scss` (in `spec/directives/import/index/partial`).
* `dir.scss` takes precedence over `dir/index.scss` (in `spec/directives/import/index/file_takes_precedence`).
* Importing `dir.scss/index.scss` doesn't work (in `spec/directives/import/index/dir_dot_scss`).
* Importing `dir.foo/index.scss` does work (in `spec/directives/import/index/dir_dot_foo`).

Also, if you add

```yaml
---
:todo:
- libsass
```

to `spec/directives/import/index/options.yml`, that will skip these tests on LibSass until it implements this feature.

> CI will fail because it needs the referenced PR to be merged

If you add "Skip dart-sass" and "Skip ruby-sass" at the end of the pull request message, the CI will skip those tests (although this is a bit flaky).

spec/directives/import/index/dir_dot_scss/error
```diff
@@ -0,0 +1 @@
+Is a directory @ io_fread - /sass/spec/directives/import/index/dir_dot_scss/dir.scss
```
nex3: I think Ruby Sass should act as though it failed to find the file if it's a directory, rather than trying to read it and forwarding the IO error. That is, it should check `File.file?` rather than `File.exist?`.

john681611: So it this work in the ruby sass repo or here?

nex3: The behavior should be changed in the Ruby Sass repo and the expectation should be changed here.

nex3:
Looks good! Waiting on the other two PRs to merge.

PR #1228, opened 2018-03-14
Add tests for arithmetic edge case

mgreter:


spec/basic/15_arithmetic_and_lists/input.scss
```diff
@@ -29,4 +29,6 @@ div {
   r: 16em * 4;
   s: (5em / 2);
   t: 1 + (2 + (3/4 + (4/5 6/7)));
+  u: (1 / 0); // Infinity
+  v: (0 / 0); // NaN
```
nex3: I'd really like to move to more self-descriptive test cases, rather than ones that are labeled arbitrarily using numbers or letters. For example, these could be written

```css
  one-over-zero: (1 / 0);
  zero-over-zero: (0 / 0);
```

This makes it easier to tell what the test cases is when looking up the expected output, and in more complex cases to determine what the point of the test is.

spec/basic/15_arithmetic_and_lists/input.scss
```diff
@@ -29,4 +29,6 @@ div {
...
 }
```
nex3: I don't understand why these test cases were added to a directory called "arithmetic and lists" when they don't seem to have anything to do with lists.

PR #1229, opened 2018-03-15
Allow mod 0

nex3:
To match the behavior of dividing by zero, this should produce NaN
rather than failing.

See also #1228

[skip ruby-sass]
[skip dart-sass]


mgreter:
FWIW: ruby sass 3.5.6 also crashes for:
```scss
  foo { bar: $foo / #000000; }
```
But since color math is deprecated I guess there is no intention to fix this ...
Therefore I also guess we can keep libsass error until https://github.com/sass/sass/issues/2144

nex3:
I'm pretty confident no one will run into that issue before Ruby Sass hits end-of-life :stuck_out_tongue_closed_eyes:.

PR #1230, opened 2018-03-16
Add specs for nested media queries

nex3:
See sass/libsass#2425
See sass/dart-sass#244

[skip dart-sass]


xzyfer:
Does this mean the Ruby Sass implementation was already correct?

nex3:
Yes.

xzyfer:
Thanks. Looks like it was just a typo on in LibSass then.

PR #1245, opened 2018-04-07
Add color arithmetic specs.

binki:
These specs are based on ruby-sass current behavior (which people likely
rely on) and documentation as read in the SASS_REFERENCE. It checks that
addition, subtraction (with a deprecation warning), division, and
multiplication function correctly.


binki:
It looks like libsass is missing the deprecation message that ruby-sass has. Could someone who has a better idea of how that should work advise me on the next step?

Thanks.

nex3:
Color arithmetic specs already exist in `spec/basic/16_hex_arithmetic`. I don't think we need any more specs.

PR #1249, opened 2018-04-16
Add spec for invisible custom property

xzyfer:
This highlighted a bug in LibSass.

@nex3 do think custom properties be invisible? My feeling is that they should be output regardless.

Note this currently causes an error in Ruby Sass

```
NoMethodError: undefined method `size' for nil:NilClass
  Use --trace for backtrace.
```

[skip ruby-sass]
[skip libsass]

### DIFF
--- a/spec/css/custom_properties/simple/input.scss
+++ b/spec/css/custom_properties/simple/input.scss
@@ -9,6 +9,7 @@
 
   // The whitespace here DOES count as a token and needs to be preserved.
   --empty: ;
+  --empty-no-space:;
 
   // Single-line comments are not supported in variables.
   --single-line: // (


nex3:
`--foo:;` should be a syntax error. [The spec](https://drafts.csswg.org/css-variables/#syntax) requires a `<declaration-value>` production, which requires at least one token of some kind. Ruby Sass's behavior is obviously still incorrect.

There's still a question of what to do for `--foo:#{""};`. We could either drop the property or error out when we detect that the resolved value is empty; my instinct is to do the latter, since it's almost always going to be an error, and I don't want people to end up writing `--foo:#{...}` instead of `--foo: #{...}` on purpose to activate the dropping behavior.

PR #1271, opened 2018-07-25
Add specs for media range support

nex3:
See sass/sass#1864

[skip dart-sass]


nex3:
> it looks like this is the corresponding accepted language proposal https://github.com/sass/language/blob/master/accepted/media-ranges.md ?

Yeah, I've set up all the issues with proposals to link to those proposals prominently.

PR #1273, opened 2018-07-26
Enable LibSass specs for hex colors with alpha

xzyfer:
[skip libsass]

### DIFF
--- a/spec/values/colors/alpha_hex/deprecation/options.yml
+++ b/spec/values/colors/alpha_hex/deprecation/options.yml
@@ -1,3 +1,4 @@
 ---
 :ignore_for:
 - ruby-sass
+- libsass
--- a/spec/values/colors/alpha_hex/initial_digit/options.yml
+++ b/spec/values/colors/alpha_hex/initial_digit/options.yml
@@ -1,4 +1,2 @@
 ---
 :start_version: '3.6'
-:todo:
-- libsass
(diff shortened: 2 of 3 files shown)


nex3:
@xzyfer Can you delete [LibSass-specific files](https://github.com/xzyfer/sass-spec/blob/70a653dade9d9cac5508a3fcc6dd1508c70d9323/spec/values/colors/alpha_hex/deprecation/error-libsass) when marking a spec as ignored for LibSass?

xzyfer:
Yep, sorry

nex3:
Also, it looks like this broke Travis: https://travis-ci.org/sass/sass-spec/builds/408503694

xzyfer:
Looks like Travis may not be pulling in the latest sassc. Libsasy defaults
to precision 5 where it looks like dart sass and ruby sass default to 10

nex3:
The default precision shouldn't matter; the engine sets the precision explicitly.

xzyfer:
I'm not sure what I'm missing then.

https://github.com/sass/sass-spec/blob/78b0f9aca1b858b6f816b90c501248c82a3a5847/lib/sass_spec/test_case.rb#L79-L81

The default appears to be 5. I don't it being set otherwise in any relevant options.yml. Yet the output in the specs appears to use precision 10.

nex3:
Since no other implementation is generating the wrong precision here, it seems like it might be a LibSass bug. Can you roll back this PR while you investigate to get the build green?

nex3:
Oh I see what's happening here--I generated the expected test cases using Dart Sass, which always uses precision 10.

Issue #1001, opened 2016-12-12
Add specs for built-in functions


nex3:
All the built-in functions need thorough specs that exercise their edge and error cases. There isn't currently a great structure in place for this; some functions have some specs scattered around the repo, but there should be a uniform location and we should make sure *all* functions are tested.

Ruby Sass has [good test coverage for functions](https://github.com/sass/sass/blob/stable/test/sass/functions_test.rb), so the best way to go about this is probably porting those. I suggest the following structure:

```
spec/
'-- functions/
    |-- color/
    | ...
    '-- number/
        |-- round/
        | ...
        '-- max/
            |-- success/
            |   |-- input.scss
            |   '-- expected_output.scss
            '-- error/
                |-- type_error_arg1/
                |   |-- input.scss
                |   '-- expected_output.scss
                '-- type_error_arg2/
                    |-- input.scss
                    '-- expected_output.scss
```

For functions with more complex modes of operation, we could replace `success/` with multiple test cases.

nex3:
We should also retrofit the existing `spec/colors/`, `spec/number-functions/`, and `spec/selector-functions/` into this format.

For the time being, I'm adding this structure lazily when I need to create a spec for a function.

stof:
@nex3 Can this be considered done now ?

PR #1300, opened 2018-11-07
Add full specs for rgb(), hsl(), rgba(), and hsla()

nex3:
Partially addresses #1001

[skip dart-sass]


nex3:
@jmesserly The GitHub UI sucks for changes like this, so to thoroughly review you'll probably want to check out the PR branch locally and look at the various input/output expectations from there. The success tests are the most important ones; it's fine to just make sure that all the error tests should fail without closely inspecting the error output for each implementation.

PR #1304, opened 2018-11-14
Rename expected_output.css to output.css

nex3:
There's no need for this to be so long, or so different from the input
file name.


nex3:
@nschonni Our of curiosity, why doesn't Node Sass run the official spec runner? I'm planning a number of changes to it, and as it stands all of those will have to be mirrored in Node Sass's custom runner.

PR #1307, opened 2018-11-15
Add spec for LibSass issue 2687

xzyfer:
[skip libsass]


xzyfer:
This broken master. The fix will be #1309.

PR #1308, opened 2018-11-16
Refactor some spec infrastructure

nex3:


lib/sass_spec/cli.rb
```diff
@@ -103,13 +103,6 @@ def self.parse
         options[:only_output_styles] << style
       end
 
-      opts.on("-r SPEC_DIR", "--root SPEC_DIR", "Root directory for the specs. ",
```
nschonni: Not sure spec directory is used by some of the libsass variants when skimming through https://github.com/search?p=1&q=org%3Asass+sass-spec&type=Code

nex3: I see some stuff about configuring the location of the sass-spec gem itself, but nothing about using the --root flag.

lib/sass_spec/runner.rb
```diff
@@ -42,24 +42,7 @@ def run
       Minitest.reporter = Minitap::TapY
     end
 
-    result = Minitest.run(minioptions)
-
-    if @options[:run_todo]
```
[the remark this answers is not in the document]
nex3: It's a big pain in the ass to support, so it might be worth altering workflows for. You can still find passing TODO tests using `--interactive --probe-todo`.

lib/sass_spec/cli.rb
xzyfer: I don't have any insight into how our wrappers do their tests. We'd have to ask them cc @bolandrm @asottile @mgreter

xzyfer: Also @scottdavis

asottile: libsass-python doesn't invoke `sass-spec` currently (mostly leaning on `libsass` itself there) so this change is fine from our side!  Thanks for checking in :)

mgreter: perl-libsass has its own spec-runner at https://github.com/sass/perl-libsass/blob/develop/t/99_sass_specs.t

nex3: @mgreter How hard would it be to move to the Ruby test runner (or, as libsass-python does, rely on LibSass's own test coverage)? I'm planning some refactors that will likely be challenging to port.

mgreter: Not possible since it would break the whole concept of cpan tests which mostly doesn't have python installed and also relies sometimes on GCC 4.4 : http://matrix.cpantesters.org/?dist=CSS-Sass

nex3: Well, I guess consider this a heads-up that there will be some pretty substantial changes in test format coming soon.

mgreter: OK, not that I like to rewrite it for the third time ...

nex3: That's why I was hoping it could re-use the Ruby runner :disappointed:. Another solution would be to just not run the language specs, since it's just a wrapper and LibSass already tests its language implementation thoroughly.

mgreter: I understand, but the cpan test matrix actually did uncover a few quirks here and there already which weren't catched by libsass CI before and it's always a good feeling when it passed the full matrix.

nex3:
It sounds like the major wrappers either use custom runners or don't run sass-spec at all, so I don't think this will break them. @nschonni or @xzyfer can I get an approval?

Issue #1297, opened 2018-10-20
Stop using language versions for LibSass


nex3:
Currently, LibSass runs specs on Travis against version 3.5 and 3.6 of the language. With the deprecation of Ruby Sass, we're also deprecating the idea of versions of the Sass language. Implementations now just support certain sets of features instead, which we're representing in sass-spec as specs that are or are not marked as TODO. LibSass specs should move to this model; in other words, they should be able to run in 4.0 mode and just have any features that aren't implemented yet skipped as TODO.

xzyfer:
LibSass specs have been updated in #1309

nex3:
@xzyfer is there a reason this is still open, or can I close it out?

xzyfer:
I left it open because ruby sass was still using language versions. Wasn't
sure if that was going to be removed also.

nex3:
Removing versions entirely is tracked by #1223; this issue is just for LibSass to stop using them.

PR #1309, opened 2018-11-16
Remove language versions for LibSass

xzyfer:
LibSass 3.5-stable is done. Always run LibSass master.

See #1297


xzyfer:
TODO specs fixed

PR #1310, opened 2018-11-16
Add specs for attribute selectors that begin with "--"

nex3:
See sass/dart-sass#527

[skip dart-sass]


nschonni:
I don't think `--` is a valid identifier https://developer.mozilla.org/en-US/docs/Web/CSS/custom-ident#Forbidden_values
A single one is, but not 2

nex3:
These tests are for identifiers that *begin with* `--` (although in point of fact, [the spec](https://drafts.csswg.org/css-syntax-3/#ident-token-diagram) considers any sequence of hyphens longer than one a valid identifier).

nschonni:
I was going by 
> Moreover, the first character must not be a decimal digit, nor a hyphen (-) followed by a decimal digit or another hyphen.

But maybe it is valid in the latest draft

PR #1313, opened 2018-11-25
Enable LibSass specs for @content args

xzyfer:
[skip libsass]

### DIFF
--- a/spec/mixin/content/arguments/receiving/output-libsass.css
+++ b/spec/mixin/content/arguments/receiving/output-libsass.css
@@ -1,0 +1,28 @@
+with-defaults nothing-passed {
+  arg1: value1;
+  arg2: value2;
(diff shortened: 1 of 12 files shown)


spec/mixin/content/arguments/receiving/output-libsass.css
```diff
@@ -0,0 +1,28 @@
...
+with-splat nothing-passed {
+  positional: ();
+  keywords: ();
+}
+with-splat positional-passed {
+  positional: value1, value2, value3;
+  keywords: ();
+}
+with-splat named-passed {
+  keywords: (arg1: value1, arg2: value2);
```
xzyfer: Changed some output due to an existing issue with inspect'ing arglists that isn't related to this feature.

Issue #1095, opened 2017-03-17
GitHub Readable spec format


nschonni:
This popped into my head so I figure I'd share it to see if made sense to anyone else. If Markdown files are used for the specs, then we would have a readable spec for people right on GitHub, it could then also easily be converted to a GH Pages site.
Here is a random set of specs I tried to pull into a prospective format https://gist.github.com/nschonni/44dad08ad6e6c9f26cec26200928f4fd

Could also add the metadata for linking the related GH issues too

stof:
@nex3 should this be closed now that HRX is used ?

saper:
@stof er, somewhat, but it's probably personal taste. I [don't know a lot about decisions behind the design](https://github.com/sass/sass-spec/pull/1484#issuecomment-567709171) of the HRX format, beyond it's [general goals](https://github.com/google/hrx/blob/870b087606f270096a083bbbf2bfd64409876d98/README.md#goals).

nex3:
I think this can be closed out. We're not planning on changing the spec format further, beyond changes to the HRX spec itself.

PR #1316, opened 2018-11-29
Change how we distinguish between failure and success tests

nex3:
We used to use the "status" file to distinguish between a test for a
case that should fail, and one that should succeed. If a test had a
"status" file with any status other than 0, it was considered to be a
failure test. Failure tests had empty "output.css" files and "error"
files that included their error messages. Success tests had no status
file, "output.css" files that contained their output, and "error"
files that included any warnings they emitted.

This change gets rid of "status" files entirely and determines whether
a test fails based on whether it has an "output.css" file versus an
"error" file. All "output.css" files for error tests have been
deleted, and "error" files for successful tests have been renamed to
"warning" instead, which more accurately encapsulates their purpose.


nex3:
cc @mgreter, since the Perl wrapper also has a custom sass-spec runner

nschonni:
Was there other errors than "65" for the bad exit? If there was, the alternate I would suggest would maybe be to use the options.yml to store the status code instead. This would potentially mean a lot more files 😝 

PS: If we're ripping this all up anyway, what do you think of #1095

nex3:
> Was there other errors than "65" for the bad exit? If there was, the alternate I would suggest would maybe be to use the options.yml to store the status code instead. This would potentially mean a lot more files

As far as I can tell, we've *never* tested the actual status code against what implementations produce. We just used the existence of the status file as a signal. And I think that's correct: the status code isn't part of the language specification, it's an executable feature that shouldn't be tested in this repo.

> PS: If we're ripping this all up anyway, what do you think of #1095

I think the idea is sound, but that Markdown-based format would require a *lot* of boilerplate, it would be difficult to tool well, and it doesn't lend itself well to having multiple test cases per file (which would be *very* useful for error tests, which tend to have many small cases that can't be combined into a single file because they error out). That issue did inspire me to create https://github.com/google/hrx, though, which I think we can use to similar effect.

nschonni:
This one will require some changes in the node-sass runner (but shouldn't be that big a deal). Might need some input from the other libsass repos with separate test runners.
/cc @mgreter

nex3:
Generally speaking, I would like to limit the number of different implementations of the spec runner we consider critical rather than blocking changes to the main runner. The value we get from having the language test suite be maintainable and reviewable is very high (especially as progress on the language itself continues to ramp back up), and the diminishing returns of re-running every language test against different wrappers of the same underlying implementation are low.

nshahan:
From the perspective of reviewing PRs, reducing the number of files by eliminating status files seems like a good change.

PR #1322, opened 2018-12-04
Update spec output to match implementation

maximus8891:
Resolves: sass/libsass#1216


xzyfer:
There should be dart-sass output in output.css. LibSass specific output should be in output.libsass.css

xzyfer:
The latest change works also given that LibSass matches ruby sass

xzyfer:
Thanks

PR #1329, opened 2018-12-15
Use precision 10 for a few more specs

glebm:
Follow-up to #1132

I'm guessing these weren't caught because of warning todos (failures are ignored with probe-todo that runs on CI, even though it's just a warning todo).


xzyfer:
Thanks

PR #1336, opened 2019-01-29
Update specs for Dart Sass errors

nex3:
This adjusts specs for errors that of the form "expected ____" that
are detected at the end of the file, but should be printed as pointing
after the previous non-whitespace character.

[skip dart-sass]

### DIFF
--- a/spec/libsass-closed-issues/issue_2569/error-dart-sass
+++ b/spec/libsass-closed-issues/issue_2569/error-dart-sass
@@ -1,6 +1,6 @@
 Error: expected "@".
   ,
-5 |     opacity: 1;
-  |     ^
+4 |   } @else {
+  |            ^
   '
-  /sass/spec/libsass-issues/issue_2569/input.scss 5:5  root stylesheet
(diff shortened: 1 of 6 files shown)


nshahan:
Moving the pointer position helps here but in my opinion the error message makes some of these errors less actionable in dart-sass when compared to libsass. In general dart-sass is describing what could be the expected next character when libsass describes the unexpected character it encountered. Is this by design?

nex3:
No, it's not. That's a separate issue, though (https://github.com/sass/dart-sass/issues/577), which I'll address after this lands.

spec/libsass-closed-issues/issue_2569/error-dart-sass
```diff
@@ -1,6 +1,6 @@
 Error: expected "@".
```
nshahan: Probably need another misleading error message for the `expected "@"` too if you don't already.

nex3: Good catch. https://github.com/sass/dart-sass/issues/584

PR #1337, opened 2019-02-01
Update specs for Dart Sass errors

nex3:
This updates specs that used to say they expected a "{" to (more
sensibly) say they expect a "}".

[skip dart-sass]

### DIFF
--- a/spec/libsass-todo-issues/issue_1720/error-dart-sass
+++ b/spec/libsass-todo-issues/issue_1720/error-dart-sass
@@ -1,4 +1,4 @@
-Error: expected "{".
+Error: expected "}".
   ,
 3 | }
   |  ^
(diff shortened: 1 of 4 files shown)


spec/libsass-todo-issues/issue_1720/error-dart-sass
```diff
@@ -1,4 +1,4 @@
-Error: expected "{".
+Error: expected "}".
   ,
 3 | }
   |  ^
```
nshahan: The others look good. This one I'm not sure about. Why is the closing curly brace expected here?

nex3: This is a weird one. This is the input file:

https://github.com/sass/sass-spec/blob/4ccb052d391212d5f4b20d77473cbba96fd27813/spec/libsass-todo-issues/issue_1720/input.scss#L1-L3

The `}` on line 2 is commented out, so the `}` on line 3 closes the interpolation. The style rule, however, remains unclosed, so the parser looks for another `}`.

nshahan: Ah thanks. I looked right over the // when I glanced at the input.

Issue #1305, opened 2018-11-15
Stop testing output formats via sass-spec


nex3:
Different output formats aren't part of the Sass language, and so they probably shouldn't be tested as part of the language tests. Their presence in this repo also makes it harder to make large changes, and increases the CI iteration time (LibSass workers take about twice as long as other workers on Travis).

I suggest testing the specific differences in each output format in individual tests in the libsass repo, like [these Dart Sass tests](https://github.com/sass/dart-sass/blob/master/test/compressed_test.dart).

nex3:
@xzyfer Is there anything stopping us from just deleting these directories?

PR #1340, opened 2019-02-09
Convert specs to HRX format

nex3:
For the most part, this mechanically converts all spec directories
into individual HRX files. I've also gone through some error
directories and consolidated multiple small error specs into
individual HRX files to reduce clutter.

Output style specs have not been converted, since they're slated for
deletion anyway (#1305). It also doesn't convert specs that rely on
non-UTF-8 source encodings, since HRX only supports UTF-8.


glebm:
Is anyone working on an HRX crate for Rust? Would be great to have one for rsass.

nshahan:
@nex3 Are there any directories that you would like me to specifically look over? Otherwise I will probably just spot check a few to see that they have the expected transformation.

nex3:
@nshahan the changes in https://github.com/sass/sass-spec/pull/1340/commits/0d02dc2171e610c441895504171d8a6d840344b0 were done by hand, so those could probably use a human eye. Otherwise, spot-checking is fine.

nex3:
@glebm I don't know of anyone working on a Rust implementation yet.

PR #1344, opened 2019-02-16
Consolidate and flesh out @import load specs

nex3:


spec/directives/import/load.hrx
```diff
@@ -0,0 +1,185 @@
...
+
+<===>
+================================================================================
+<===> precedence/sass_before_css/options.yml
+---
+:todo:
+- libsass # sass/libsass#2699
+
+<===> precedence/sass_before_css/input.scss
+@import "other";
```
jmesserly: Probably not worth it or covered elsewhere, but I was wondering if it'd be worth testing the error case where both ".sass" and ".scss" are present (since it tests the other extension combos: css+sass and css+scss).

nex3: Errors are already covered [here](https://github.com/sass/sass-spec/blob/master/spec/directives/import/error/conflict.hrx).

PR #1349, opened 2019-02-21
Add @use member usage and load error tests

nex3:
[skip dart-sass]


nex3:
You only need to review [the commits after the merge](https://github.com/sass/sass-spec/pull/1349/files/821bd8a294eaf504dc7ddbe0648bad0198f56bf8..1bff1390dc6e9846c8a2f2d0c767c65ec2241e6f); everything else was already reviewed when it landed on `master`.

spec/core_functions/function_exists.hrx
```diff
@@ -149,3 +170,32 @@ Error: wrong number of arguments (2 for 1) for `function-exists'
...
+a {b: function-exists(member)}
+
+<===> conflict/other1.scss
+@function member() {@return from other1}
+
+<===> conflict/other2.scss
+@function member() {@return from other2}
+
+<===> conflict/error
+Error: Multiple global modules have a function named "member".
```
nshahan: I don't see where this error is coming from in dart-sass but could it include the names of the modules that have conflicting functions? It would make it easier to track down when many modules are used.

spec/directives/use/error/load.hrx
```diff
@@ -0,0 +1,314 @@
...
+3 | @use "other";
+  | ^^^^^^^^^^^^
+  '
+  /sass/spec/directives/use/error/load/conflict/extension/sass_and_scss/input.scss 3:1  root stylesheet
+
+<===>
+================================================================================
+<===> conflict/extension/scss_and_css/input.scss
+// This import can't be resolved because it could refer to either the ".sass" or
+// ".css" file.
```
nshahan: Same question here as I added to https://github.com/sass/dart-sass/pull/603. Might cause confusion on your second build?

spec/core_functions/function_exists.hrx
nex3: There's already a TODO for that: https://github.com/sass/dart-sass/blob/5a64dec4ec54ef4fc8fdf0ca0c16a83fec81463b/lib/src/async_environment.dart#L589

PR #1353, opened 2019-03-02
Update the README

nex3:
It now explains how to run specs for each major implementation, how
specs are laid out, and the available options. It also provides some
guidance for where to put new specs.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,178 +1,317 @@
-sass-spec
-=========
+# sass-spec
+
+### A cross-implementation [Sass][] test suite
+
+* [Running Specs](#running-specs)
+  * [Dart Sass](#dart-sass)
+  * [LibSass](#libsass)
+  * [Ruby Sass](#ruby-sass)
+* [Spec Structure](#spec-structure)
+  * [HRX](#hrx)
+  * [Specifying Warnings](#specifying-warnings)
+  * [Implementation-Specific Expectations](#implementation-specific-expectations)
+  * [Options](#options)
+    * [`:todo`](#todo)
+    * [`:warning_todo`](#warning_todo)
+    * [`:ignore_for`](#ignore_for)
+* [Organization](#organization)
+* [Interactive Mode](#interactive-mode)
+
+[Sass]: https://sass-lang.com
 
 [![Build Status](https://travis-ci.org/sass/sass-spec.svg)](https://travis-ci.org/sass/sass-spec)
 
-A test suite for Sass. The test cases are all in the `/spec` folder.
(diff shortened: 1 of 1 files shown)


nex3:
FYI @xzyfer, @mgreter, @glebm

README.md
```diff
@@ -1,178 +1,317 @@
...
+
+[Dart Sass]: https://sass-lang.com/dart-sass
+[the `sass` package]: https://npmjs.com/package/sass
+[Dart]: https://www.dartlang.org/
+
+```sh
+# If you already have a clone of the Dart Sass repo, you can use that instead.
+git clone https://github.com/sass/dart-sass
+(cd dart-sass; pub get)
+export DART_SASS_PATH=`pwd`/dart-sass
```
nschonni: I think `$pwd` is more portable than backticks, but i'm not a bash expert

README.md
```diff
@@ -1,178 +1,317 @@
...
+ul li {
+  list-style-type: none;
+}
+```
 
-Conversely, if you edit the Sass `Gemfile` and set `gem 'sass-spec',
-:path => "..."` then the Sass unit tests will run against your
-development version of sass-spec.
+HRX archives can contain directories, which means that they can also a single
+file can define multiple different This allows us to write many specs for the
```
nshahan: Unfinished sentence.

README.md
```diff
@@ -1,178 +1,317 @@
...
+Each spec is defined by a directory with an `input.scss` or `input.sass` file
+and either:
+
+* An `output.css` file, in which case the spec asserts that the Sass
+  implementation compiles the input to the output. These specs are known as
+  "success specs".
+* An `error` file, in which case the spec asserts that the Sass implementation
+  prints the error message to standard error and exits with a non-zero status
+  code when it compiles the input. These specs are known as "error specs".
+
```
nshahan: Maybe add a short callout with link to the "implementation-specific-expectations" section here.

README.md
nex3: Backquotes should work for any [`sh`-compliant shell](https://pubs.opengroup.org/onlinepubs/9699919799/utilities/V3_chap02.html#tag_18_06_03).

README.md
nex3: Done.

README.md
nex3: Wow, this whole paragraph is a mess. Done.

README.md
nschonni: Think I was coming more from https://github.com/koalaman/shellcheck#style but it looks like that s a style rather than bug rule https://github.com/koalaman/shellcheck/issues/364#issuecomment-97578330

PR #1357, opened 2019-03-11
Add partial specs for loading CSS with @use

nex3:
This doesn't specify the interaction between @use and @import, since
sass/language#75 is unresolved.

[skip dart-sass]

### DIFF
--- a/spec/directives/use/load.hrx
+++ b/spec/directives/use/load.hrx
@@ -1,3 +1,295 @@
+<===>
+TODO: spec out use + import once sass/language#75 is resolved.
+
+<===> css_order/use_into_use/input.scss
+@use "midstream";
+
+a {file: input}
+
+<===> css_order/use_into_use/midstream.scss
+@use "upstream";
+
+a {file: midstream}
+
+<===> css_order/use_into_use/upstream.scss
+a {file: upstream}
+
+<===> css_order/use_into_use/output.css
+a {
+  file: upstream;
+}
+
+a {
+  file: midstream;
+}
+
+a {
+  file: input;
+}
+
+<===>
+================================================================================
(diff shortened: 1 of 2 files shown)


spec/directives/use/load.hrx
```diff
@@ -0,0 +1,453 @@
...
+
+<===> import_order/import_below_rule/output.css
+@import "upstream.css";
+@import "midstream.css";
+@import "input.css";
+
+a {
+  file: midstream;
+}
+
```
nshahan: Why is  `@import "input.css";` above the `a {file: midstream}` in the output here?

spec/directives/use/load.hrx
```diff
@@ -0,0 +1,453 @@
...
+/* midstream comment before first import */
+
+@import "midstream1.css";
+
+/* midstream comment before second import */
+
+@import "midstream2.css";
+
+/* input comment before use */
```
nshahan: Why do the "before use" comments appear after the contents of the `@use` file?

spec/directives/use/load.hrx
nex3: CSS `@import`s are required to appear before any other rules (other than `@charset`) in the document. All these `import_order` tests ensure that we maintain that invariant while hewing as closely as possible to the original source order.

spec/directives/use/load.hrx
nex3: It would be *very* complicated to keep track of which comments appeared above or between which `@use` rules in the source text and use that to interpolate them appropriately into the CSS, especially once you get into multiple `@use`s of the same module. The next best thing is just to keep the contents of an individual file together as much as possible (with the caveat that CSS `@import`s are split out).

PR #1360, opened 2019-03-15
Add specs for CR, CR LF, and FF line endings

nex3:
See https://github.com/sass/dart-sass/issues/623
See https://github.com/sass/libsass/issues/2843

[skip dart-sass]


spec/css/comment.hrx
```diff
@@ -0,0 +1,133 @@
...
+<===>
+================================================================================
+<===> converts_newlines/scss/cr/options.yml
+---
+:todo:
+- libsass # sass/libsass#2843
+
+<===> converts_newlines/scss/cr/input.scss
+/* foo
 * bar */
```
nshahan: Is there a `\r` carriage return in this line that github isn't showing?

nex3: Yep.

PR #1361, opened 2019-03-16
Add specs for content-exists in control statements

xzyfer:
See https://github.com/sass/libsass/issues/2842

[skip libsass]


spec/core_functions/content-exists/controls/controls.hrx
```diff
@@ -0,0 +1,28 @@
...
+@mixin test-content-exists() {
+  @if content-exists() {
+    @content;
+  }
+  @else {
+    content-exists: false;
+  }
+}
+
+.should-be-true {
```
nex3: Nit: I'd call these classes `.with-content` and `.without-content`. It's already clear what the expected result is, since it's right there in the output file, so it's more useful to document what's meaningfully different about how they're invoked.

Actually, when you split these into separate specs, you should probably put `with_content` and `without_content` into the test names and just make these selector names `a`.

nex3:
Also, please don't mark a PR as "skip dart-sass" unless it's paired with a Dart Sass PR that's running against this PR. Otherwise, master will go red once this lands. I've removed the "skip" line in the description and re-run the tests.

stof:
@Awjin this PR should be closed now, as #1367 replaced it

PR #1363, opened 2019-03-20
Add a style guide for writing new specs

nex3:
I've tried to capture the best practices I've come up with while
writing tests in the HRX format, mixed with best practices from other
testing traditions and a few rules that mostly serve to just establish
a consistent style.

### DIFF
--- a/STYLE_GUIDE.md
+++ b/STYLE_GUIDE.md
@@ -181,7 +181,7 @@ files should be split apart if they become too large (see below).
 
 ### DON'T have HRX files longer than 500 lines or so
 
-Very large HRX files can be as difficult to naviate as hundreds of tiny files.
+Very large HRX files can be as difficult to navigate as hundreds of tiny files.
 Once an HRX file gets larger than about 500 lines, it's probably time to split
 it into multiple files. To do so:
 
@@ -631,9 +631,10 @@ a {
 
 </details>
 
-When writing a spec with multiple files, make the files other than the import
-partials unless there's a concrete reason not to. This helps make it clearer at
-a glance that they're additional input files, rather than expectation files.
+When writing a spec with multiple files, make the files (other than `input.scss`
+or `input.sass`) partials unless there's a concrete reason not to. This helps
+make it clearer at a glance that they're additional input files, rather than
+expectation files.
 
 ### DO name single imported or used files "other"
 
@@ -725,7 +726,7 @@ issue in the implementation in question tracking the feature the spec tests. If
 there isn't an issue yet, create one. Make sure to also link that issue back to
 the specs for it!
 
-[marking a spec as "todo"]: README.md#todo
+[marking a spec as `:todo`]: README.md#todo
 
 ## Documentation
 ### CONSIDER adding a comment explaining your spec
@@ -907,21 +908,18 @@ a b, a c {
 
 </details>
 
-The first file of each spec within an HRX file other than the first should begin with:
+Each subdirectory within an HRX file after the first should begin with [an HRX
+comment][] containing 80 `=` characters:
+
+[an HRX comment]: https://github.com/google/hrx/blob/master/README.md
 
 ```hrx
 <===>
 ================================================================================
 ```
 
-This is an HRX comment, and has no effect on the semantics of the file. It just
-serves to visually separate specs from one another.
-
-These comment separators should also be added before specs' parent directories
-if they're added separately, such as for an `options.yml` or [a `README.md`][]
-file.
-
-[a `README.md`]: #consider-adding-readme-md-files-to-parent-directories
+Because this is an HRX comment, it has no effect on the contents of the
+directory. It just serves to visually separate specs from one another.
 
 ### DO order files consistently
 
--- a/STYLE_GUIDE.md
+++ b/STYLE_GUIDE.md
@@ -856,7 +856,8 @@ Error: Expected token.
 
 HRX supports its own comment syntax, but that should never be used for specs.
 Instead [use Sass comments][] to document an individual file or spec and
-[`README.md` files][] to document entire directories.
+[`README.md` files][] to document entire directories. Sass comments provide more
+flexibility to provide documentation for specific portions of the test files.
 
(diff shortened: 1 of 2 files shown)


STYLE_GUIDE.md
```diff
@@ -0,0 +1,1245 @@
...
+  |              ^
+  '
+  /sass/spec/css/custom_properties/error/empty/input.scss 3:14  root stylesheet
+```
+
+</details>
+
+HRX supports its own comment syntax, but that should never be used for specs.
+Instead [use Sass comments][] to document an individual file or spec and
+[`README.md` files][] to document entire directories.
```
nshahan: Consider adding a brief explanation why to not write them.

STYLE_GUIDE.md
```diff
@@ -0,0 +1,1245 @@
...
+
+<===> double/output.css
+a b, a c {
+  d: e;
+}
+```
+
+</details>
+
+The first file of each spec within an HRX file other than the first should begin with:
```
nshahan: I'm not sure what this means. Can you reword when to use the comment divider?

STYLE_GUIDE.md
```diff
@@ -0,0 +1,1245 @@
...
+
+```hrx
+<===>
+================================================================================
+```
+
+This is an HRX comment, and has no effect on the semantics of the file. It just
+serves to visually separate specs from one another.
```
nshahan: Are these a fixed length? Is there an HRX spec where they are described you could link to?

nex3:
@xzyfer @nschonni ping!

STYLE_GUIDE.md
```diff
@@ -0,0 +1,1245 @@
...
+
+```css
+a b {
+  c: d;
+}
+```
+
+</details>
+
+All specs should be written in HRX files unless they contain invalid UTF-8
```
nschonni: Maybe mention that the filenames don't need to be UTF-8, since that was brought up in #1365

STYLE_GUIDE.md
```diff
@@ -0,0 +1,1245 @@
...
+
+<===>
+================================================================================
+<===> empty/input.scss
+// CSS requires at least one token in a custom property.
+.empty {
+  --property:;
+}
+
+<===> empty/error
```
glebm: Should this be `error/empty/error`?

STYLE_GUIDE.md
```diff
@@ -0,0 +1,1245 @@
...
+}
+
+<===> single_value/output.css
+a {
+  --b: c;
+}
+
+<===>
+================================================================================
+<===> empty/input.scss
```
glebm: Should this be `error/empty/input.scss`?

STYLE_GUIDE.md
```diff
@@ -0,0 +1,1245 @@
...
+3. Any files imported or used by `input.scss` or `input.sass`.
+4. `output.css` or `error`
+5. `warning`
+
+If a spec has [implementation-specific expectations][], those expectations
+should go directly after the corresponding default expectation.
+
+[implementation-specific expectations]: README.md#implementation-specific-expectations
+
+### DO use Dart Sass's output as the default
```
glebm: Does this also apply to errors? Currently, dart-sass errors are sometimes inferior to ruby and libsass, e.g.:

libsass and ruby:

> Invalid CSS after "... (foo: bar) or ": expected @supports condition (e.g. (display: flexbox)), was "{"

dart-sass:

> expected "("

STYLE_GUIDE.md
nex3: The filename in that issue *is* UTF-8. I believe [Git itself expects paths to be UTF-8 encoded](https://git-scm.com/docs/git-commit#_discussion), so I don't think we need to be explicit about how they're handled here.

STYLE_GUIDE.md
nex3: Yes; this is the "bad" example, intended to indicate the incorrect way to do things.

STYLE_GUIDE.md
nex3: Yes. This isn't about saying which errors are best, it's just about being consistent in formatting. If any implementation's errors need improvement, that's something that should be handled on that implementation's issue tracker, not by swapping output positions in the specs.

STYLE_GUIDE.md
```diff
@@ -0,0 +1,1245 @@
...
+
+```
+<===>
+CSS requires at least one token in a custom property.
+<===> empty/input.scss
+.empty {
+  --property:;
+}
+
+<===> empty/error
```
glebm: This should be error/empty, because this bad example is for comments (and to be consistent with the path on line 788).

STYLE_GUIDE.md
glebm: If `error` is not what we want the canonical error message to look like, but is merely the current error message of dart-sass, perhaps all errors should be language-specific?

I.e. no `error`, only `error-dart-sass` and `error-libsass`.

nex3: I think that could be a reasonable approach. Let's discuss it on a separate issue, though; until the infrastructure supports it, I'd like to keep this as the recommended approach.

STYLE_GUIDE.md
nex3: This is notionally within an `error.hrx` file (see the stack trace below), so I think it's fine as-is. Note that the spec that explicitly includes `error/` is doing so to demonstrate it as a sibling of a non-error spec.

STYLE_GUIDE.md
nshahan: Why not use HRX comments to describe a the specs? I'm not advocating for this, I was just saying I don't see any explanation why Sass comments are preferred.

nex3: Ah, got it. Added.

nex3:
@xzyfer @nschonni I'd still really value your feedback on this!

nex3:
Looks like @xzyfer and @nschonni aren't available right now, so I'm going to go ahead and merge this. Feel free to file issues for after-the-fact feedback.

PR #1365, opened 2019-03-25
Fix --test-probe

glebm:
1. Changes `folder` to `dir.to_s` (`folder` no longer exists).
2. Changes `passing.any?` to `!passing.empty?`, as `any?` has pitfalls:
   Took me a few minutes to debug when I tried with `input_path` instead of `dir.to_s`, and that was returning `nil` (`[nil].any? #=> false`).

### DIFF
--- a/lib/sass_spec/runner.rb
+++ b/lib/sass_spec/runner.rb
@@ -60,10 +60,10 @@ class SassSpec::Runner
       passing = []
       test_cases.each do |test_case|
         if test_case.todo? && test_case.result?
-          passing << test_case.folder
+          passing << test_case.dir.to_s
         end
       end
-      if passing.any?
+      if !passing.empty?
         puts "The following tests pass but were marked as TODO for #{@options[:engine_adapter].describe}:"
         puts passing.join("\n")
       else
--- a/spec/libsass/Sáss-UŢF8.hrx
+++ b/spec/libsass/Sáss-UŢF8.hrx
@@ -1,9 +1,0 @@
-<===> input.scss
-span.utf8-in-path {
-  margin: auto;
-}
-
-<===> output.css
-span.utf8-in-path {
-  margin: auto;
-}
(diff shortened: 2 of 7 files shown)


nex3:
Why does this change `libsass/Sáss-UŢF8` to a non-HRX file? It should still run the test as HRX.

glebm:
The original bug (https://github.com/sass/libsass/pull/1774) this test was added for involved the interaction of Windows command line path handling (some arcane Windows path NFC normalization stuff); it should be tested on a real filesystem.

nex3:
HRX tests are run with real files. The test runner creates a directory named after the HRX file and invokes the Sass compiler against that directory.

glebm:
Right but this test also runs using a real sassc binary (in addition to running via sass-spec).

glebm:
https://github.com/sass/libsass/blob/2050492fd001e2422b6739fa859faf5e8e10f795/appveyor.yml#L83-L92

nex3:
LibSass's specs shouldn't have such a tightly-coupled dependency on the specific structure of this repo—that's liable to break. I strongly recommend just creating a matching directory in the LibSass repo directly. That's what we do for Dart Sass tests that exercise details of the CLI.

glebm:
I can open a corresponding issue in the libsass repo to do this.
In the meantime, can you please merge this? Later on we can simply revert https://github.com/sass/sass-spec/pull/1365/commits/bae59cab9386ee79086766b953b1fb674c7a570c here

nex3:
Can you add a TODO to the spec to re-HRXify it when the LibSass issue is fixed?

glebm:
Opened https://github.com/sass/libsass/issues/2854 and added a TODO there (this way re-hrxifying will be a simple git revert, won't need to revert the TODO as well)

nex3:
I'd really rather have a TODO here too, so someone doesn't accidentally change this without realizing it'll break LibSass.

glebm:
OK, I'll amend the commit and post here again when done

glebm:
Amended

PR #1366, opened 2019-03-25
comment.hrx: libsass fixed in sass/libsass#2849

glebm:
Refs sass/libsass#2849

### DIFF
--- a/spec/css/comment.hrx
+++ b/spec/css/comment.hrx
@@ -76,11 +76,6 @@ should include their own tests for CR LF newlines.
 
 <===>
 ================================================================================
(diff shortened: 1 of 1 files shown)


glebm:
Looks like my editor or git messed with newlines.

nex3:
FYI, when you're adding tests for a particular implementation, you can add "[skip libsass]" (or whatever) to the PR message and the runner won't run the specs on LibSass. Then if you add "See https://github.com/sass/sass-spec/pull/1366" to the corresponding LibSass PR, *that* runner will run the specs in this PR rather than in `master`.

glebm:
Ah, thanks

PR #1369, opened 2019-04-04
Drop support for Ruby Sass

nex3:
Closes #1369


nex3:
[Yep!](https://github.com/sass/ruby-sass/blob/stable/sass.gemspec#L44-L53)

PR #1371, opened 2019-04-05
Run specs from within the spec directory

nex3:
This means all reported paths will be relative, which makes it much
easier to move specs without having to regenerate expected error
messages.


lib/sass_spec/engine_adapter.rb
```diff
@@ -111,7 +119,9 @@ def version
   end
 
   def compile(sass_filename, precision)
-    @stdin.puts "--no-color --no-unicode #{@args} #{sass_filename}"
+    dirname, basename = File.split(sass_filename)
+    @stdin.puts "!cd #{File.absolute_path(dirname)}"
```
glebm: I'm not sure what `!cd` is but it may be necessary to shell-escape the path (`Shellwords.escape`)

lib/sass_spec/engine_adapter.rb
```diff
@@ -54,8 +54,11 @@ def version
 
 
   def compile(sass_filename, precision)
-    cmd = "#{@command} --precision #{precision} -t expanded"
-    result = capture3_with_timeout("#{cmd} #{sass_filename}", :binmode => true, :timeout => @timeout)
+    cmd = "#{File.absolute_path(@command)} --precision #{precision} -t expanded"
```
glebm: This won't work if `File.absolute_path(@command)` contains spaces.

While you could manually escape the path like this:

```ruby
"#{Shellwords.escape File.absolute_path(@command)}"
```

Consider simply passing the arguments individually:

```ruby
capture3_with_timeout(
  File.absolute_path(@command), '--precision', precision, '-t', expanded,
  binmode: true, timeout: timeout
)
```

(see https://ruby-doc.org/stdlib-1.9.3/libdoc/open3/rdoc/Open3.html#method-c-popen3)

glebm: Also, not sure why we call `File.absolute_path` here, capture3 is already invoked in `$PWD`.

nex3: > This won't work if `File.absolute_path(@command)` contains spaces.
> 
> While you could manually escape the path like this:
> 
> ```ruby
> "#{Shellwords.escape File.absolute_path(@command)}..."
> ```

Done.

> Consider simply passing the arguments individually:
> 
> ```ruby
> capture3_with_timeout(
>   File.absolute_path(@command), '--precision', precision, '-t', expanded,
>   binmode: true, timeout: timeout
> )
> ```
> 
> (see https://ruby-doc.org/stdlib-1.9.3/libdoc/open3/rdoc/Open3.html#method-c-popen3)

This would require rewriting `capture3_with_timeout` to avoid using `spawn()`, which is probably a good idea, but not something I'd like to lump into this PR.

> I'm also not sure why we call `File.absolute_path` here, capture3 is already invoked in `$PWD`.

We change the PWD with `Dir.chdir`, so we need to make sure the command path remains valid in the new directory.

lib/sass_spec/engine_adapter.rb
nex3: It isn't; it's just used as a signal to the inner process that this is the directory to move to. See [these lines](https://github.com/sass/sass-spec/blob/453385d2a400fdaeb3b44b3821f67d20405c521d/lib/sass_spec/engine_adapter.rb#L85-L88).

lib/sass_spec/engine_adapter.rb
glebm: `spawn` also support this!

See https://ruby-doc.org/core-2.2.0/Process.html#method-c-spawn:

>   cmdname, arg1, ...          : command name and one or more arguments (This form does not use the shell. See below for caveats.)

nex3: Oh neat! I must have misread the API docs. Done.

lib/sass_spec/engine_adapter.rb
```diff
@@ -1,4 +1,5 @@
 require 'open3'
+require 'shellwords'
```
glebm: nit: unused require

lib/sass_spec/engine_adapter.rb
```diff
@@ -54,8 +55,12 @@ def version
 
 
   def compile(sass_filename, precision)
-    cmd = "#{@command} --precision #{precision} -t expanded"
-    result = capture3_with_timeout("#{cmd} #{sass_filename}", :binmode => true, :timeout => @timeout)
+    command = File.absolute_path(@command)
+    dirname, basename = File.split(sass_filename)
+    result = Dir.chdir(dirname) do
+      capture3_with_timeout(command, "--precision", precision.to_s, "-t", "expanded", basename,
+        binmode: true, timeout: @timeout)
```
glebm: nit: Instead of `Dir.chdir(dirname) { ... }` you can pass `chdir: dirname`

PR #1374, opened 2019-04-10
travis: Build libsass in parallel

glebm:


glebm:
Looks like `nproc` reports 1 on travis

PR #1375, opened 2019-04-11
Drop support for multiple language versions

### LINKED ISSUE #1223: Remove start_version and end_version
Once #1222 is done, I don't think it'll make much sense at all to talk about version numbers for Sass as a language, or to track them as part of our specs. Instead, I think we should focus on which features are supported by which implementations.

Dart Sass may eventually want to switch to multi-tier development, like Ruby Sass's `stable`/`next`/`master` distinction, but I still don't think tracking language versions is the right way to do that. Instead, I think we should extend `todo` and `ignore_for` to support implementation-level version numbers. For example:

```yaml
---
:ignore_for:
- dart-sass < 1.1
```

would only run the test on Dart Sass 1.0.x.

nex3:
Closes #1223


glebm:
LGTM for the Ruby changes, still need @xzyfer to spot check the spec changes.

lib/sass_spec/engine_adapter.rb
```diff
@@ -66,6 +51,10 @@ def compile(sass_filename, precision)
       [result[:stdout], result[:stderr], result[:status].exitstatus]
     end
   end
+
+  def to_s
+    "#@description #@version"
```
glebm: Consider avoiding shorthand interpolation syntax throughout.

Search and replace from `@description` to `description` (e.g. changing an instance variable to a method) breaks these in a way that doesn't raise an exception.

This even caused outages, e.g. http://status.cloudamqp.com/incidents/vj62pnp62tj9
There is a detailed discussion here: https://bugs.ruby-lang.org/issues/10541

nex3: Done.

nex3:
I don't know if @xzyfer is likely to have time to review this... he hasn't been active on the issue tracker lately (see #1361 and #1363). Can I land this for now and have him look it over later when he's got time?

glebm:
Sure

mgreter:
FYI: this currently breaks libsass CI:
```
sass_spec/cli.rb:105:in `parse': invalid option: -V (OptionParser::InvalidOption)
```

Beside that there are other breaking changes from earlier commits:
https://travis-ci.org/sass/libsass/builds/515994566

mgreter:
Fixed by https://github.com/sass/libsass/pull/2865

nex3:
@mgreter Thanks for the quick fix!

PR #1376, opened 2019-04-11
 Add todo spec test for libsass issue 2863 

mgreter:
https://github.com/sass/libsass/issues/2863

### DIFF
--- a/spec/libsass-todo-issues/issue_2863.hrx
+++ b/spec/libsass-todo-issues/issue_2863.hrx
@@ -1,0 +1,21 @@
+<===> input.scss 
+$squoted: "'dquoted'";
+$dquoted: "\"squoted\"";
+
+test {
+    str-slice-single: str-slice( $squoted, 1, 2 );
+    str-slice-double: str-slice( $dquoted, 1, 2 );
+    str-index-single: str-index( $squoted, "q" );
+    str-index-double: str-index( $dquoted, "q" );
+    str-insert-single: str-insert( $squoted, "p", 2 );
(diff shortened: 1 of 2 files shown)


nex3:
Even though [the new style guide](https://github.com/sass/sass-spec/blob/88f4ef08a606792656ddb79c4855472e75822881/STYLE_GUIDE.md) hasn't officially landed yet, please follow it when adding any new specs.

mgreter:
@nex3 can you elaborate where I got it wrong?

glebm:
A couple things stand out:

1. Use short names like `a` for selectors (instead of `test` etc).
2. Test naming: Instead of tying tests to issue numbers, organize them based on features.

   E.g. this test should go to something like `sass-spec/string-functions/str-{slice,index,insert}.hrx` and within the HRX the tests should be named something like `quotes/input.scss`.

nex3:
Exactly (see [DO use single-letter names for irrelevant placeholders](https://github.com/sass/sass-spec/blob/master/STYLE_GUIDE.md#do-use-single-letter-names-for-irrelevant-placeholders), [DO organize specs by which part of the language they test](https://github.com/sass/sass-spec/blob/master/STYLE_GUIDE.md#do-organize-specs-by-which-part-of-the-language-they-test), and [DO use descriptive paths](https://github.com/sass/sass-spec/blob/master/STYLE_GUIDE.md#do-use-descriptive-paths)). [DO test only one thing per spec](https://github.com/sass/sass-spec/blob/master/STYLE_GUIDE.md#do-test-only-one-thing-per-spec) is also relevant.

PR #1377, opened 2019-04-12
Activate specs for issue 2863

mgreter:
This PR activates specs for sass/libsass#2863


mgreter:
Now checking if probe-todo report passing test

glebm:
A couple things stand out:

1. Use short names like `a` for selectors (instead of `test` etc).
2. Test naming: Instead of tying tests to issue numbers, organize them based on features.

E.g. this test should go to something like `sass-spec/string-functions/str-{slice,index,insert}.hrx` and within the HRX the tests should be named something like `quotes/input.scss`.

PR #1378, opened 2019-04-12
Ignore - Ci challange

mgreter:
Just checking that CI is doing it's job ... should fail


mgreter:
@nex3 could you please check why my change that should have triggered a failed build did not fail!
Something seems off! First thought I made an error, but even altering existing tests fail to fail!???

mgreter:
OK, it did fail for the libsass run ...

nex3:
Can we close this now?

PR #1380, opened 2019-04-13
Add tests for `quote`

glebm:
Refs https://github.com/sass/libsass/issues/2811

[skip libsass]

libsass PR: https://github.com/sass/libsass/pull/2873

### DIFF
--- a/spec/string-functions/quote.hrx
+++ b/spec/string-functions/quote.hrx
@@ -1,0 +1,80 @@
+<===> unquoted/input.scss
+a {
+  a: quote(a);
+}
+
+<===> unquoted/output.css
+a {
+  a: "a";
+}
+
+<===>
+================================================================================
+<===> quoted-double/input.scss
+a {
+  a: quote("a");
+}
+
+<===> quoted-double/output.css
+a {
+  a: "a";
+}
+
+<===>
+================================================================================
+<===> quoted-single/input.scss
+a {
+  a: quote('a');
+}
+
+<===> quoted-single/output.css
+a {
+  a: "a";
+}
+
+<===>
+================================================================================
+<===> quote-mark-selection-single/input.scss
+a {
+  a: quote(unquote('"'));
+}
+
+<===> quote-mark-selection-single/output.css
+a {
+  a: '"';
+}
+
+<===>
+================================================================================
+<===> quote-mark-selection-double/input.scss
+a {
+  a: quote(unquote('"') + unquote("'"));
+}
+
+<===> quote-mark-selection-double/output.css
+a {
+  a: "\"'";
+}
+
+<===>
+================================================================================
+<===> error/argument-type/input.scss
+a {
+  a: quote((1, 2, 3));
+}
+
+<===> error/argument-type/error
+Error: $string: 1, 2, 3 is not a string.
+  ,
+2 |   a: quote((1, 2, 3));
+  |      ^^^^^^^^^^^^^^^^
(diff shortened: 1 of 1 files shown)


glebm:
In the future, the libsass PR should be merged first so we don't break the libsass build

spec/string-functions/quote.hrx
```diff
@@ -0,0 +1,80 @@
+<===> unquoted/input.scss
+a {
+  a: quote(a);
```
nex3: Style nits:

* Placeholder names [should be alphabetical](https://github.com/sass/sass-spec/blob/master/STYLE_GUIDE.md#do-use-single-letter-names-for-irrelevant-placeholders), rather than all being `a`.

* Rules these small should be [one-liners](https://github.com/sass/sass-spec/blob/master/STYLE_GUIDE.md#prefer-one-line-rules-when-they-contain-a-single-child-with-no-children).

For example:

```scss
a {b: quote(c)}
```

spec/string-functions/quote.hrx
```diff
@@ -0,0 +1,80 @@
...
+}
+
+<===> unquoted/output.css
+a {
+  a: "a";
+}
+
+<===>
+================================================================================
+<===> quoted-double/input.scss
```
nex3: Style nit: paths should be [underscore-separated](https://github.com/sass/sass-spec/blob/master/STYLE_GUIDE.md#prefer-underscore-separated-paths).

spec/string-functions/quote.hrx
```diff
@@ -0,0 +1,80 @@
...
+}
+
+<===> quoted-single/output.css
+a {
+  a: "a";
+}
+
+<===>
+================================================================================
+<===> quote-mark-selection-single/input.scss
```
nex3: What does "selection" indicate in this path name?

glebm: That the `quote` function selects the most suitable outer quotes (`'` or `"`).
English is not my first language, sorry if this doesn't make sense!

nex3: The type of quote that's printed for a string is chosen when that string is serialized to CSS; it's not an aspect of the string object itself. The `quote()` function shouldn't know anything about it.

glebm: > The type of quote that's printed for a string is chosen when that string is serialized to CSS

Where is this specified?

`libsass` does use different types of quoting depending on the code path.

Can we keep this test here? It's conceivable that other future implementations might make the same mistake.

nex3: We don't have a full prose specification (although we're working on building one piecemeal in [sass/sass](/sass/sass)), but there should be no observable case where the way a string is created (with single or double quotes) affects the way it's serialized in the reference implementation.

We can keep the tests (maybe name them `quote_unquoted_quote/single` and `quote_unquoted_quote/double`?), but they should have comments indicating that they're regression tests and linking to the issues in question.

glebm: Done in https://github.com/sass/sass-spec/commit/dc5f5d551e4172eeccbb8b40922edfaf1a9777b5

spec/string-functions/quote.hrx
glebm: Done

spec/string-functions/quote.hrx
glebm: Done

PR #1382, opened 2019-04-13
Add spec for `inspect` with nested list

glebm:
Refs https://github.com/sass/libsass/issues/2654

### DIFF
--- a/spec/core_functions/inspect.hrx
+++ b/spec/core_functions/inspect.hrx
@@ -371,6 +371,26 @@ a {
 
 <===>
 ================================================================================
+<===> list/nested/options.yml
+---
+:todo:
+- libsass # sass/libsass#2654
+
+<===> list/nested/input.scss
+a {
+  $a: inspect([() (())]);
+  a: $a;
+  b: type-of($a);
+}
+
+<===> list/nested/output.css
+a {
+  a: [() ()];
+  b: string;
+}
+
+<===>
+================================================================================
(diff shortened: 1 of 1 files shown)


spec/core_functions/inspect.hrx
```diff
@@ -369,6 +369,26 @@ a {
...
+<===>
+================================================================================
+<===> list/nested/options.yml
+---
+:todo:
+- libsass # sass/libsass#2654
+
+<===> list/nested/input.scss
+a {
+  $a: inspect([() (())]);
```
nex3: Why is this `(())` here? That's no different than `()`, and if we're testing that those two are parsed identically, that should probably go in the specs for lists rather than the specs for `inspect()`.

glebm: I'm not sure. This is for https://github.com/sass/libsass/issues/2654.

The important bit is:

```
@debug $example; //=> [() ()]
@debug inspect($example); //=> [(() ())]
```

So it does seem to be specific to inspect.

nex3: I would just get rid of the extra parentheses around the second list, then:

```suggestion
  $a: inspect([() ()]);
```

glebm: Thanks, done in https://github.com/sass/sass-spec/commit/18b91d4abdc9ad6f6a8b90f1945ed3e5b95c583c

PR #1383, opened 2019-04-13
Add test for nested `&.:not`

glebm:
Refs https://github.com/sass/libsass/issues/2630

### DIFF
--- a/spec/nesting/not.hrx
+++ b/spec/nesting/not.hrx
@@ -1,0 +1,10 @@
+<===> multiple_parent_selectors_with_trailing_ident/input.scss
+// Regression test for sass/libsass#2630
+.a, .b {
+  :not(&-c) {d: e}
+}
+
+<===> multiple_parent_selectors_with_trailing_ident/output.css
+:not(.a-c, .b-c) {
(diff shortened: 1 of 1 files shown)


nex3:
I think this test is a lot more complicated than it needs to be. In Node Sass 4.11.0, the error reproduces with as little as:

```scss
.a, .b {
  :not(&-c) {
    d: e;
  }
}
```

glebm:
Done

spec/nesting/not.hrx
```diff
@@ -0,0 +1,9 @@
+<===> with-parent/input.scss
```
nex3:
```suggestion
<===> multiple_parent_selectors_with_trailing_ident/input.scss
```

It's good to be explicit about what's being tested, especially when it's a case like this that was triggered by a confluence of different factors.

spec/nesting/not.hrx
```diff
@@ -0,0 +1,9 @@
...
+.a, .b {
```
nex3: Please add a comment indicating that this is a regression test.

PR #1386, opened 2019-04-29
Add specs for attribute selectors with modifiers

nex3:
See sass/dart-sass#656

[skip dart-sass]


spec/css/selector.hrx
```diff
@@ -0,0 +1,199 @@
...
+
+   ---^
+
+<===>
+================================================================================
+<===> error/attribute/modifier/unicode/input.scss
+// Attribute modifiers must be ASCII alphabetical characters.
+[a=b ï] {c: d}
+
+<===> error/attribute/modifier/unicode/error
```
jathak: It looks like this is failing becomes of a missing error message here.

nex3: Fixed, thanks!

lib/sass_spec/util.rb
```diff
@@ -3,7 +3,7 @@ class << self
     # Normalizes the whitespace in the given CSS to make it easier to compare
     # across implementations.
     def normalize_output(css)
-      css.gsub(/(?:\r?\n)+/, "\n")
+      css.gsub(/(?:\r?\n)+/, "\n").force_encoding("utf-8")
```
glebm: `force_encoding` doesn't re-encode, it simply forces Ruby to interpret them as UTF-8.
If they're already UTF-8, it shouldn't be necessary. If they're not UTF-8, this won't help (you'd want `encode('utf-8')` instead).

What was the issue here?

nex3: Both [Dart Sass](https://github.com/sass/sass-spec/blob/6a11df8b633114b595eab96d831b31646aaa5661/lib/sass_spec/engine_adapter.rb#L97-L98) and [LibSass](https://github.com/sass/sass-spec/blob/6a11df8b633114b595eab96d831b31646aaa5661/lib/sass_spec/engine_adapter.rb#L46) explicitly set their output streams to binary encoding. I believe this is necessary for some of the LibSass regression tests involving invalid UTF-8-encoded input.

PR #1389, opened 2019-05-06
Merge master into feature.use

nex3:
[skip dart-sass]


nex3:
Because this is merging in https://github.com/sass/sass-spec/pull/1372, which deleted the very large `spec/output_styles` directory.

glebm:
Perhaps this should be a rebase instead of a merge?

nex3:
Rebases of many commits tend to make it harder to resolve merge conflicts, and provide a less accurate view of the evolution of a branch over time. Also you can't do pull requests for them, which means you don't get CI verification of their validity before merging.

PR #1390, opened 2019-05-13
Improve url() ! tests

glebm:
1. A lone `!` is invalid -- I don't think the compiler should be expected to allow it.
2. Added a few more valid examples.

Checked using the following JavaScript:

    try {
      new URL(url, "https://google.com");
      console.log(`${url} - OK`);
    } catch {
      console.warn(`${url} - invalid`);
    }

Refs sass/libsass#2880
Refs sass/dart-sass#646
Refs #1384

### DIFF
--- a/spec/css/url.hrx
+++ b/spec/css/url.hrx
@@ -5,20 +5,60 @@
 
 <===>
 ================================================================================
-<===> exclam/only/input.scss
-a {b: url(!)}
+<===> exclam/absolute/path/input.scss
+a {b: url(http://a.b/c!d)}
 
-<===> exclam/only/output.css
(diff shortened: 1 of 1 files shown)


nex3:
According to [CSS Syntax Level 3](https://www.w3.org/TR/css-syntax-3/#url-token-diagram), a `<url-token>` without quotes can contain any unescaped character other than `"`, `'`', `(`, `)`, `\`, whitespace, or non-printable characters. Whether it constitutes a valid WhatWG URL or not is not syntactically relevant.

stof:
@nex3 should this be left open if the change is invalid ?

PR #1402, opened 2019-06-12
Add specs for new module-system features for existing functions

nex3:
[skip dart-sass]


nex3:
I always try to split "moving existing tests around" and "adding new tests" into separate commits. Can you try reviewing each commit separately and see if that's easier?

nshahan:
It's hard to tell in the last commit 1b190ef25cc0cc6682af710a79fd453e2fb1b407. It looks like `spec/core_functions/meta/get_function.hrx` was deleted and it's contents split into multiple files with some updated specs at the same time.

nex3:
Oh, that's true, I did split that one up separately because my new specs caused it to hit the 500-line limit. I'll try to do those in separate commits too in the future.

PR #1405, opened 2019-06-17
Make specs for remaining color functions style-guide-compliant

nex3:


spec/core_functions/color/scale_color/error/args.hrx
```diff
@@ -0,0 +1,60 @@
...
+<===> unknown/error
+Error: No argument named $hue.
+  ,
+1 | a {b: scale-color(red, $hue: 10%)}
+  |       ^^^^^^^^^^^^^^^^^^^^^^^^^^^
+  '
+  input.scss 1:7  root stylesheet
+
+<===> unknown/error-libsass
+Error: Function scale-color has no parameter named $ambience
```
jathak: This (and below) should be `$hue` to match the input and Dart Sass error

spec/core_functions/color/scale_color/error/args.hrx
```diff
@@ -0,0 +1,60 @@
...
+  |       ^^^^^^^^^^^^^^^^^^^
+  '
+  input.scss 1:7  root stylesheet
+
+<===>
+================================================================================
+<===> unknown/options.yml
+---
+:todo:
+- sass/libsass#2901
```
jathak: This should be sass/libsass#2902

spec/core_functions/color/scale_color/error/args.hrx
nex3: We just shouldn't include LibSass errors for specs that are marked as TODO for LibSass.

PR #1407, opened 2019-06-17
Add deprecation warning for libsass global variable creation

mgreter:
It seems there were some strange newlines in some of these hrx files.

### DIFF
--- a/spec/libsass-closed-issues/issue_1081.hrx
+++ b/spec/libsass-closed-issues/issue_1081.hrx
@@ -1,8 +1,3 @@
-<===> options.yml
----
-:warning_todo:
-- sass/libsass#2834
-
 <===> input.scss
 $foo: foo !global !default;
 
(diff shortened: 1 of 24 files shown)


spec/libsass-closed-issues/issue_1081.hrx
```diff
@@ -52,3 +47,8 @@ declare new variables. Consider adding `$foo: null` at the top level.
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^
   '
     input.scss 1:1  root stylesheet
+
+<===> warning-libsass
+DEPRECATION WARNING on line 1, column 5 of /sass/spec/libsass-issues/issue_1081/input.scss:
+!global assignments won't be able to declare new variables in future versions.
+Consider adding `$foo: null` at the top level.
```
nex3: This warning can be confusing, because the issue with the particular case where the `!global` declaration is at the top-level is more that `!global` is unnecessary. We're probably going to adjust the warning in this case in Dart Sass, and you may want to do so as well (see https://github.com/sass/dart-sass/issues/723).

PR #1408, opened 2019-06-18
Materialize parent HRX directories if necessary

nex3:
If an HRX directory is being materialized with #with_real_files and it
contains a file that reaches outside itself (with `../`), materialize
its parent directory as well. This allows individual tests to share
utility libraries that are defined in the same HRX archive.

### DIFF
--- a/lib/sass_spec/directory.rb
+++ b/lib/sass_spec/directory.rb
@@ -164,9 +164,14 @@ class SassSpec::Directory
   def with_real_files
     return yield unless @archive
 
+    files = @archive.entries.select {|entry| entry.is_a?(HRX::File)}.to_a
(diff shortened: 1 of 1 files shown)


lib/sass_spec/directory.rb
```diff
@@ -195,6 +200,13 @@ def to_s
 
   private
 
+  # Returns whether `file` contains enough `../` references to reach outside
+  # this directory.
+  def _reaches_out?(file)
+    depth = file.path.count("/")
```
rambleraptor: Do we know if file paths have trailing slashes?

nex3: The HRX spec guarantees that they won't; otherwise they'd be parsed as `HRX::Directory`s instead.

PR #1420, opened 2019-06-28
fix: Enable libsass 2914

nschonni:
Depends on https://github.com/sass/libsass/pull/2933

PR #1421, opened 2019-06-28
fix: Enable for libsass 2894

nschonni:
Depends on https://github.com/sass/libsass/pull/2934
[skip libsass]


nex3:
Same request as for #1420: please add "[skip libsass]" to the PR description

nex3:
I'm landing this even though Travis is red because multiple changes (including this one) were landed in LibSass at once, so we need to land multiple fixes at once to get everything green again.

PR #1422, opened 2019-07-02
Make meta function specs style-guide-compliant

nex3:


spec/core_functions/meta/mixin_exists.hrx
```diff
@@ -0,0 +1,384 @@
...
+<===> different_module/undefined/output.css
+a {
+  b: false;
+}
+
+<===>
+================================================================================
+<===> different_module/defined/input.scss
+@use "other";
+a {b: mixin-exists("c", "color")}
```
jathak: Should this be `mixin-exists("c", "other")`?

spec/core_functions/meta/mixin_exists.hrx
```diff
@@ -0,0 +1,384 @@
...
+@mixin d() {}
+
+<===> different_module/chosen_prefix/output.css
+b {
+  c: true;
+}
+
+<===>
+================================================================================
+<===> different_module/through_use/options.yml
```
jathak: Unnecessary given the options.yml in the outer directory? (same below)

spec/core_functions/meta/mixin_exists.hrx
nex3: It sure should!

spec/core_functions/meta/mixin_exists.hrx
nex3: Done.

PR #1425, opened 2019-07-08
[WIP][DNM] First batch of specs for libsass parser backport

mgreter:
See https://github.com/sass/libsass/pull/2918


mgreter:
The current diff is about around where I want to aim for libsass 4.0. Will take me probably one week alone to sort out, rebase and squash all changes into groups that make sense to review ...

mgreter:
@nex3 I seem to stumble a lot over the error "HRX::ParseError: Parse error on line 1, column 17 of spec/libsass/variable-scoping/blead-global/expanding/at-root.hrx: Invalid character U+000D". This seems to do with CRLF vs LF feeds. Not sure why the HRX parser fails about them with a hard error? My naive [perl parser][1] processes them without a problem ...

[1]: https://github.com/sass/perl-libsass/blob/develop/t/99_sass_specs.t#L12-L34

PR #1426, opened 2019-07-08
Allow arguments to be passed to any command

mgreter:
Renames `--dart-args` to `--cmd-args`

Allows libsass to work around #1427.

@nex3 I probably will merge this very soon since #1427 is quite a blocker ATM for me.
I think this change makes sense even on its own, but feel free to change it as you see fit.


mgreter:
CI is at least green again ...

nex3:
*Please* wait to get a review before landing PRs. If you want to build further PRs on top of them, that's fine; you can just rebase the upstream commit away once it lands.

PR #1429, opened 2019-07-11
Add specs for at-rules in nested imports

nex3:
See sass/dart-sass#755

[skip dart-sass]


spec/directives/import/nested.hrx
```diff
@@ -0,0 +1,125 @@
...
+<===>
+================================================================================
+<===> at_rule/declaration_child/input.scss
+a {@import "other"}
+
+<===> at_rule/declaration_child/_other.scss
+@b {c: d}
+
+<===> at_rule/declaration_child/output.css
+@b {
```
nshahan: I'm not familiar with the expectation here but just want to confirm that the `@b` should be moved to top level here but not when it is childless?

nex3: That's right. An at-rule with children is assumed to behave similar to `@media`/`@supports` *et al* where browsers expect it to appear at the root and contain style rules.

Issue #944, opened 2016-10-15
Run specs against expanded mode by default


nex3:
Currently the bulk of the specs are written in nested mode, which is generally pretty unpopular. New implementations may not even want to implement it. Expanded mode is more common and more in line with standard CSS style; it should be the default, and only specs that specifically intend to exercise style support should be defined against other styles.

nschonni:
I agree that it is probably easier to read, but are you going to change the language default too http://sass-lang.com/documentation/file.SASS_REFERENCE.html#output_style
If so, it should be just a matter of regenerating the tests. Otherwise, it's still doable, just need to change the runner defaults and then regenerate everything.

nex3:
I'm not sure if we'll change the executable default—it seems like more hassle than it's worth. But we could consider it.

xzyfer:
I support this ❤️  I see a world where LibSass only supports expanded and compressed outputs.

nschonni:
> I support this  I see a world where LibSass only supports expanded and compressed outputs.

Now would probably be the right time, because you could depricate it in 3.5, and then remove it in 4.0.

xzyfer:
I think it's too early for that. It's still just a thought.

PR #1430, opened 2019-07-15
Add spec tests for compressed mode

mgreter:


nex3:
We intentionally removed output-style-specific tests from this repo in https://github.com/sass/sass-spec/pull/1372, because they're not part of the language spec, they made it difficult to make large changes across the repo, and they substantially increased CI iteration time. I don't think we want to add them back.

mgreter:
I get the point to not have a multitude of duplicated specs, but IMO it makes sense to test the bare minimum. It is the main test for libsass and I would like to make sure we don't regress on different output-styles. As a side-note it also helps libsass to get full code coverage, as certain code paths are only triggered in compressed/compact output style. I'm also ok to only support compressed output style.

nex3:
I think those tests should live in LibSass's own repo, [like we do for Dart Sass](https://github.com/sass/dart-sass/blob/master/test/compressed_test.dart). Because they're not part of the language, there's no expectation that any implementation's compression will be the same as any other's, which by and large negates the value of having them in a shared location.

mgreter:
The issue is that this is not as simple to accomplish in c++ as it might be in dart, and it would be easy by using the sass-spec runner. I also wouldn't care if we can get this only for libsass ...

nex3:
The sass-spec runner can run on any folder, including ones not in this repo—could you create a `compressed_spec` directory in the `libsass` repo and run the spec runner on that?

mgreter:
Of course we could, but I would need the changes in this PR none the less.
But I would also like to avoid the need to setup a special spec runner just for
a special directory. It would just be much easier to employ the current spec runner
for this. And IMO the compressed mode is even more canonical than any other
output mode (e.g. why did we "agree" to have all specs in expanded mode?).

As I wrote earlier, can't we just have the current output mode and additionally a
"emit the minimum chars to be css compatible" mode? Thanks for consideration.

nex3:
> Of course we could, but I would need the changes in this PR none the less.

Why? Can't you just pass `--cmd-args --style=compressed`?

> But I would also like to avoid the need to setup a special spec runner just for a special directory in CI. It would just be much easier to employ the current spec runner for this.

I'm sorry, but I don't think it makes sense to add what are essentially implementation-specific tests to a multi-implementation repo. Sometimes some extra work is necessary to get the organization as neat as possible.

> And IMO the compressed mode is even more canonical than any other output mode (e.g. why did we "agree" to have all specs in expanded mode?).

We discussed this in https://github.com/sass/sass-spec/issues/944: expanded mode matches the most commonly-used CSS style, and in practice it's the only human-readable style that Dart Sass implements.

> As I wrote earlier, can't we just have the current output mode and additionally a "emit the minimum chars to be css compatible" mode? Thanks for consideration.

I don't want compression to be part of the Sass specification—I don't want to dictate to each implementation what sorts of optimizations they should offer, or indeed whether they should offer any at all or delegate that to a downstream compression tool.

xzyfer:
@mgreter FWIW I've long been in favour of removing compressed mode from LibSass. IMHO it's beyond our scope to offer industry competitive compression. The tooling in this space is has come a very long way since compressed mode was added to Sass and has never been easier to use. 

I would be favour of dropping compressed output entirely as alternative to this problem.

PR #1431, opened 2019-07-15
Fix a filename typo

kaj:
Fix what I assue must be a typo in a filename in `different_module/defined` in `spec/core_functions/meta/mixin_exists.hrx`


nex3:
Thanks!

PR #1434, opened 2019-07-15
Update feature-exists() specs for sass/libsass#2929

nex3:
This should have been landed along with sass/libsass#2929, but that
ship has sailed.

### DIFF
--- a/spec/core_functions/meta/feature_exists.hrx
+++ b/spec/core_functions/meta/feature_exists.hrx
@@ -106,7 +106,7 @@ Error: $feature: 1 is not a string.
   input.scss 1:7  root stylesheet
 
 <===> error/type/error-libsass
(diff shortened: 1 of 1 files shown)


spec/core_functions/meta/feature_exists.hrx
```diff
@@ -111,7 +106,7 @@ Error: $feature: 1 is not a string.
   input.scss 1:7  root stylesheet
 
 <===> error/type/error-libsass
-Error: argument `$name` of `feature-exists($name)` must be a string
+Error: argument `$feature` of `feature-exists($name)` must be a string
```
nschonni: Oops, shouldn't the signiture also be updated, or did I miss that?
```suggestion
Error: argument `$feature` of `feature-exists($feature)` must be a string
```

PR #1439, opened 2019-07-17
fix: Add error for non-int str-insert index

nschonni:
[skip libsass]


nschonni:
My C++ is pretty limited, so I'm just knocking off the stuff I understand. I'm not sure if some of that larger stuff is being handled in @mgreter Dart Sass compatibility PR

PR #1442, opened 2019-07-17
fix: Update call parameter to $function

nschonni:
[skip libsass]


nex3:
Can you also get rid of the TODO from the `named` test?

nschonni:
Done

PR #1443, opened 2019-07-18
fix: non-integers passed to str-slice

nschonni:
[skip libsass]


nschonni:
@nex3 any idea why this one isn't skipping libsass properly?

nex3:
Travis CI doesn't provide direct access to the message of the pull request that triggered the build, so we have to get it through the GitHub API. Because you're not a repo admin, though, you don't have access to @SassBot's GitHub credentials through Travis, so you're operating on the bot's (very limited) GitHub request quota. If that fails, it doesn't know to skip libsass.

PR #1445, opened 2019-07-21
chore: Flatten Travis config

nschonni:
- Remove deprecated "sudo" option
- Use named matrix sets instead of conditions

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,47 +1,36 @@
-sudo: false
-
 language: ruby
 
 # Only building master means that we don't run two builds for each pull request.
 branches:
   only: [master]
 
-env:
-  global:
-    - SASS_LIBSASS_PATH=$TRAVIS_BUILD_DIR/../libsass
-    - SASS_SASSC_PATH=$TRAVIS_BUILD_DIR/../sassc
-    - SASS_SPEC_PATH=~$TRAVIS_BUILD_DIR/sass-spec
-
 matrix:
   fast_finish: true
   include:
-    - env: IMPL=libsass COMMAND="../sassc/bin/sassc"
-    - env: IMPL=dart-sass
+    - name: "LibSass"
+      env:
+        - IMPL=libsass
+        - COMMAND="../sassc/bin/sassc"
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -1,47 +1,36 @@
...
+        - SASS_SASSC_PATH=$TRAVIS_BUILD_DIR/../sassc
+        - SASS_SPEC_PATH=~$TRAVIS_BUILD_DIR/sass-spec
+      before_script:
+        - git clone https://github.com/sass/libsass.git $SASS_LIBSASS_PATH
+        - (cd $SASS_LIBSASS_PATH; git checkout $GITISH)
+        - git clone https://github.com/sass/sassc.git $SASS_SASSC_PATH
+        - (cd $SASS_SASSC_PATH; git checkout $GITISH)
+        - make -C $SASS_SASSC_PATH
+      script: bundle exec sass-spec.rb --impl $IMPL -c $COMMAND;
+    - name: "Dart-Sass"
```
nex3:
```suggestion
    - name: "Dart Sass"
```

.travis.yml
```diff
@@ -1,47 +1,36 @@
...
+      env: IMPL=dart-sass
+      before_script:
+        - curl -o dart.zip "https://storage.googleapis.com/dart-archive/channels/dev/release/latest/sdk/dartsdk-linux-x64-release.zip"
```
nex3:
```suggestion
        - curl -o dart.zip "https://storage.googleapis.com/dart-archive/channels/stable/release/latest/sdk/dartsdk-linux-x64-release.zip"
```

I'm not sure why I had this using Dart's dev channel before.

nschonni: I tried using the `dart:` key for the built-in travis support, but it didn't seem to work when I tried it

nex3: Yeah, that assumes you have a `pubspec.yaml` at the root of your project, and if you don't it'll break.

nschonni: Gotcha, didn't really dig too far 😄

PR #1453, opened 2019-08-13
Syle-guide-ify is-superselector() specs

nex3:
[skip dart-sass]


nshahan:
These changes would be a lot easier to review if they were made is separate commits. Change the contents of the files, then in a separate commits change the file paths, names, convert to hrx etc. I can't see the changes to the content because they are all being shown as file deletions and new files.

nex3:
In this case, they actually *are* file deletions and new files. I rewrote the specs from scratch and then afterwards deleted whatever existing tests I could find that covered the same functionality.

nshahan:
Ah ok thanks. I'll start looking at them.

spec/core_functions/selector/is_superselector/simple/pseudo/no_arg.hrx
```diff
@@ -0,0 +1,57 @@
...
+<===> class/and_element/output.css
+a {
+  b: false;
+}
+
+<===>
+================================================================================
+<===> element/equal/input.scss
+a {b: is-superselector(":c", ":c")}
+
```
nshahan: Duplicated above in class/equal/input.scss

spec/core_functions/selector/is_superselector/simple/pseudo/no_arg.hrx
```diff
@@ -0,0 +1,57 @@
...
+<===> element/equal/output.css
+a {
+  b: true;
+}
+
+<===>
+================================================================================
+<===> element/unequal/input.scss
+a {b: is-superselector(":c", ":d")}
```
nshahan: Also duplicated above in class/unequal/input.scss

spec/core_functions/selector/is_superselector/simple/pseudo/selector_arg/current.hrx
```diff
@@ -0,0 +1,80 @@
...
+<===> prefix/superset/input.scss
+a {
+  b: is-superselector(
+      ":-pfx-current(c d, e f, g h)",
+      ":-pfx-current(c d.i, e j f)");
+}
+
+<===> prefix/superset/output.css
+a {
+  b: false;
```
nshahan: Should be true here?

spec/core_functions/selector/is_superselector/simple/pseudo/selector_arg/has.hrx
```diff
@@ -0,0 +1,47 @@
...
+
+<===> subset/output.css
+a {
+  b: false;
+}
+
+<===>
+================================================================================
+<===> bare_sub/input.scss
+a {b: is-superselector(":has(c d.i, e j f)", "c d, e f, g h")}
```
nshahan: Should this be testing the same set of selectors except bare in the sub?

spec/core_functions/selector/is_superselector/simple/pseudo/selector_arg/host.hrx
```diff
@@ -0,0 +1,47 @@
...
+
+<===> subset/output.css
+a {
+  b: false;
+}
+
+<===>
+================================================================================
+<===> bare_sub/input.scss
+a {b: is-superselector(":host(c d.i, e j f)", "c d, e f, g h")}
```
nshahan: Same here, should the selectors in $sub be the same except bare?

spec/core_functions/selector/is_superselector/simple/pseudo/no_arg.hrx
nex3: Done.

spec/core_functions/selector/is_superselector/simple/pseudo/no_arg.hrx
nex3: Done.

spec/core_functions/selector/is_superselector/simple/pseudo/selector_arg/current.hrx
nex3: Unlike almost all selector pseudos, `:current()`'s contents have to match *exactly* rather than being a superselector match. [The `:current()` pseudo](https://www.w3.org/TR/selectors-4/#the-current-pseudo) matches the currently-displayed element *if it matches the argument*, and otherwise matches the innermost parent of the argument. This means that if you broaden the set of elements matched by the argument, it can actually *stop matching* an element that it used to. So it's only safe to consider it a superselector of exactly identical selectors.

spec/core_functions/selector/is_superselector/simple/pseudo/selector_arg/has.hrx
nex3: Yeah, good point, that's a better way to structure this.

PR #1454, opened 2019-08-15
Use specs folder instead of spec folder

rambleraptor:
RSpec expects to [find tests in the 'spec/' folder](https://github.com/rspec/rspec-core/issues/1930#issuecomment-90803094).

RSpec will probably be the unit test runner that sass-spec uses and contributors expect to see tests in the `spec/` folder.

I've moved the language specs into `specs/` and changed the runner to use that folder by default.

This changes so many files (too many files!) but just a folder rename.


rambleraptor:
The one deleted file is fun. Macs and Windows aren't case-sensitive, so cloning this repo squashes that one file by default.

It's normally not an issue but it means that this test case isn't fully tested for any users on Mac/Windows. (It also means I can't commit/alter that file on a Mac, which is a pain, but I can find a way).

What would you like to do?

error message:
```
spwarning: the following paths have collided (e.g. case-sensitive paths
on a case-insensitive filesystem) and only one from the same
colliding group is in the working tree:

  'spec/libsass/Sáss-UŢF8.hrx'
  'spec/libsass/Sáss-UŢF8.hrx'
```

nex3:
Would it be possible to configure RSpec to use `spec/`? I'm not opposed to doing a giant move, but it is likely to cause some churn for other repos that invoke the spec runner, outstanding pull requests and feature branches, and so on. If it's easy to just not bother, that may be the better part of valor.

rambleraptor:
Yeah, let's just do that then. I wasn't aware of downstream dependencies on this.

PR #1456, opened 2019-08-16
Test case metadata units

rambleraptor:
This is the first unit test!

I'll probably be biasing towards integration tests with Aruba, but this was a good way to get rspec setup.

I added RSpec + instructions to Gemfile/Readme accordingly.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -5,3 +5,7 @@ gem "command_line_reporter", '~> 3.0'
 gem "ruby-terminfo", '~> 0.1.1'
 gem "diffy", '~> 3.1'
 gem "hrx", '~> 1.0'
+
+group :development do
+  gem 'rspec'
+end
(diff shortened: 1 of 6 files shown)


Gemfile
```diff
@@ -5,3 +5,7 @@ gem "command_line_reporter", '~> 3.0'
 gem "ruby-terminfo", '~> 0.1.1'
 gem "diffy", '~> 3.1'
 gem "hrx", '~> 1.0'
+
+group :development do
```
nex3: I'm not sure it makes sense to have a separate "development" section, since this will only ever really be run in development mode.

README.md
```diff
@@ -283,3 +283,10 @@ X. Exit testing.
 Any option can also be applied to all future occurences of that type of failure
 by adding `!` after it. For example, if you want to mark *all* failing specs as
 `:todo` for the current implementation you'd type `I!`.
+
+## Tests
+The unit tests are located in the `tests/` directory. To run the tests, run
```
nex3: It's probably a good idea to be clear here that these are unit tests *for the spec runner itself*, rather than tests for the Sass language.

Gemfile
rambleraptor: Totally fair!

PR #1458, opened 2019-08-17
Style-guide-ify specs for selector-append()

nex3:


spec/core_functions/selector/append.hrx
```diff
@@ -0,0 +1,240 @@
...
+Error: Can't append * to .c.
+  ,
+1 | a {b: selector-append(".c", "*")}
+  |       ^^^^^^^^^^^^^^^^^^^^^^^^^^
+  '
+  input.scss 1:7  root stylesheet
+
+<===> error/universal/output-libsass.css
+a {
+  b: .c*;
```
jathak: Is this valid both ways, or should this be a todo for libsass?

spec/core_functions/selector/append.hrx
```diff
@@ -0,0 +1,240 @@
...
+  - 1:1  root stylesheet
+  ,
+1 | a {b: selector-append(".c", "&")}
+  |       ^^^^^^^^^^^^^^^^^^^^^^^^^^
+  '
+  input.scss 1:7  root stylesheet
+
+<===> error/parent/output-libsass.css
+a {
+  b: .c;
```
jathak: Same as above

PR #1461, opened 2019-09-11
First full CLI test

rambleraptor:
This is the first unit test that calls the CLI from start-to-end.

I didn't want to have the unit tests have a hard dependency on a version of dart-sass. I built out a small (+ kind of hacky) utility that takes in a hrx file path + outputs the "output" version of that hrx (i.e. the same thing that dart-sass would output).

Let me know your thoughts!


rambleraptor:
I completely forgot to update this. So sorry about that!

I'm a little confused about how the stub can output additional metadata. We're expecting this stub to act like Dart Sass (but a much simpler version). Dart Sass takes in the name of a .scss file and outputs the resulting CSS to stdout.

The stub does the same thing without any knowledge of Sass. So, it takes in the name of a scss file (with no Sass) and outputs the CSS to stdout (which should be identical to the input).

If we print anything else to stdout, the test runner will start to fail unit tests because the stub no longer works like Dart Sass. If we print anything to stderr, we lose the ability to do compare errors.

nex3:
We don't need the stub to interact with any real tests, just the dummy tests we set up for testing purposes, right? So it doesn't *really* need to act like Dart Sass, except in the broadest sense that it reads input files and emits some sort of output. We can have that output be anything we want that facilitates testing, including diagnostics.

rambleraptor:
Right, but the sass-spec runner is going to compare the output of the stub
with the hrx file...

As I’m typing out that thought, I understand what you’re saying. You’re
saying that we can have the output.css portion of the hrx file be something
that’s not valid CSS. So, the hrx file would include the input metadata or
whatever else we want it to. I’ll make that happen. Thanks for the help!

rambleraptor:
Changes made! Thanks for the help

tests/fixtures/colors/basic.hrx
```diff
@@ -0,0 +1,9 @@
+<===> input.scss
+p {
+  color: #ff8000;
+}
+<===> output.css
+input: ["--precision", "10", "-t", "expanded", "-I", "/Users/alexstephen/sass-spec/spec", "input.scss"]
```
nex3: We should probably set this up so that that it doesn't require committing machine-specific paths.

rambleraptor: Oh yes, that's a great point!

rambleraptor: Changes made!

tests/sass_stub
```diff
@@ -0,0 +1,13 @@
...
+# This program stubs out the Sass CLI. It does not apply any Sass
+# transformations and simply outputs the input file to stdout.
+
+# Version flag needs to return a value.
+if ARGV.last == '-v'
+  puts '0.1'
+else
+  # sass-spec passes in full paths to some commands. We should
+  # use the last portion of the path instead.
+  args = ARGV.map { |a| a[0] == '/' ? a.split('/').last : a }
```
nex3: Won't this logic fail for Windows paths? Maybe a simpler thing to do would be to just look for `-I` and omit it and the following argument.

rambleraptor: Didn't think about! Paths omitted.

PR #1462, opened 2019-09-17
fix: unrecognized pseudos with arguments

jquense:
See sass/libsass#2990
[skip libsass]


nex3:
Closing due to inactivity.

PR #1465, opened 2019-09-24
Remove adjust-hue() from sass:color

nex3:
See sass/sass#2740

[skip dart-sass]


spec/core_functions/modules/color.hrx
```diff
@@ -319,6 +308,24 @@ Error: Undefined function.
...
   input.scss 2:7  root stylesheet
 
+<===>
+================================================================================
+<===> error/adjust_hue/input.scss
+@use "sass:color";
+a {b: color.adjust-hue(#abcdef, 10)}
+
+<===> error/adjust_hue/error
+Error: The function adjust-hue() isn't in the new module system.
```
nshahan: When would you expect someone to hit this message. Is it when someone tried to migrate their package by hand and messed it up? The migrator tool wont do this right? 

I think this message could be more clear for someone not following Sass development too closely. "adjust-hue() was not included in the color module in Sass 2.0.  Recommendation... "

nex3: > When would you expect someone to hit this message. Is it when someone tried to migrate their package by hand and messed it up? The migrator tool wont do this right?

That's right.

> I think this message could be more clear for someone not following Sass development too closely. "adjust-hue() was not included in the color module in Sass 2.0. Recommendation... "

I don't want to list a specific version number (partly because we don't know for sure which Dart Sass version the module system will launch in, partly because Sass *as a language* no longer has version numbers), but you're right that explicitly mentioning the color module is probably a good idea.

PR #1473, opened 2019-10-04
Add todo spec test for libsass issue 2994

mgreter:
https://github.com/sass/libsass/issues/2994

### DIFF
--- a/spec/libsass-todo-issues/issue_2994.hrx
+++ b/spec/libsass-todo-issues/issue_2994.hrx
@@ -1,0 +1,20 @@
+<===> input.scss
+.one-screen-page {
+	@extend %context-dark;
+}
+
+%context-dark {
+	.button-secondary-outline {
+		&:hover,
+		&:focus,
+		&:active,
+		&:hover {
+			color: #fca;
+		}
+	}
+}
+
+<===> output.css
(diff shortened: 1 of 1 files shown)


nex3:
This spec (along with many others you've landed over the past week) doesn't match the style guide and was landed without review. I've asked you before to avoid landing PRs like this, so I've switched on branch protection to require it from now on.

mgreter:
Yep, I did under the libsass directories, for all others I added you as reviewer. If that's your point, fine with me, means I wont contribute any more spec tests in the future. So be it that way.

nex3:
The ultimate goal of the style guide is to make *every* spec in the repository compliant, regardless of where it originated. This should produce a hierarchy of specs, organized by topic, that together cover the entire language. If we find that an implementation has a regression in behavior that the specs missed, we can add a test in a well-understood location that specifically targets that regression. We've already started doing so in cases like https://github.com/sass/sass-spec/pull/1464.

When new specs are added that aren't style-guide-compliant and don't fit into the global spec hierarchy, it just means more work to eventually migrate those specs in the future. Specs that are fairly large chunks of real-world Sass code are particularly difficult to migrate, because it's not clear what specific piece of functionality they cover.

PR #1481, opened 2019-10-07
Enable partially passing libsass spec tests

mgreter:
There is still one supposedly passing libsass todo test:
https://travis-ci.org/sass/libsass/jobs/594726983

The following tests pass but were marked as TODO for libsass:
sass-spec/spec/directives/use/error/extend/scope/private

No idea why it thinks this is a todo test ...
Quite certain this is a bug in ruby spec runner.

### DIFF
--- a/spec/core_functions/meta/load_css/with.hrx
+++ b/spec/core_functions/meta/load_css/with.hrx
@@ -263,6 +263,25 @@ c {
 
 <===>
 ================================================================================
+<===> core_module/indirect/input.scss
+// Regression test for sass/dart-sass#838.
+@use "sass:meta";
+@include meta.load-css("other", $with: (c: e));
+
+<===> core_module/indirect/_other.scss
+@use "sass:color";
+
+$c: d !default;
+
+a {b: $c}
+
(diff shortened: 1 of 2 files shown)


spec/core_functions/color/hsl/multi_argument_var.hrx
```diff
@@ -36,6 +29,11 @@ a {
 
 <===>
 ================================================================================
+<===> 1_of_2/options.yml
+---
+:todo:
+- libsass
```
nex3: Can you add issue references for any of these TODOs that don't have them yet?

mgreter: No I can't without spending substantial time to find them.
I should have preserved those that had references.
Or did I remove any of them?

nex3: If you don't know of an existing issue, can you file a new one? I understand that this came from an existing TODO without an issue reference, but since you're touching it anyway we should update it to point to the right place.

nex3:
Why did you delete this without merging?

PR #1489, opened 2019-10-25
Update specs for unquoting non-strings

Awjin:
This behavior is now fully deprecated. Make sure specs are expecting
errors now, instead of deprecation warnings.

See sass/libsass#3017

[skip libsass]

### DIFF
--- a/spec/libsass-closed-issues/issue_1106.hrx
+++ b/spec/libsass-closed-issues/issue_1106.hrx
@@ -22,15 +22,6 @@ Error: $string: null is not a string.
   '
   input.scss 7:14  root stylesheet
 
-<===> output.css
-a {
-  foo: bar;
-}
-
-<===> warning
-DEPRECATION WARNING: Passing null, a non-string value, to unquote()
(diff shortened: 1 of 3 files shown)


nex3:
If you're curious why the LibSass tests are failing (as I was), it's because:

* You're sending this pull request from your own fork of the repo, rather than from a branch of this repo,
* which causes Travis to consider this build untrusted,
* which causes it not to decrypt the encrypted environment variables, such as the read-only GitHub credentials,
* which means it's making unauthenticated GitHub requests,
* which causes those requests to fail due to rate-limiting,
* which means that it can't access the text of the pull request message (see https://github.com/travis-ci/travis-ci/issues/9302),
* which means it doesn't see `[skip libsass]` and know not to run the LibSass tests.

For future PRs, it's probably a good idea to push branches to the same repo to work around this.

Awjin:
Gotcha, thanks for the detailed explanation. To save us all a headache I'll cherrypick these changes (along with your requested changes) to a new branch on this repo.

PR #1490, opened 2019-10-25
Update specs for unquoting non-strings

Awjin:
This behavior is now fully deprecated. Make sure specs are expecting
errors now, instead of deprecation warnings.

See sass/libsass#3043

[skip libsass]


Awjin:
> I hadn't noticed that the latter two tests were actually ignored for Dart Sass. Now that LibSass's behavior matches Dart Sass's, I think you can just delete those tests entirely.

Agreed, done.

stof:
@nex3 should this be merged now that libsass is deprecated (and so the libsass PR was closed rather than being merged) ?

Awjin:
@stof thanks for the catch. Since libsass isn't doing any new feature work I'm closing this and the corresponding libsass PR.

PR #1492, opened 2019-11-04
Unit test for testing --run-todo flag

rambleraptor:
Unit test for testing the --run-todo flag.

This also fixes a couple bugs (2 test related, 1 not) that I found while I was building this test:

* Fixes a bug where the default implementation is not set. `sass-spec.rb --help` states: `--impl NAME Sets the name of the implementation being tested. Defaults to 'sass'`. The default value does not appear to be set anywhere. For using the --command flag, the implementation gets set to the full path of the executable by default.

* Fixes a bug in test_case_metadata.rb where it was deleting fixtures

* Changed the namespace of the past unit test fixtures because it didn't say anything useful about the fixture.

(I can split this up if it helps)


tests/cli_spec.rb
```diff
@@ -6,7 +6,23 @@
...
                 "'#{Dir.pwd}/tests/sass_stub' "\
-                "#{Dir.pwd}/tests/fixtures/colors"
+                "#{Dir.pwd}/tests/fixtures/basic"
     expect(last_command_started).to be_successfully_executed
   end
+
+  it 'should not run todo specs by default' do
+    run_command "#{Dir.pwd}/sass-spec.rb --command "\
+                "'#{Dir.pwd}/tests/sass_stub' "\
+                "#{Dir.pwd}/tests/fixtures/todo"
```
nex3: These `run_command` invocations have a lot of boilerplate... consider factoring out a helper function.

tests/fixtures/todo/basic.hrx
```diff
@@ -0,0 +1,25 @@
+<===> no-todo/input.scss
```
nex3: HRX style nits: real tests use underscores rather than hyphens in directory names, and end all files with a trailing newline.

tests/cli_spec.rb
rambleraptor: Completely agree! I've already got a PR going for the next test, so I'll make a helper function there.

PR #1493, opened 2019-11-05
Tests + bug fixes on --limit flag

rambleraptor:
This is a set of unit tests on the --limit flag.

It contains two bug fixes:

* Limit flag accepts negative numbers, despite what docs say (which leads to unintuitive behavior)
* When using limits, the runner would always run limit + 1 tests.

This also adds the Unit tests into Travis runs.

### DIFF
--- a/lib/sass_spec/test.rb
+++ b/lib/sass_spec/test.rb
@@ -8,7 +8,7 @@ require_relative "util"
 # Holder to put and run test cases
 class SassSpec::Test < Minitest::Test
   def self.create_tests(test_cases, options = {})
-    test_cases[0..options[:limit]].each do |test_case|
+    test_cases[0...options[:limit]].each do |test_case|
       define_method("test__#{test_case.name}") do
         runner = SassSpecRunner.new(test_case, options)
         test_case.finalize(runner.run)
--- a/lib/sass_spec/test.rb
+++ b/lib/sass_spec/test.rb
@@ -8,6 +8,7 @@ require_relative "util"
 # Holder to put and run test cases
 class SassSpec::Test < Minitest::Test
   def self.create_tests(test_cases, options = {})
+    options[:limit] = test_cases.length + 1 unless options[:limit]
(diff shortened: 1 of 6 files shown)


lib/sass_spec/test.rb
```diff
@@ -8,7 +8,8 @@
 # Holder to put and run test cases
 class SassSpec::Test < Minitest::Test
   def self.create_tests(test_cases, options = {})
-    test_cases[0..options[:limit]].each do |test_case|
+    options[:limit] = test_cases.length + 1 unless options[:limit]
+    test_cases[0...options[:limit]].each do |test_case|
```
rambleraptor: `..` on Ruby is inclusive-inclusive, while `...` is inclusive...exclusive (which is what you'd expect `..` to be)

lib/sass_spec/test.rb
```diff
@@ -8,7 +8,8 @@
 # Holder to put and run test cases
 class SassSpec::Test < Minitest::Test
   def self.create_tests(test_cases, options = {})
-    test_cases[0..options[:limit]].each do |test_case|
+    options[:limit] = test_cases.length + 1 unless options[:limit]
```
nex3:
```suggestion
    options[:limit] ||= test_cases.length + 1
```

tests/cli_spec.rb
```diff
@@ -4,25 +4,35 @@
 
 RSpec.describe 'run tests', type: :aruba do
   it 'runs a single spec' do
-    run_command "#{Dir.pwd}/sass-spec.rb --command "\
-                "'#{Dir.pwd}/tests/sass_stub' "\
-                "#{Dir.pwd}/tests/fixtures/basic"
+    run_command command('basic')
```
nex3: Why not something like `run_sass 'basic'` which includes the `run_command` invocation?

rambleraptor: That's a much, much better way of doing this.

PR #1494, opened 2019-11-09
Update specs to print call-site for @errors in mixins and functions

Awjin:
See sass/dart-sass#474
[skip dart-sass]

### DIFF
--- a/spec/libsass/error-directive-nested/function.hrx
+++ b/spec/libsass/error-directive-nested/function.hrx
@@ -18,8 +18,8 @@ Error: test
 <===> error-dart-sass
 Error: test
   ,
-2 |   @error test;
-  |   ^^^^^^^^^^^
+8 |     c: c();
+  |        ^^^
   '
-  input.scss 2:3  c()
+  input.scss 8:8  c()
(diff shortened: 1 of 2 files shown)


spec/libsass/error-directive-nested/function.hrx
```diff
@@ -18,8 +18,8 @@ Error: test
 <===> error-dart-sass
 Error: test
   ,
-2 |   @error test;
-  |   ^^^^^^^^^^^
+8 |     c: c();
+  |        ^^^
   '
-  input.scss 2:3  c()
+  input.scss 8:8  c()
```
nex3: Ideally, this wouldn't even include this line—that would give it the same stack trace as errors thrown by built-in functions:

```
Error: $string: 1 is not a string.
  ╷
2 │   a: unquote(1);
  │      ^^^^^^^^^^
  ╵
  test.scss 2:6  root stylesheet
```

If that's tricky to make work, though, don't worry too much about it.

Awjin: This is cleaner, I'll make it work.

PR #1496, opened 2019-11-14
Fix errors for overloaded function arguments

Awjin:
See sass/dart-sass#520
[skip dart-sass]

### DIFF
--- a/spec/core_functions/color/hsl/error/five_args.hrx
+++ b/spec/core_functions/color/hsl/error/five_args.hrx
@@ -1,9 +1,9 @@
-<===> error/too_many_args/input.scss
+<===> input.scss
 a {
   b: hsl(0, 100%, 50%, 0.5, 0);
 }
 
-<===> error/too_many_args/error
+<===> error
 Error: Only 4 arguments allowed, but 5 were passed.
   ,
 2 |   b: hsl(0, 100%, 50%, 0.5, 0);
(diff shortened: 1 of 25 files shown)


spec/core_functions/color/hsl/error/zero_args.hrx
```diff
@@ -0,0 +1,12 @@
+<===> input.scss
```
nex3: Good catch!

spec/core_functions/color/hsl/error/zero_args.hrx
```diff
@@ -0,0 +1,12 @@
...
+zero-args {
```
nex3: Even though the tests around this predate [the style guide](https://github.com/sass/sass-spec/blob/master/STYLE_GUIDE.md), these new tests should follow it.

Awjin: Done. I know it's frowned upon to impose new style on old code, but I saw enough low-hanging fruit to justify cleaning up the surrounding tests too.

spec/core_functions/color/hsl/error/five_args.hrx
```diff
@@ -1,19 +1,19 @@
-<===> input.scss
-five-arg {
-  x: hsl(0, 100%, 50%, 0.5, 0);
+<===> error/too_many_args/input.scss
```
nex3: These are already in an error directory (spec/core_functions/color/hsl/error) so they don't need an additional error component here. I think it's fine to also leave off too_many_args, since there's only one meaningful test for passing five arguments anyway.

Awjin: Good points, fixed.

PR #1497, opened 2019-11-19
Add specs for configuration through imports

jathak:
See sass/sass#2772

[skip dart-sass]

### DIFF
--- a/spec/directives/import/configuration.hrx
+++ b/spec/directives/import/configuration.hrx
@@ -1,0 +1,135 @@
+<===> options.yml
+---
+:todo:
+- sass/libsass#3040
+
+<===>
+================================================================================
+<===> same-file/input.scss
+$d-a: configured;
+@import "other";
+
+<===> same-file/_other.scss
+$a: original !default;
+b {c: $a}
+
+<===> same-file/_other.import.scss
+@forward "other" as d-*;
+
+<===> same-file/output.css
+b {
+  c: configured;
+}
+
+<===>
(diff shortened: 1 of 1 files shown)


jathak:
@nex3: What additional specs should be added for this?

spec/directives/import/configuration.hrx
```diff
@@ -0,0 +1,135 @@
...
+  c: configured;
+}
+
+b {
+  c: configured;
+}
+
+<===>
+================================================================================
+<===> import-twice/with-change/input.scss
```
nex3: I think you also want a test that verifies that, after the second import, the variable does have the new value even though the module hasn't been reloaded.

nex3:
It just occurred to me that you should probably also add a spec for a local variable set for a nested import.

nex3:
Also, add a spec that verifies no error is thrown if a variable is set in the import context that *isn't* defined in the forwarded module.

nex3:
And based on our conversation in the implementation CL:

* A spec that verifies that, if the midstream file overrides the input's definition of the variable, the input's definition is still used in the configuration.
* A spec that verifies that, if the midstream file defines a new variable with the same name as a configurable variable, it doesn't get included in the config.

jathak:
Okay, I think I added all the tests you suggested. The one for confirming that a declaration between duplicate `@import` rules still sets it in the current file is todo for now because of sass/dart-sass#888.

PR #1500, opened 2019-12-20
Specs for clamp() and hypot()

Awjin:
Adds specs for new built-in math functions.

[skip dart-sass]

### DIFF
--- a/spec/core_functions/math/clamp.hrx
+++ b/spec/core_functions/math/clamp.hrx
@@ -1,0 +1,292 @@
+<===> options.yml
+---
+:todo:
+- sass/libsass#3044
+
+<===>
+================================================================================
+<===> chooses_min/input.scss
+@use "sass:math" as math;
+a {b: math.clamp(1, 0, 2)}
+
+<===> chooses_min/output.css
+a {
+  b: 1;
+}
+
+<===>
+================================================================================
+<===> chooses_number/input.scss
+@use "sass:math" as math;
+a {b: math.clamp(0, 1, 2)}
+
+<===> chooses_number/output.css
+a {
+  b: 1;
+}
+
+<===>
+================================================================================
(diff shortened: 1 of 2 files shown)


spec/core_functions/modules/math/abs.hrx
```diff
@@ -0,0 +1,125 @@
+<===>
+================================================================================
+<===> options.yml
+---
+:todo:
+- libsass
```
nex3: We want to have as much functionality tested on LibSass as possible, so for functions that do have global definitions we should continue to test them globally rather than through `@use`. Similarly, we should keep `spec/core_functions/modules/math.hrx` around until all implementations support `@use` and we can switch the tests to use it by default.

spec/core_functions/modules/math/clamp.hrx
```diff
@@ -0,0 +1,171 @@
+<===>
+================================================================================
+<===> options.yml
```
nex3: I think this should go in `spec/core_functions/math` even though it uses the modules, so it's in the same place as the other core function tests.

spec/core_functions/modules/math/abs.hrx
Awjin: How do you feel about letting this new directory thoroughly test the module functions, and keeping `core_functions/math/` around to exercise the global/LibSass functions?

Once everyone supports `@use` we could delete LibSass-specific stuff from `core_functions/math/` and have those specifically for globals.

The con is that we have to maintain both `core_functions/modules/math/` and `core_functions/math/`, but I don't anticipate these functions changing much.

Awjin: (There are several incoming specs for exponential and trig functions that need a home in `modules/math`, and if that directory is inevitable, my personal preference would be to lay the foundation here.)

PR #1502, opened 2019-12-27
Add specs for exponential math functions.

Awjin:
[skip dart-sass]


spec/core_functions/math/log.hrx
```diff
@@ -0,0 +1,162 @@
+<===> options.yml
+---
+:todo:
+- sass/libsass#3044
+
+<===>
+================================================================================
+<===> input.scss
```
nex3: A test file shouldn't have subdirectories that also contain tests (this isn't in the style guide but it probably should be), so this should be in its own directory. I'd call it `positive` to match the other specs here.

spec/core_functions/math/log.hrx
```diff
@@ -0,0 +1,162 @@
...
+a {b: math.log(2, 0)}
+
+<===> base/zero/output.css
+a {
+  b: 0;
+}
+
+<===>
+================================================================================
+<===> base/one/input.scss
```
nex3: Test a base between zero and one?

spec/core_functions/math/log.hrx
```diff
@@ -0,0 +1,162 @@
...
+a {b: math.log(2, 1)}
+
+<===> base/one/output.css
+a {
+  b: Infinity;
+}
+
+<===>
+================================================================================
+<===> error/number_has_units/input.scss
```
nex3: Don't forget to test type errors as well.

spec/core_functions/math/log.hrx
```diff
@@ -0,0 +1,162 @@
...
+@use "sass:math" as math;
+a {b: math.log(1px, 1px)}
+
+<===> error/number_and_base_have_units/error
+Error: Expected $number to have no units.
+  ,
+2 | a {b: math.log(1px, 1px)}
+  |       ^^^^^^^^^^^^^^^^^^
+  '
+  input.scss 2:7  root stylesheet
```
nex3: If you wanted to, I think you could omit this.

spec/core_functions/math/pow.hrx
```diff
@@ -0,0 +1,320 @@
+<===> options.yml
+---
+:todo:
+- sass/libsass#3044
+
+<===>
+================================================================================
+<===> input.scss
+@use "sass:math" as math;
+a {b: math.pow(2, 3)}
```
nex3: It's probably also a good idea to test a real number with a fractional exponent as well.

nex3:
Another thing I forgot—for functions like `log()` that take optional arguments that default to `null`, it's good to add a test for explicitly passing `null` to ensure that some implementation doesn't accidentally make it default to something else (in this case, `$e`).

nex3:
One more thing I forgot (and I may have missed this in the last review, too): all built-in function tests should test passing arguments by name, to ensure that all implementations use the same argument names.

Awjin:
> For functions that have special behavior for certain real numbers, it's probably a good idea to test that they behave that way for numbers that Sass considers equivalent to those numbers (within the precision limit) as well. For example, `math.log(2, 0.000000000001)` should return `-Infinity`, and `math.pow(1.000000000001, Infinity)` should return `NaN`.

Should Sass round between operations? i.e. does `log(log(2))` evaluate to `log(0.30102999566`...`)` or `log(0.3010299957)`?

spec/core_functions/math/log.hrx
Awjin: Done.

spec/core_functions/math/log.hrx
Awjin: Done.

spec/core_functions/math/log.hrx
Awjin: Done.

spec/core_functions/math/log.hrx
Awjin: Done.

spec/core_functions/math/pow.hrx
Awjin: Done.

PR #1504, opened 2020-01-07
Add specs for trigonometric math functions.

Awjin:
[skip dart-sass]


spec/core_functions/math/acos.hrx
```diff
@@ -0,0 +1,126 @@
...
+  input.scss 2:7  root stylesheet
+
+<===>
+================================================================================
+<===> error/units/input.scss
+@use "sass:math" as math;
+a {b: math.acos(1px)}
+
+<===> error/units/error
+Error: Expected 1px to have no units.
```
nex3:
```suggestion
Error: $number: Expected 1px to have no units.
```

Also below.

spec/core_functions/math/atan2/arguments.hrx
```diff
@@ -0,0 +1,138 @@
+<===> options.yml
+---
+:todo:
+- sass/libsass#3044
```
nex3: Rather than giving each subdirectory of `math/atan2` an identical `options.yml` file, just put one directly in `math/atan2`. It'll apply to everything beneath it.

PR #1505, opened 2020-01-11
Throw multi-span exceptions

nex3:
This shouldn't land until after dart-lang/source_span#49 is released.

[skip dart-sass]


nex3:
@nshahan Everything within `spec/core_functions` is essentially the same (other than a few meta functions), so you can skim it.

PR #1511, opened 2020-02-07
Don't throw conflict errors for identical members

nex3:
See sass/sass#2820

[skip dart-sass]


spec/directives/use/error/member.hrx
```diff
@@ -244,6 +244,33 @@ Error: This mixin is available from multiple global modules.
...
+a {b: $c}
+
+<===> conflict/same_value/_other1.scss
+$c: d;
+
+<===> conflict/same_value/_other2.scss
+$c: d;
+
+<===> conflict/same_value/error
+Error: This variable is available from multiple global modules.
```
nshahan: Is it possible to reference the variable here like in the error for `@forward`:

`Error: Two forwarded modules both define a variable named $a.`

It would make this error sound less vague.

nex3: The idea behind not doing that is that the variable name will always be included in the highlighted text. I try not to include information in the error messages that's redundant with the highlight, because it increases the amount of text the user has to read without actually increasing the information it's conveying. That said, if you still think it would be clearer, I could update it.

nshahan: I'm fine either way and it's up to personal preference, but personally I think it's clearer if there is more detail up front, and then the various source spans show the evidence. That way I know what I'm looking for in those spans.

nex3: Since this is kind of a philosophical question about how to approach error messaging in general, I'll leave this as-is for now but put it on the docket for a future design meeting.

Issue #1516, opened 2020-02-26
Restructure directories to conform to style guide


Awjin:
The directory structure has grown organically over time, and as such does not completely adhere to the recommendations set in the [style guide](https://github.com/sass/sass-spec/blob/master/STYLE_GUIDE.md#do-use-descriptive-paths).

For example, some directories have generic names that don't describe their function (`basic/`, `misc/`, etc). Some are redundant and could be combined with other directories (`colors/`, `mixin/`, `operations/`, etc). Some files are at root and could be reorganized into sub-directories.

We should decide on an ideal directory structure, and move specs into that structure.

saper:
https://github.com/sass/sass-spec/pull/1484 has a discussion whether we need regression tests (organized by issue, not by the feature they test) in the repository as well.

Awjin:
Thanks for the context. My personal opinion is that regression tests should be added to the *feature*, so that over time the feature specs become more robust.

Regardless, the bulk of reorganization can be done without touching the regression tests, so I think this work should proceed in parallel.

saper:
Thank you. Yes, I fully agree that there is enough work to be done in the spec (feature) tests alone.

Awjin:
Here's a first pass at a new structure (excluding `libsass` directories):

## Doesn't need reorg:

```
core_functions/
css/
directives/
operations/
values/
```

## Needs reorg:

### Move redundant directories into existing directories

```
colors/ —> values/colors/
extend-tests/ -> directives/extend/
media-import.hrx -> directives/import/
mixin/ -> directives/mixin/
```

### Move generic specs into the feature that they actually test

```
basic/
errors/
misc/
sass/
sass_4_0/
scss/
scss-tests/
```

### New top-level directories

* `variables/`

```
variables.hrx -> variables/
scope/ -> variables/scope/
```

* `parser/selectors/` becomes top-level `selectors/`

### Rename parser/ to syntax/, then move files into it

```
nesting/ -> syntax/nesting/
sass/, scss/ -> Instead of grouping syntax weirdness into sass/ and scss/,
                have a spec for each feature that causes the syntax to behave
                weirdly, then for each feature spec have both sass/ and scss/
                inputs.
```

## Proposed directory structure

```
css/
core_functions/
directives/
operations/
selectors/
syntax/
values/
variables/
```

Awjin:
I'm struggling most with the somewhat generic `syntax` folder—could this be made more explicit, or split up even further?

[...]

PR #1520, opened 2020-02-28
Move non-style-guide-compliant specs into non_conformant/

Awjin:
This change moves all older specs, except for `libsass*` specs, into a `non_conformant/` directory.

It does not otherwise add or delete directories.

See https://github.com/sass/sass-spec/issues/1516


nex3:
Just a couple nits:

* It's probably a good idea to include `non_conformant/README.md` to explain to a casual reader what's up with that directory.

* I'd name this PR something like "move non-style-guide-compliant specs into non_conformant/", since it doesn't fully create the conformant directories.

spec/non_conformant/README.md
```diff
@@ -0,0 +1,9 @@
+# `non_conformant/`
+
+## Specs that don't conform to the style guide
```
nex3: I don't think you need either of these headers—the location of the README is probably enough to indicate what it's referring to.

Awjin: Done.

stof:
@Awjin by curiosity, why not moving the `libsass*` folders ? AFAICT, these are also non conformant specs.

Awjin:
LibSass had a separate workflow for their tests, so I didn't want to touch the LibSass folders in this change. That being said, now that LibSass is deprecated, we will need a path forward for managing those tests.

Issue #1418, opened 2019-06-28
Add comprehensive equality specs


nex3:
Currently, specs for value equality (like most specs) are spread haphazardly throughout the repo. We should organize them consciously and make sure all edge cases are being tested. Equality specs for each kind of value should live in that value's directory (`values/$type/equality.hrx`), and they should share `assert-equal()` and `assert-unequal()` mixins that verify not only that the values are equal according to `==` but that all functions that care about equality (`index()` and various map functions) consider them equal as well.

stof:
@nex3 I think a comprehensive set of specs also need to include another kind of tests: tests about equality between different types of values (where different types are never equal, except for the empty list and the empty map).

PR #1524, opened 2020-03-03
Color specs

Awjin:
See https://github.com/sass/sass-spec/issues/1516

Adds specs for the value `color`.


spec/operators/inequality/color.hrx
```diff
@@ -0,0 +1,106 @@
+<===> boolean/true/input.scss
```
nex3: Rather than have separate inequality tests that wholly duplicate the equality tests, it's probably better to write a helper mixin that verifies that two values are equal according to both `==` and `!=`. I think it's reasonable to delay this until https://github.com/sass/sass-spec/issues/1418 as well, in which case maybe just leave out inequality specs for now.

spec/values/color.hrx
```diff
@@ -0,0 +1,47 @@
+<===> keyword/input.scss
```
nex3: I think these specs should be substantially expanded:

* We should verify that these values are actually parsed as colors, not just that they're emitted as written. Maybe pass them to `adjust-color()` with no other args, which will both verify that its argument is a color and return a color value without an associated string representation?

* We need to all color names Sass supports, and that those names are case-insensitive.

* We need to verify that hex digits are parsed correctly into red/green/blue/alpha channels for all hex representations.

PR #1528, opened 2020-04-30
Invalid CSS generated with 2.3.0 (Nesting + child combinator + selector lists + interpolation)

ahorek:
https://github.com/sass/libsass/issues/3083

### DIFF
--- a/spec/libsass/nested_interpolation.hrx
+++ b/spec/libsass/nested_interpolation.hrx
@@ -1,0 +1,14 @@
+<===> input.scss
+$abc: 'qux';
+
+.foo {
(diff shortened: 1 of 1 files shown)


spec/libsass/nested_interpolation.hrx
```diff
@@ -0,0 +1,14 @@
+<===> input.scss
```
nex3, line 1: Please update this spec to follow [the style guide](https://github.com/sass/sass-spec/blob/main/STYLE_GUIDE.md), including putting it in a directory that matches the part of the language it tests.

nex3 (changes requested):

PR #1529, opened 2020-05-01
Fix compound selector children output order

mgreter:
While fixing https://github.com/sass/libsass/issues/3084 I came across these.

### DIFF
--- a/spec/non_conformant/basic/12_pseudo_classes_and_elements.hrx
+++ b/spec/non_conformant/basic/12_pseudo_classes_and_elements.hrx
@@ -2,7 +2,7 @@
 a b {
   color: red;
   :first-child, :nth-of-type(  -2n+1 ) {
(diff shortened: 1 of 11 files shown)


mgreter:
Minimal reproduction case:
```scss
:before {
    &.red-icon {
        foo: bar;
    }
}
```

nex3:
What do these spec changes have to do with sass/libsass#3084 or with your reproduction case? None of them involve pseudo-elements like `:before`, and all of them look wrong to me—they're rearranging compound selectors to be in a different order than the user originally wrote them.

mgreter:
IMHO we should either throw an error or reorder invalid selectors provided by users (a warning in that case may also be useful), but when we do parentization we should probably always generate valid css.

nex3:
To keep things in one place I'm going to follow up on this at https://github.com/sass/dart-sass/issues/996.

PR #1530, opened 2020-05-01
test pseudo element order

ahorek:
for https://github.com/sass/dart-sass/issues/996


Awjin:
The current behavior is working as intended.

mgreter:
Please note that this used to work in ruby sass.

PR #1532, opened 2020-05-12
Add specs for comments within a selector.

jathak:
See sass/dart-sass#1002.

[skip dart-sass]

### DIFF
--- a/spec/css/selector.hrx
+++ b/spec/css/selector.hrx
@@ -69,6 +69,11 @@
 
 <===>
 ================================================================================
+<===> inline_comments/options.yml
+---
+:todo:
+- sass/libsass#3124
+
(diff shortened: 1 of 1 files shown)


spec/css/selector.hrx
```diff
@@ -55,6 +55,65 @@
...
+b {
+  x: y;
+}
+
+<===>
+================================================================================
+<===> inline_comments/silent/with_comma_in_comment/input.sass
+a // comment,
+b
+  x: y
```
nex3, line 76: Consider [adding a comment](https://github.com/sass/sass-spec/blob/master/STYLE_GUIDE.md#consider-adding-a-comment-explaining-your-spec) explaining why this works the way it does.

PR #1535, opened 2020-05-21
Add libsass warnings for invalid selectors

mgreter:
I would like to warn users if we encounter invalid selectors.
I believe this is always unwanted and can lead to hard to debug issues.

### DIFF
--- a/spec/core_functions/selector/extend/no_op.hrx
+++ b/spec/core_functions/selector/extend/no_op.hrx
@@ -46,6 +46,11 @@ a {
   b: ::c.d;
 }
 
+<===> conflict/pseudo_element/unknown/warning-libsass
+WARNING on line 1, column 30 of /sass/spec/core_functions/selector/extend/no_op/conflict/pseudo_element/unknown/input.scss:
+LibSass detected an invalid selector order: "::c.d"
+This will make any group of selectors containing it invalid!
+
 <===>
 ================================================================================
 <===> conflict/pseudo_element/class_syntax/input.scss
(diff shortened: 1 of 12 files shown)


nex3:
This arguably violates the Sass principle of knowing as little about CSS as possible to generate correct output. There's no guarantee that CSS won't add a pseudo-element selector in the future that *does* have selectable attributes. I'm not even 100% confident that no such element exists today.

It's also strange to warn users about this particular potential mistake when Sass generally doesn't warn users about any other similar genre of mistake. Why stop here? Why not warn if they're using `*` in a descendant selector, or multiple IDs in the same compound selector, or even using an undefined psuedo-class or -element?

The answer, I believe, is that correctness warnings are the job of a dedicated tool like a linter, and that a preprocessor should focus on making transformations to the CSS source as given. Sass can never do as good a job of warning about incorrect selectors as stylelint, and giving users the expectation that Sass will protect them in this way can set them up for failure when it inevitably does not.

mgreter:
For 1) `Note: In general, a <pseudo-element-selector> is only valid if placed at the end of the last <compound-selector> in a <complex-selector>. `.

https://developer.mozilla.org/en-US/docs/Web/CSS/Pseudo-elements

mgreter:
For the rest, I kinda don't really care anymore, so feel free to close this PR!!!

mgreter:
> §5.2 The universal selector follows the same syntax rules as other type selectors: only one can appear per compound selector, and it must be the first simple selector in the compound selector.

> A compound selector is a sequence of simple selectors that are not separated by a combinator, and represents a set of simultaneous conditions on a single element. If it contains a type selector or universal selector, that selector must come first in the sequence. Only one type selector or universal selector is allowed in the sequence. (A compound selector is represented by <compound-selector> in the selectors grammar.) A given element is said to match a compound selector when it matches all simple selectors in the compound selector.

The suggested warning was exactly about those two cases, pseudo-selector not last, and multiple tag/universal selectors not at the start.

PR #1536, opened 2020-05-27
Add specs for more escaped at-rules

jathak:
See sass/dart-sass#1011.

[skip dart-sass]


spec/directives/if-else.hrx
```diff
@@ -0,0 +1,35 @@
...
+@Else {a {b: c}}
+
+<===> capitalized_else/output.scss
+@Else {
+  a {
+    b: c;
+  }
+}
+
+<===> escaped/options.yml
```
nex3, line 12: Style nit: [use comment dividers to separate specs](https://github.com/sass/sass-spec/blob/master/STYLE_GUIDE.md#do-use-comment-dividers-to-separate-specs-within-an-hrx-file).

jathak:
What would be the best file structure for the escaped tests? Right now `@function` and `@if` each have their own directory with `escaped.hrx` containing a single test. Would it make more sense to just have a single `escaped.hrx` file in `directives` that contains tests for all the rules?

nex3:
Per the style guide, [specs should be ordered by the part of the language they test](https://github.com/sass/sass-spec/blob/master/STYLE_GUIDE.md#do-organize-specs-by-which-part-of-the-language-they-test) rather than by which aspect of that feature they test, so I think it makes sense to keep these files in their directive directories.

Generally, I recommend trying to think of the repo as one homogeneous file hierarchy, regardless of which levels of that hierarchy are written in HRX files versus physical directories. We choose which directories to make HRX versus physical based on [how many lines](https://github.com/sass/sass-spec/blob/master/STYLE_GUIDE.md#dont-have-hrx-files-longer-than-500-lines-or-so) the HRX file would contain, which means that sometimes it leaves orphan tests that don't fit in any larger subdirectory and end up in their own single-test HRX files. It's a little awkward, but ultimately I think it's better than trying to shoehorn those tests elsewhere just to avoid having small files.

spec/directives/if_else/escaped.hrx
```diff
@@ -0,0 +1,24 @@
+<===> options.yml
```
nex3, line 1: Rather than making a new `if_else` directory, I'd probably just test `@else` within the `if` directory (since it's inherently tightly coupled).

PR #1537, opened 2020-06-01
Refactor non-conformant selector specs

jathak:


spec/css/selector.hrx
```diff
@@ -55,6 +55,58 @@
...
+<===> placeholder/pseudoselectors/options.yml
+---
+:todo:
+- sass/libsass#2422
+
+<===> placeholder/pseudoselectors/matches/input.scss
+// Style rules with placeholders should only be deleted if they can't match
+// *any* real elements.
+a:matches(%b) {w: x}
+a:matches(%b, c) {y: z}
```
nex3, line 81: These should be [separate test cases](https://github.com/sass/sass-spec/blob/master/STYLE_GUIDE.md#do-test-only-one-thing-per-spec). Same for the `:not` tests.

spec/css/selector.hrx
```diff
@@ -55,6 +55,85 @@
...
+<===>
+================================================================================
+<===> placeholder/pseudoselectors/matches/solo/options.yml
+---
+:todo:
+- sass/libsass#2422
+
+<===> placeholder/pseudoselectors/matches/solo/input.scss
+// Style rules with placeholders should be deleted if they can't match any real
+// elements.
```
nex3, line 79: Rather than repeating this across multiple tests, it's probably a good idea to either:

* Update the comment to relate to the specific behavior each test is covering (as it differs from other tests), or...

* Put the tests that need the same comment in a shared (virtual) directory so they can share a single `README.md` file with this comment.

jathak: Done. I've added specific comments for each test.

Issue #1517, opened 2020-02-26
Add executable that prints information about the specs


Awjin:
It would be helpful to generate reports on the status of the repo. Some examples of information that would be useful are:

* How many specs are marked as TODO for which implementations?
* Where are all the specs that correspond to this GitHub issue?

PR #1541, opened 2020-06-04
Add stats executable for reporting repo state.

Awjin:
Closes https://github.com/sass/sass-spec/issues/1517


lib/sass_spec/stats/todo.rb
```diff
@@ -0,0 +1,240 @@
...
+      todos_missing_issue(files)
+    else
+      todos(files)
+    end
+  end
+
+  def todos_by_implementation(files, implementation)
+    regexes = []
+    if implementation == "dart-sass"
+      dart_sass = Regexp.new(':todo:(?:\n- libsass)*?(?:\n- sass\/libsass)*?\n- sass\/dart-sass#\d+$', Regexp::MULTILINE)
```
nex3, line 74: Why not search the parsed YAML contents of these files rather than use regexps? It seems like that would be more robust and probably easier to read as well.

lib/sass_spec/stats/todo.rb
```diff
@@ -0,0 +1,240 @@
...
+      libsass,
+      libsass_missing_issue,
+    ])
+    print_report(todos)
+  end
+
+  def find_todos(files, regexes)
+    todos = {}
+
+    files.each do |file|
```
nex3, line 124: I'd recommend using `SassSpec::Directory#glob` here to find all the `options.yml` files. It already has built-in logic for normalizing away the differences between files in HRX and physical files.

lib/sass_spec/stats/todo.rb
```diff
@@ -0,0 +1,240 @@
...
+      matches.uniq.each do |match|
+        entry = {}
+        if match =~ /dart-sass/
+          impl = "dart-sass"
+        elsif match =~ /libsass/
+          impl = "libsass"
+        end
+        entry[:impl] = impl
+        entry[:issue] = match =~ /#/ ? match.split("#")[-1] : "?"
+        todos[path].push(entry)
```
nex3, line 144: Not necessarily something that needs to be fixed in this PR, but something that will probably need to be fixed at some point, is changing this to count the number of *specs* that are marked as TODO rather than counting the number of *options files*. A single options file at the root of a directory applies to every spec beneath that directory, so eventually you'll probably need to search for all tests (identified by `input.{scss,css,sass}` files) within the directory that contains the `options.yml`.

lib/sass_spec/stats/todo.rb
```diff
@@ -0,0 +1,240 @@
...
+    end
+
+    return todos
+  end
+
+  def print_report(todos)
+    longest_path = 0
+    todos.keys.each do |path|
+      longest_path = [longest_path, path.length].max
+    end
```
nex3, line 155:
```suggestion
    longest_path = todos.keys.map(&:length).max
```

Bummer that `command_line_reporter` doesn't handle this for you.

lib/sass_spec/stats/todo.rb
```diff
@@ -0,0 +1,240 @@
...
+
+      matches.uniq.each do |match|
+        entry = {}
+        if match =~ /dart-sass/
+          impl = "dart-sass"
+        elsif match =~ /libsass/
+          impl = "libsass"
+        end
+        entry[:impl] = impl
+        entry[:issue] = match =~ /#/ ? match.split("#")[-1] : "?"
```
nex3, line 143: I'm not a huge fan of using arbitrary strings as meaningful placeholders. Consider just letting `entry[:issue]` be `nil` if there is no issue.

lib/sass_spec/stats/todo.rb
```diff
@@ -0,0 +1,240 @@
...
+    parser.parse!(args)
+    new(options)
+  end
+
+  def initialize(options)
+    @options = options
+  end
+
+  def stats
+    files = Dir.glob(File.join(SassSpec::SPEC_DIR, '**/*')).uniq
```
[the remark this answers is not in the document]
Awjin: You're right, there shouldn't be any duplicates here.

lib/sass_spec/stats/todo.rb
Awjin: Great idea, thanks

lib/sass_spec/stats/todo.rb
Awjin: Thanks for pointing me to this!

lib/sass_spec/stats/todo.rb
```diff
@@ -0,0 +1,240 @@
...
+
+class SassSpec::Stats::Todo
+  include CommandLineReporter
+
+  def self.parse(args)
+    options = {}
+    parser = OptionParser.new
+    parser.banner = "Usage: ./sass-spec.rb todo [filter]\n\nThis sub command prints stats on todos.\n\n"
+
+    parser.on("--impl IMPLEMENTATION",
```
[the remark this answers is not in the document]
Awjin: That makes sense to me

lib/sass_spec/stats/todo.rb
```diff
@@ -0,0 +1,240 @@
...
+      if !["dart-sass", "libsass"].include? impl
+        warn "Implementation must be one of [dart-sass, libsass].\n\n"
+        warn parser.help()
+        return nil
+      end
+      options[:impl] = impl
+    end
+
+    parser.on("--issue GITHUB_ISSUE",
+              "Find todos corresponding to a Github issue in the form `dart-sass#123`.") do |issue|
```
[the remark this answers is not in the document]
Awjin: Good point!

lib/sass_spec/stats/todo.rb
```diff
@@ -0,0 +1,240 @@
...
+      end
+
+      if !options[:issue]
+        warn "Invalid issue.\n\n"
+        warn parser.help()
+        return nil
+      end
+    end
+
+    parser.on("--missing-issue", "Find todos that are missing a Github issue.") do
```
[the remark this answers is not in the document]
Awjin: Hmm afaict OptionParser doesn't support this

lib/sass_spec/stats/todo.rb
```diff
@@ -0,0 +1,240 @@
...
+
+    if @options[:impl]
+      todos_by_implementation(files, @options[:impl])
+    elsif @options[:issue]
+      todos_by_issue(files, @options[:issue])
+    elsif @options[:missing_issue]
+      todos_missing_issue(files)
+    else
+      todos(files)
+    end
```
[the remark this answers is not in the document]
Awjin: Agreed, function approach is much better here

lib/sass_spec/stats/todo.rb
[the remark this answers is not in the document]
nex3: You can handle it manually by checking if `options[:missing_issue]` and `options[:issue]` are both set.

Awjin:
I addressed most comments, printing a nicely formatted report and specs are in flight.

lib/sass_spec/stats/todo.rb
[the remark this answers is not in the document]
nex3: Sorry, I think I mixed up some flag names and wasn't totally clear here. I was thinking you could combine `--impl` (now `--todo`) and `--issue` into one flag named `--todo`, so the user can pass either `--todo dart-sass` or `--todo dart-sass#123` without having to remember which format went with which flag.

lib/sass_spec/stats.rb
```diff
@@ -0,0 +1,146 @@
...
+              "Find specs that correspond to a Github issue.") do |issue|
+      unless issue =~ /.+#\d+/
+        warn("Invalid issue.\n\n")
+        puts parser.help()
+      end
+      issue = issue.split("#")
+      number = issue.pop
+      issue = issue.join().split("/")
+      impl = issue.pop
+      options[:issue] = "#{impl}##{number}"
```
nex3, line 35:
```suggestion
      unless issue =~ /([^/]+#\d+)$/
        warn("Invalid issue.\n\n")
        puts parser.help()
      end
      options[:issue] = $1
```

lib/sass_spec/stats.rb
```diff
@@ -0,0 +1,146 @@
...
+    parser.parse!(args)
+    options[:dirs] = args.dup.map{ |d| File.expand_path(d) } if args.any?
+    new(options)
+  end
+
+  def initialize(options)
+    @options = options
+  end
+
+  ##
```
nex3, line 57: It's probably best to follow the doc comment style in the rest of the repo, which doesn't have this leading `##`. I think you can omit the `@param` and `@returns` as well.

lib/sass_spec/stats.rb
```diff
@@ -0,0 +1,146 @@
...
+    end
+
+    if @options[:issue]
+      metadata = filter_todos_by_issue(metadata)
+    end
+
+    if @options[:missing_issue]
+      metadata = filter_todos_missing_issue(metadata)
+    end
+  end
```
nex3, line 74: Is this supposed to return `metadata` as well?

lib/sass_spec/stats.rb
```diff
@@ -0,0 +1,146 @@
...
+
+  ##
+  # Given the options provided to the CLI, filters through spec metadata to
+  # aggregate statistics about the specs. Prints a stats report.
+  def stats
+    metadata = get_metadata()
+
+    if @options[:todo]
+      metadata = filter_todos_by_implementation(metadata)
+    end
```
nex3, line 65: Style nit:

```suggestion
    metadata = filter_todos_by_implementation(metadata) if @options[:todo]
```

This will let you write what currently takes 11 lines in 3 without losing much readability :smiley:.

lib/sass_spec/stats.rb
```diff
@@ -0,0 +1,146 @@
...
+  ##
+  # Gets the raw YAML representation of a spec's metadata. Useful when
+  # inspecting for patterns that get normalized away by TestCaseMetadata.
+  #
+  # @param [SassSpec::TestCaseMetadata] metadata
+  # @return [YAML]
+  def get_yaml(metadata)
+    dir = SassSpec::Directory.new(metadata.name)
+    YAML.load(dir.read("options.yml"))
+  end
```
nex3, line 124: Rather than re-reading the metadata here, maybe it would be better to have the Metadata object store the raw options map.

lib/sass_spec/stats.rb
```diff
@@ -0,0 +1,146 @@
...
+  ##
+  # Filter a set of metadata to those with a todo for the given implementation.
+  #
+  # @param [Array<SassSpec::TestCaseMetadata>]
+  # @return Array<SassSpec::TestCaseMetadata>]
+  def filter_todos_by_implementation(metadata)
+    metadata.select do |meta|
+      todos = meta.options[:todo]
+      next unless todos
+      todos.select { |todo| todo.include? @options[:todo] }.any?
```
nex3, line 85: Can you use `Metadata#todo?` here?

lib/sass_spec/stats.rb
```diff
@@ -0,0 +1,146 @@
...
+    dir = SassSpec::Directory.new(metadata.name)
+    YAML.load(dir.read("options.yml"))
+  end
+
+  ##
+  # Gets all of the metadata in the spec directories.
+  #
+  # @return [Array<SassSpec::TestCaseMetadata>]
+  def get_metadata
+    dirs = @options[:dirs] || [SassSpec::SPEC_DIR].map do |dir|
```
nex3, line 131:
```suggestion
    dirs = (@options[:dirs] || [SassSpec::SPEC_DIR]).map do |dir|
```

lib/sass_spec/stats.rb
```diff
@@ -0,0 +1,146 @@
...
+      SassSpec::Directory.new(dir)
+    end
+
+    option_files = dirs.flat_map do |dir|
+      dir.glob("**/options.yml").map do |file|
```
nex3, line 136: At this point, I think it would be pretty easy to make this find *all* test cases by making this:

```suggestion
      dir.glob("**/input.{scss,sass,css}").map do |file|
```

PR #1542, opened 2020-06-04
Expand SassSpec::Directory specs

nex3:
This also makes Directory work with nested paths.


lib/sass_spec/directory.rb
```diff
@@ -88,21 +104,38 @@ def glob(pattern)
 
   # Returns whether a file exists at `path` within this directory.
   def file?(path)
-    return @archive[path].is_a?(HRX::File) if hrx?
-    File.exist?(File.join(@path, path))
+    if hrx?
+      @archive[path].is_a?(HRX::File)
+    elsif (dir, basename = split_if_nested(path))
```
Awjin, line 109: (Here and below) looks like these assignments might not be valid?

nex3: Updated the Ruby version to avoid this issue.

PR #1543, opened 2020-06-05
Refactor `@content arguments specs

jathak:


spec/directives/mixin/content/arguments/none.hrx
```diff
@@ -0,0 +1,105 @@
...
+a {
+  @include mixin {
+    x: y;
+  }
+}
+
+<===> no_parens/no_using/output.css
+a {
+  x: y;
+}
```
nex3, line 15: It seems strange to have a spec under `content/arguments` that calls a mixin without passing any arguments to its content block.

spec/directives/mixin/content/arguments/receiving.hrx
```diff
@@ -0,0 +1,148 @@
...
+@mixin mixin {
+  @content;
+}
+
+a {
+  @include mixin using ($arg1: value1, $arg2: value2) {
+    arg1: $arg1;
+    arg2: $arg2;
+  }
+}
```
nex3, line 11: This seems redundant with the `default_using` specs in `none.hrx`.

spec/directives/mixin/content/arguments/receiving.hrx
```diff
@@ -0,0 +1,148 @@
...
+a {
+  positional: (value1,);
+  keywords: (arg2: value2);
+}
+
+<===> with_splat/both_passed/output-libsass.css
+a {
+  positional: value1;
+  keywords: (arg2: value2);
+}
```
nex3, line 148: It's probably also a good idea to add a spec covering passing an ArgumentList object that contains both keywords and positional args to a `@content`.

spec/directives/mixin/content/arguments/scope.hrx
```diff
@@ -0,0 +1,74 @@
...
+    content-var: $var;
+  }
+}
+
+<===> no_argument/output.css
+a {
+  mixin-var-before: mixin-argument;
+  content-var: top-level;
+  mixin-var-after: mixin-argument;
+}
```
nex3, line 26: As above, since this isn't actually declaring or passing content arguments, it seems like it doesn't belong here.

connorskees:
Hello, 

I've come across an interesting case when compiling Bootstrap that may be relevant here.


The minimized version is
```scss
@mixin foo($arg) {
  @include bar {
    color: $arg;
  }
}

@mixin bar {
  @include baz {
    @content;
  }
}

@mixin baz {
  @content;
}

@mixin font-size($value) {
  @include foo($value);
}

a {
  @include font-size(1rem);
}
```
which I suppose can be thought of as a triple chain of `@content`.

PR #1559, opened 2020-09-14
Add specs for nested maps in map-get

jathak:
[skip dart-sass]

### DIFF
--- a/spec/core_functions/map/get.hrx
+++ b/spec/core_functions/map/get.hrx
@@ -78,6 +78,52 @@ a {
 
 <===>
 ================================================================================
+<===> nested/options.yml
+---
+:todo:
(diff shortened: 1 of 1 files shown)


spec/core_functions/map/get.hrx
```diff
@@ -76,6 +76,52 @@ a {
...
+
+<===> nested/found/full_path/output.css
+a {
+  b: f;
+}
+
+<===>
+================================================================================
+<===> nested/not_found/top_level/input.scss
+a {b: map-get((c: (d: (e: f))), d)}
```
nex3, line 107: Use `inspect()` here so it's clear that it returns `null`.

nex3 (approved):

Issue #1568, opened 2020-09-18
Add specs for nested map functions in sass:map


nex3:
See https://github.com/sass/sass/issues/1739 and https://github.com/sass/sass/pull/2836.

We're currently landing provisional support for this in the [feature.nested-maps](https://github.com/sass/sass-spec/commits/feature.nested-maps) branch, including the following PRs:

* https://github.com/sass/sass-spec/pull/1558
* https://github.com/sass/sass-spec/pull/1559
* https://github.com/sass/sass-spec/pull/1561
* https://github.com/sass/sass-spec/pull/1563
* https://github.com/sass/sass-spec/pull/1566

PR #1569, opened 2020-09-18
Add specs for map.deep-remove()

jathak:
See sass/sass#1739
See #1568

[skip dart-sass]

### DIFF
--- a/spec/core_functions/map/deep_remove.hrx
+++ b/spec/core_functions/map/deep_remove.hrx
@@ -1,0 +1,139 @@
+<===> options.yml
+---
+:todo:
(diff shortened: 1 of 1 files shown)


nex3 (changes requested):
You probably also want to test a case where the final value isn't actually a map.

nex3:
https://github.com/sass/dart-sass/pull/1091#discussion_r491226231 makes me realize that this should probably also test a case `remove()` dives deep into maps that don't exist, and verify that it doesn't actually create intermediate maps if they aren't necessary.

nex3 (approved):

Issue #1570, opened 2020-09-18
Add specs for HWB in sass:color


nex3:
See https://github.com/sass/sass/issues/2834.

PR #1576, opened 2020-10-10
Add specs for color.whiteness() and .blackness()

nex3:
See sass/sass#2834
See sass/sass-spec#1570
[skip dart-sass]

### DIFF
--- a/spec/core_functions/color/blackness.hrx
+++ b/spec/core_functions/color/blackness.hrx
@@ -1,0 +1,153 @@
+<===> options.yml
+---
+:todo:
+- sass/libsass#3128
+
+<===>
+================================================================================
+<===> max/input.scss
+@use 'sass:color';
+a {b: color.blackness(black)}
+
+<===> max/output.css
+a {
+  b: 100%;
+}
+
+<===>
+================================================================================
+<===> min/input.scss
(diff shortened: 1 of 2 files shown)


spec/core_functions/color/blackness.hrx
```diff
@@ -0,0 +1,153 @@
...
+
+<===>
+================================================================================
+<===> middle/zero_whiteness/input.scss
+@use 'sass:color';
+a {b: color.blackness(color.hwb(0, 0%, 50%))}
+
+<===> middle/zero_whiteness/output.css
+a {
+  b: 49.8039215686%;
```
jathak, line 36: Given the [leeway the spec gives](https://github.com/sass/sass/blob/master/proposal/color-4-hwb.md#blackness) on the values that `color.whiteness` and `color.blackness` return, should we be encoding the exact behavior of Dart Sass in these tests?

nex3: I thought about writing these entirely in terms of `hwb(hue($color), whiteness($color), blackness($color))` but I do want to make sure that we're at least aware if the implementation changes such that these numbers also change. Other implementations are still free to make separate (valid) output files, as with any allowed output difference.

PR #1579, opened 2020-10-28
Rewrite sass-spec runner in Typescript

tesseralis:
Rewrite the sass-spec test runner in Node.

## Running

1. `npm i` to install dependencies
2. `npm run sass-spec -- [options]` to run the test suite. All supported options the same as the ruby spec runner

## Running tests

1. `npm run test`

## Implementation Notes

* Tests for a single HRX archive are run in one block. This ensures the directories are not written to disk more than once and, in interactive mode, the HRX file isn't written more than once.
* The runner only writes input files to disk. Everything else, including changes from interactive mode, are done in memory.
* In Interactive Mode, the behavior for "update expected output" has changed. It now deletes any implementation-specific overrides.
  * In the Ruby runner, if the actual result is the same type as the expected result (e.g. a successful output), it will update the override. If the results are different (expected output but got error), it will write create a directory with *both* an output and error file, which is invalid.

## TODO

* [x] flesh out the remaining test cases
* [ ] Make libsass run faster
* [ ] Fix interrupts (Ctrl-C) on libsass
* [ ] Fix or delete the tests breaking the dart-sass runner


sass-spec.js
```diff
@@ -0,0 +1,192 @@
...
+  }
+  for (const filename of files) {
+    const filepath = path.resolve(dir, filename)
+    const filestat = await fs.stat(filepath)
+    if (filestat.isDirectory()) {
+      // If we run into a subdirectory, recurse into it
+      await iterateDir(filepath, opts, cb)
+    } else if (filename.endsWith(".hrx")) {
+      // If HRX, expand it into a directory and recurse into it
+      await unarchive(dir, filename)
```
nex3, line 81: At some point, you'll probably want to avoid unarchiving until you actually need to run tests in a directory, and otherwise just using the HRX API to load file contents. Otherwise you'll be spending a lot of cycles on filesystem operations that you may not end up using, especially when running only a subset of specs.

sass-spec.js
```diff
@@ -0,0 +1,192 @@
...
+
+/**
+ * Run through the given directory
+ * @param dir the directory to iterate over
+ * @param opts the options to run the thing with:
+ *    - impl -- the sass implementation to use
+ *    - todo -- whether todos should be run
+ * @param cb the callback to run on each directory
+ */
+async function iterateDir(dir, opts, cb) {
```
nex3, line 57: It's probably premature right now, but I recommend using the Ruby [`Directory` class](https://github.com/sass/sass-spec/blob/master/lib/sass_spec/directory.rb) as inspiration for an API for accessing files transparently between HRX and the physical filesystem. It's one of the newer and cleaner additions to the existing spec runner, and it's been very helpful as a basis for more complex features like the `--interactive` flag.

nex3 (approved):
This looks like a good start!

> This PR is against master, though I'm thinking for better development speed we could make a ts-sass-spec branch and make all the PRs off of it, and then merge to master when the whole project is done?

Sounds good to me.

sass-spec.js
tesseralis: I have looked at it! I just wanted to start out with something simpler because it seemed like a pretty feature-rich class and I wasn't sure which features we still wanted to include or not.

package.json
```diff
@@ -19,5 +19,16 @@
...
   },
-  "homepage": "https://github.com/sass/sass-spec#readme"
+  "homepage": "https://github.com/sass/sass-spec#readme",
+  "dependencies": {
+    "js-yaml": "^3.14.0",
+    "node-hrx": "^0.1.0"
+  },
+  "devDependencies": {
+    "prettier": "^2.1.2",
+    "tap": "^14.10.8"
```
stof, line 29: `tap` is not even a dev dependency here, as it is used by the implementation of the spec runner

lib-js/directory.ts
```diff
@@ -0,0 +1,102 @@
...
+import { withArchive } from "./hrx"
+
+interface YamlOptions {
+  ":todo"?: string[]
+  ":ignore_for"?: string[]
+  ":warning_todo"?: string[]
+  ":precision"?: number
+}
+
+function hasOptionForImpl(option: string[] | undefined, impl: string) {
```
nex3, line 14: It would help me for reviews to have these private functions/interfaces documented as well.

lib-js/directory.ts
```diff
@@ -0,0 +1,102 @@
...
+  bin: string
+  rootDir: string
+  testDir: string
+  // TODO guarantee that this is `run` or `probe`
+  todoMode?: string
+}
+
+type AllOptions = DirOptions & OverrideOptions
+
+type DirIteratee = (dir: string, options: AllOptions) => Promise<void>
```
nex3, line 53: Style nit: I'd just define this inline, I think that's clearer than making it a named type if you aren't going to reuse it.

lib-js/directory.ts
nex3, line 14: Don't forget return types on these functions!

nex3 (commented):
I think you're getting to the point where it would be really helpful to make the core logic more object-oriented. The Ruby test runner design of having a core `Directory` object that abstracts out the difference between reading and writing real/virtual files has proven very helpful, and I suspect it would also be useful to have a `TestCase` object that encapsulates a bunch of the logic related to resolving the configuration of a given test so it doesn't all have to be done procedurally.

nschonni:
FYI, had a pretty simple runner before for node-sass https://github.com/sass/node-sass/blob/601d31fbbd20f1d7d5521fd4e4d4bf6552e44d74/test/spec.js but I has since been removed and predated the HRX stuff

nex3:
I'm trying to run this locally and getting the following error:

```shellsession
$ git clean -xfd
$ npm i
$ npm run sass-spec -- --dart ../dart spec/core_functions/color/blackness

> sass-spec@3.5.4 sass-spec /home/nweiz/goog/sass/spec
> ts-node sass-spec.ts "--dart" "../dart" "spec/core_functions/color/blackness"

(node:543882) ExperimentalWarning: The fs.promises API is experimental
(node:543882) UnhandledPromiseRejectionWarning: Error: ENOTEMPTY: directory not empty, rmdir '/home/nweiz/goog/sass/spec/spec/core_fun
ctions/color/adjust_hue'                                                                                                             
(node:543882) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an asy
nc function without a catch block, or by rejecting a promise which was not handled with .catch(). (rejection id: 1)                  
(node:543882) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are no
t handled will terminate the Node.js process with a non-zero exit code.
```

nex3:
Looks like I was running with Node 11, which was causing some problems. Maybe the package.json should have an [`"engines"`](https://docs.npmjs.com/cli/v6/configuring-npm/package-json#engines) field?

tesseralis:
@nex3 right, sorry! I'll add it in and figure out what versions work on it.

lib-js/spec-directory/virtual-directory.ts
```diff
@@ -0,0 +1,203 @@
...
+      root,
+      parentOpts
+    )
+  }
+
+  // Get an HRX archive from a string
+  static async fromContents(contents: string, path = "/tmp") {
+    const stream = Readable.from(contents)
+    const archive = await archiveFromStream(stream)
+    // TODO where should the temp path be?
```
nex3, line 83: I'd require the caller to pass in a path explicitly. As far as sass-spec is concerned, a path is an integral component of a directory.

lib-js/spec-directory/virtual-directory.ts
```diff
@@ -0,0 +1,203 @@
...
+  async cleanup() {
+    // remove the physical directory
+    await fs.promises.rm(this.path, { recursive: true, force: true })
+    // if files were written to this directory, write to the root archive file
+    if (await this.hasModifications()) {
+      const hrx = await toHrx(this)
+      await fs.promises.writeFile(this.path + ".hrx", hrx, {
+        encoding: "utf-8",
+      })
+    }
```
nex3, line 187: This seems like another unnecessary coupling. I understand the instinct to buffer all the modifications, but since interactive mode isn't performance-sensitive, I think you could simplify this a lot by just rewriting the entire HRX file every time the user performs a modification.

nex3 (changes requested):
Got about halfway through. I'll try to get the rest reviewed tomorrow or early next week.

lib-js/interactor.ts
```diff
@@ -0,0 +1,198 @@
+import readline from "readline"
+import TestCase from "./test-case"
+
+// Select properties of the test case that can be used in requirements
+export type TestCaseArg = Pick<TestCase, "impl" | "actual" | "result">
```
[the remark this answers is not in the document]
tesseralis: Test typing — this way we can test the functions involved without having to create a whole test case and running it. I admit it's kind of a hack, I'll see if I can come up with a better idea. Probably an interface that the TestCase type implements, instead of doing it backwards like this?

lib-js/interactor.ts
```diff
@@ -0,0 +1,202 @@
...
+        })
+      })
+    }
+
+    const type = test.result().failureType || ""
+
+    // If a repeated choice is chosen for a given failure type, run that choice
+    if (this.memory[type]) {
+      const choice = this.memory[type]
+      // we're guaranteed that the stored chosen option returns a test result
```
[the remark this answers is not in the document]
tesseralis: oh whoops, it was a previous implementation that returned something.

lib-js/options.ts
```diff
@@ -0,0 +1,48 @@
+export interface RunOptions {
+  ":ignore_for"?: string[]
+  ":todo"?: string[]
+  ":warning_todo"?: string[]
+  ":precision"?: number
+}
+
+export type RunOption = ":ignore_for" | ":todo" | ":warning_todo"
```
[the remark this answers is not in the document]
tesseralis: I only intended to use this type on methods that have an implementation array. This way you can't, say, call `getOptionForImpl(':precision', 'dart-sass')`. I can make the name more precise, however.

lib-js/options.ts
```diff
@@ -0,0 +1,48 @@
...
+  function mergeOption(option: RunOption) {
+    return [...(base[option] ?? []), ...(ext[option] ?? [])]
+  }
+  return {
+    ":ignore_for": mergeOption(":ignore_for"),
+    ":todo": mergeOption(":todo"),
+    ":warning_todo": mergeOption(":warning_todo"),
+    ":precision": ext[":precision"] ?? base[":precision"],
+  }
+}
```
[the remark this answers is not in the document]
tesseralis: classes tend to not be used in JS modules unless absolutely necessary. I think most people would just do `import * as RunOptions from './run-options' if they wanted to make sure the methods were grouped together (which I can do for this and other files if it makes things clearer)

lib-js/spec-directory/hrx.ts
```diff
@@ -0,0 +1,112 @@
...
+
+async function getHrxSections(
+  dir: SpecDirectory,
+  root: string
+): Promise<string[]> {
+  if (dir.isTestDir()) {
+    return [await getTestDirHrx(dir, root)]
+  } else {
+    return getNormaliDirHrx(dir, root)
+  }
```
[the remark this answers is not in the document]
tesseralis: The distinction I think is that for test files, subdirs should be listed in the same section as another input, whereas for non-test directories, they should be included as separate sections?

test/interactor/options.test.ts
```diff
@@ -0,0 +1,51 @@
...
+      await expectOption(key, {})
+    }
+  })
+
+  it("does show the 'show output' option when the failure type is `warning_difference", async () => {
+    const arg = { result: { failureType: "warning_difference" } }
+    await expectOption("o", arg, false)
+  })
+
+  it("Shows the 'show error' option only when errors and warnings are available", async () => {
```
nex3, line 33: Consistency nit
```suggestion
  it("shows the 'show error' option only when errors and warnings are available", async () => {
```

test/spec-directory/cleanup.test.ts
```diff
@@ -0,0 +1,30 @@
...
+    } catch (e) {}
+  })
+
+  it.skip("does the cleanup on a SIGINT", async () => {
+    // const mockExit = jest.spyOn(process, "exit").mockImplementation()
+    const mock = jest.fn()
+    await withAsyncCleanup(mock, async () => {
+      process.kill(process.pid, "SIGINT")
+    })
+    expect(mock).toHaveBeenCalled()
```
nex3, line 28: How is this not just killing the process?

test/spec-directory/mutations.test.ts
```diff
@@ -0,0 +1,25 @@
+import path from "path"
+import { fromPath } from "../../lib-js/spec-directory"
+
+// Tests for methods on SpecDirectory that mutate its contents
+describe("SpecDirectory mutations", () => {
+  describe("writeFile", () => {
```
nex3, line 6: I'd also test writing a file when none exists yet.

test/spec-directory/mutations.test.ts
```diff
@@ -0,0 +1,25 @@
...
+      await dir.writeFile("output.css", "NEW OUTPUT")
+      expect(await dir.readFile("output.css")).toEqual("NEW OUTPUT")
+      await dir.writeFile("output-libsass.css", "MORE OUTPUT")
+      expect(await dir.readFile("output-libsass.css")).toEqual("MORE OUTPUT")
+      expect(await dir.listFiles()).toContain("output-libsass.css")
+    })
+  })
+
+  describe("deleteFile", () => {
+    it("removes the contents of the virtual file", async () => {
```
nex3, line 18: Also test error-handling for a non-existent file, a directory, and an HRX file within a real directory.

test/spec-directory/iteration.test.ts
```diff
@@ -0,0 +1,60 @@
+import path from "path"
+import { fromPath, SpecDirectory } from "../../lib-js/spec-directory"
+
+describe("SpecDirectory iteration", () => {
```
nex3, line 4: It looks like these tests only test one of real directory/virtual directory, even though the underlying implementation differs between them. It's probably a good idea to test both.

test/test-case/expected.test.ts
```diff
@@ -0,0 +1,120 @@
...
+      const result = await testCase.expected()
+      expect(result).toEqual(expected)
+    }
+
+    it("returns contents of `output.css` when there are no overrides", async () => {
+      await expectOutput(`
+<===> output.css
+OUTPUT`)
+    })
+    it("returns `output-[impl].css` contents when there is an impl-specific override", async () => {
```
nex3, line 25: Style nit: can you add empty lines between each of these test cases?

nex3 (changes requested):

lib-js/test-case/test-case.ts
```diff
@@ -0,0 +1,249 @@
...
+ *
+ * Contains methods for running the test and updating the underlying directory
+ * based on the results.
+ */
+export default class TestCase {
+  dir: SpecDirectory
+  impl: string
+  compiler: Compiler
+  todoMode?: string
+  private _actual?: SassResult
```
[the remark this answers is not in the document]
tesseralis: To distinguish it from the public `actual()` method below, which ensures that the test has run before attempting to get the test information.

test/spec-directory/cleanup.test.ts
tesseralis: …it does, which is why this test is skipped for now. 😓

Probably the best way to test this would be to call the `withAsyncCleanup` function from a script, then exec that script from the test and check that cleanup occurs. But that makes the tests run a lot slower (because of the overhead of ts-node compilation).

lib-js/spec-directory/spec-directory.ts
```diff
@@ -0,0 +1,143 @@
...
+  /** Get the list of direct files filenames of this directory */
+  abstract listFiles(): Promise<string[]>
+  /** Returns true if this directory has the given filename as a subitem */
+  abstract hasFile(filename: string): boolean
+  /** The file contents of the given filename */
+  abstract readFile(filename: string): Promise<string>
+  /** Update the contents of the given file */
+  abstract writeFile(filename: string, contents: string): Promise<void>
+  /** Remove the file from this directory */
+  abstract removeFile(filename: string): Promise<void>
```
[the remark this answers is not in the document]
nex3: That's fine, just make sure it's documented and that there are runtime argument checks to verify that only a single-level path is being used.

lib-js/test-case/test-case.ts
```diff
@@ -0,0 +1,249 @@
...
+      return { isSuccess: false, error: stderr }
+    }
+  }
+
+  async run(): Promise<TestResult> {
+    if (this._result) {
+      throw new Error(`Test case ${this.dir.relPath()} has already been run`)
+    }
+    // Remember to cache the results of the run, regardless of result
+    this._result = await this.doRun()
```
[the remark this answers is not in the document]
nex3: Having the test case track both the intrinsic properties of the test itself as derived from the filesystem *and* the contingent properties that depend on how the user is interacting with it seems like a separation of concerns violation. I'd rather have the test runner keep a map from test cases to results.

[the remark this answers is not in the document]
tesseralis: I think the problem with keeping the test cases and test results separate but linked with a mapping is that, in interactive mode, the user can *change* the test case through the various overwriting options. You have to pass in the test case anyways in order to overwrite it.

I don't think it's a separation of concerns violation because the test case and its results are linked: if you mark a test result as passing, you have to update the test contents so that it passes.

That said, I do think having to call `run()` is cumbersome and have some ideas on how to fix that (e.g. making sure the test results are calculated when constructing the TestCase object)

lib-js/spec-directory/virtual-directory.ts
tesseralis: I'm only using this method for testing, on methods where the actual path doesn't matter as much. Should I just make that clear in the documentation?

lib-js/spec-directory/virtual-directory.ts
```diff
@@ -0,0 +1,203 @@
...
+function createSubdirCache(dir: HrxDirectory) {
+  const cache: Record<string, HrxDirectory> = {}
+  for (const itemName of dir) {
+    const subitem = dir.get(itemName)
+    if (subitem?.isDirectory()) {
+      cache[itemName] = subitem
+    }
+  }
+  return cache
+}
```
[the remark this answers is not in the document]
tesseralis: The file cache is needed because the node-hrx `Directory` doesn't allow updating file contents. I opted to not store the original `Directory` object because, once the directory object is created, it can be modified and the original contents of the `HrxDirectory` will be outdated. It's a way to safeguard against future contributors, say, trying to implement another file listing method by calling `HrxDirectory.list()`, since that will be outdated and incorrect if there have been changes to the filesystem.

lib-js/spec-directory/spec-directory.ts
```diff
@@ -0,0 +1,143 @@
...
+    this.parentOpts = parentOpts
+    this.subitems = {}
+  }
+
+  /** The path of this directory relative to the top level that was created */
+  relPath() {
+    // make sure to include the root dir as part of the name
+    // (e.g. if the root path is `spec`, everything should be listed as `spec/thing`)
+    const rootDir = path.dirname(this.root.path)
+    return path.relative(rootDir, this.path)
```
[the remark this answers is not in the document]
tesseralis: I think so? There's an implicit implication that all the specs start from the `spec` directory (it's passed in as the `load-path` option, it's the top place to check for options.yml). I'm writing an integration test to test the script itself and it's useful to be able to pass in a different directory. `relPath` is used as a key, so I think it makes sense for it to return the same thing for two instances of the test run regardless of where sass-spec is run from.

lib-js/compiler.ts
```diff
@@ -0,0 +1,177 @@
...
+    const child = child_process.spawn("dart", [
+      "--enable-asserts",
+      `--packages=${repoPath}/.packages`,
+      dartFilename,
+    ])
+    // TODO figure out how to delete the file without crashing the process
+    // fs.unlinkSync(dartFilename)
+    // child.on("spawn", () => {
+    //   fs.unlinkSync(dartFilename)
+    // })
```
tesseralis, line 129: @nex3 your other commend got marked outdated, but I tried tempy and it runs into the same problem—node exits with an `EPIPE` if the file is deleted while the spawned process is still existent.

I can think of a couple ways we can resolve this:

1. still use tempy to generate a file name but just let it live in the tmp directory
2. make an explicit `cleanup` method for the compiler to call at the end of the test run.
3. Do more research and figure out how to delete the file.

lib-js/spec-directory/spec-directory.ts
```diff
@@ -0,0 +1,143 @@
...
+  }
+
+  // TODO should we move this function out to its own file?
+  /**
+   * Iterate through the subpaths of this directory, running the iteratee
+   * on all
+   * @param paths the paths to match against
+   * @param iteratee the function to call for each matching subdirectory
+   */
+  async forEachTest(paths: string[], iteratee: SpecIteratee) {
```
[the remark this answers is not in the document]
tesseralis: I decided to have two methods together because I wanted the file setup/cleanup to occur when each virtual directory is being iterated over. I know the ruby version does this the other way (compiles a list of test directories, then instantiates the files for each test directory) but I think this implementation is cleaner and requires less functionality to be implemented?

To replicate the old behavior we'd need a reference to the parent directory of virtual directories so that the files can be instantiated, and we need to read the input file to see if it imports anything so that we don't end up creating the whole archive for each file.

With the current implementation, each directory only needs to know about itself and its children, and only the archive root is responsible for knowing when to create and update files. IMO this is easier to think about (and implement and test). What do you think?

tesseralis:
@nex I'm doing some cleanup work and have a couple of things to discuss with you:

## libsass/executable speed

Right now, the ruby version takes ~55s to run on my machine for running libsass as an executable, while the new node/ts version takes ~70s. The --dart command runs at comparable speeds (~40s) in both versions. I've tried all the different node `child_process` functions, pipe options, etc. but nothing seems to be able to take the time down.

Considering libsass is deprecated, and this might boil down to "node is slower than ruby at spawning processes", is the additional +15s an acceptable increase?

## Behavior on invalid specs

I've been looking at the behavior on different types of invalid specs and am getting inconsistent behavior from the ruby runner:

* when multiple input files are defined (input.sass and input.scss), an error is thrown and the process exits
* when no output or error file is found, the test is registered as a failure (F)
  * the error message is "Expected output.css file does not exist"
* when *both* an output and error file are found, or a warning and an error file are present, the test is registered as an error (E)
  * the error message is nonspecific: "NoMethodError: undefined method `expected_path' for #<SassSpec::TestCase:0x00007f851b112360>"

How would you like these cases handled?

lib-js/spec-directory/virtual-directory.ts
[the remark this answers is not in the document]
nex3: Gotcha. Ideally we'd update `node-hrx` to support mutating files, but that may not be worth it in the short term. Can you document this in the code?

lib-js/spec-directory/virtual-directory.ts
nex3: Yeah, that's fine.

lib-js/compiler.ts
nex3: I wonder why this worked in Ruby but not in Node...

Could you add an `exit` listener that only unlinks the file once the process is done?

nex3:
> libsass/executable speed

I think a 25% speed regression for libsass is fine.

> Behavior on invalid specs

I think marking them as either failures or errors is fine, whichever's easiest to implement.

tesseralis:
> I think marking them as either failures or errors is fine, whichever's easiest to implement.

But you'd like invalid specs to not cause the process to crash, right?

lib-js/compiler.ts
tesseralis: That works! Thanks for the idea

lib-js/test-case/expected.ts
```diff
@@ -0,0 +1,66 @@
...
+  if (dir.hasFile(`output-${impl}.css`) && dir.hasFile(`error-${impl}`)) {
+    throw new Error(`Found both output and error file for ${impl}.`)
+  }
+
+  if (dir.hasFile(`output.css`) && dir.hasFile(`error`)) {
+    throw new Error(`Found both \`output.css\` and \`error\` file.`)
+  }
+}
+
+function expectsSuccess(dir: SpecDirectory, impl: string) {
```
nex3, line 15: This name is kind of confusing—it sounds like it's verifying that `dir` expects a successful outcome, but it will pass for directories that expect errors as well. Maybe `checkOutputExpectation`?

lib-js/test-case/util.ts
```diff
@@ -25,11 +25,17 @@ type FailureType =
...
 export interface TestResult {
-  type: "pass" | "fail" | "todo" | "skip"
+  type: "pass" | "fail" | "todo" | "skip" | "error"
+  // If `fail`, the type of failure, message, and possible diff
   failureType?: FailureType
   message?: string
   diff?: string
+  // If `error`, the thrown error
+  error?: Error
 }
```
nex3: Consider making it explicit which fields go with which type:

```ts
export type TestResult =
  | ErrorResult
  | FailResult
  | { type: "pass" | "todo" | "skip" }

export interface ErrorResult {
  type: "error"
  error: Error
}

export interface FailResult {
  type: "fail"
  failureType: FailureType
  message: string
  diff: string
}
```

nex3 (commented):
Looking good! Can you go through and add return types to all the functions that don't have them yet?

lib-js/test-case/expected.ts
tesseralis: It does — it returns `true` if it expects success, and `false` if it expects an error. It's not really a `check` because it's supposed to return a boolean, not an error. I've added a comment, but do you have a better name idea?

lib-js/test-case/util.ts
tesseralis: I had it this way before, but it ended up being somewhat cumbersome (e.g. having to do an additional check in some cases even if you *know* a type is a `FailResult`). I can try and split it up again, however.

tesseralis:
> Looking good! Can you go through and add return types to all the functions that don't have them yet?

I've added return types to all public methods and exported members. I've left test functions and private members off since those can tend to shift more often. I've also not put return types on overloaded methods, since those will error out anyways if the return type doesn't match the signature of the overloaded method.

Does that work?

lib-js/test-case/util.ts
tesseralis: okay so I recreated the issue. The problem is, in `interactor.ts`, we have a couple of functions that take `TestCase` as an argument. Methods on Interactor should only work on `FailResult`. So if we pass in the updated description, we get a bunch of errors like this:

```
lib-js/interactor.ts:74:42 - error TS2339: Property 'failureType' does not exist on type 'TestResult'.
  Property 'failureType' does not exist on type '{ type: "pass" | "todo" | "skip"; }'.
```

The problem is that all the methods in the Interactor take a `testCase` as an argument:

```
  {
    key: "G",
    description: (test: TestCase) => `Ignore test for ${test.impl} FOREVER`,
    requirement: (test: TestCase) => test.result().failureType !== "unnecessary_todo",
    resolve: (test: TestCase) => test.markIgnore(),
  },
```

There's no easy way to tell typescript that a particular `testCase` instance has a `FailResult`. So if you wanted it to work you'd have to do one of the following:

* Add checks in every method that `test.result()` is a `FailResult`, purely for type-checking (since we do a similar check in the functions that call these)
* Do an unsafe assertion that `test.result()` is a `FailResult`, (e.g. `(test.result() as FailResult).failureType`)
* Make `TestCase` take `TestResult` type as a generic parameter (which opens up another can of complexity)
* Pass the typed `FailResult` as an argument to all these submethods, which is extraneous and makes testing hard
* reorganize the structure of the interactor somehow

I do agree that splitting it up would be more type-safe, but given the complexity of the workarounds, would it be alright to just put in a TODO about the reasoning for now?

nex3: Making it a TODO for now sounds fine.

nex3:
I'd really prefer return types everywhere if at all possible. If there's an overload that's causing type problems, feel free to leave that as a TODO, but I find return types particularly helpful for getting my bearings in a new codebase and this is going to be new to me for a while :smile:.

tesseralis:
@nex3 alright! Added all the types and the TODO for the overloads. Please let me know if there's anything else that needs to be done before this is merged!

nex3 (approved):
Looks great! Thanks for all your hard work, Nat!

types/node-hrx/index.d.ts
```diff
@@ -0,0 +1,26 @@
+declare module "node-hrx" {
+  import { Readable } from "stream"
+
+  class File {
+    path: string
+    body: string
+    comment?: string
+    isDirectory(): this is Directory
```
stof, line 8: isn't this simpler to write it as `false` here and `true` below (and the opposite in `Directory`), as an item can never be both a File and a Directory

types/node-hrx/index.d.ts
```diff
@@ -0,0 +1,26 @@
+declare module "node-hrx" {
```
stof, line 1: Are these type definitions for node-hrx complete or do they cover only the parts used by the runner ? If they are complete, I suggest upstreaming them to DefinitelyTyped.

nex3: Honestly it may be more sustainable to just convert node-hrx to TypeScript directly, since afaik this is now one of the main projects using it.

mullib (commented):
> #1666

PR #1580, opened 2020-10-28
Add specs covering double width chars for str-index

stof:
Other `str-*` functions dealing with string indexes have such tests but they were missing for `str-index`. This ensures that `str-index` handles them properly (which is required to be useful when other functions handle them properly).

### DIFF
--- a/spec/core_functions/string/index.hrx
+++ b/spec/core_functions/string/index.hrx
@@ -58,6 +58,31 @@ a {
 
 <===>
 ================================================================================
+<===> double_width_character/input.scss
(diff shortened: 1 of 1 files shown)


nex3:
It looks like these specs are broken somehow?

stof:
`combining_character` indeed does not work. I took the input from existing `combining_character` tests in other functions and I added a `a` after that that I would search, but I messed up the input as `a` would be taken as part of the escape sequence too. I was missing the space to end it before the `a`. this is now fixed.

nex3 (approved):

PR #1585, opened 2020-10-29
Refactor specs about modulo for numbers

stof:
I was looking at specs for the behavior of the modulo operator. I finally discovered that they are covered by some `libsass-closed-issues` specs only. So I decided to contribute here by refactoring them to follow the guidelines instead.

- `issue_2236/*` is what I kept to create the new specs, but refactored to follow guidelines
- `issue_103.hrx` was added to cover an issue which was actually "libsass does not implement the modulo operator at all", which is obviously covered by the new specs
- `issue_1029.hrx` covers an issue for which the root cause was that `$negativeValue % 2` returned a negative value rather than a positive one (leading to extending `%shift--1` instead of `%shift-1` in the code). This is also covered the specs as they cover all sign combinations to enforce them (which was not the case yet at the time this was added)


spec/values/numbers/modulo/floats.hrx
```diff
@@ -0,0 +1,45 @@
+<===> positive_positive/input.scss
+a {
+  b: +1.2 % +4.7;
```
nex3, line 3:
```suggestion
  b: 1.2 % 4.7;
```

I think including the explicit `+` here is a bit confusing—the numbers are still unambiguously positive without it.

nex3 (changes requested):
Thanks for doing these refactors! They're super helpful for cleaning up the repo :smile:.

I think it's probably worth adding a few additional test cases. There currently aren't any tests for when the left-hand side is larger than the right-hand side, which is very relevant for the modulo operation.

spec/values/numbers/modulo/floats.hrx
stof: I agree about that. The explicit `+` was already there in the old spec and I copied it verbatim. I will update that

stof:
@nex3 updated with more test cases

nex3 (approved):

Issue #1551, opened 2020-08-17
Cloning the repo on a case insensitive filesystem (macOS and probably also Windows) fails


stof:
This has been reported in https://github.com/scssphp/scssphp/pull/127#issuecomment-674942793
Here is the git output when running on macOS:

```
$ git clone https://github.com/sass/sass-spec.git
Cloning into 'sass-spec'...
remote: Enumerating objects: 65, done.
remote: Counting objects: 100% (65/65), done.
remote: Compressing objects: 100% (53/53), done.
remote: Total 58049 (delta 30), reused 27 (delta 10), pack-reused 57984
Receiving objects: 100% (58049/58049), 7.39 MiB | 3.55 MiB/s, done.
Resolving deltas: 100% (21180/21180), done.
warning: the following paths have collided (e.g. case-sensitive paths
on a case-insensitive filesystem) and only one from the same
colliding group is in the working tree:

  'spec/libsass/Sáss-UŢF8.hrx'
  'spec/libsass/Sáss-UŢF8.hrx'
```

The 2 files even seem to contain the same content.
I suggest removing one of them.

stof:
For reference, the files are those:
- https://github.com/sass/sass-spec/blob/73222792c42a516d62e2e25c3f5d9e35f5567030/spec/libsass/Sa%CC%81ss-UT%CC%A7F8.hrx
- https://github.com/sass/sass-spec/blob/73222792c42a516d62e2e25c3f5d9e35f5567030/spec/libsass/S%C3%A1ss-U%C5%A2F8.hrx

stof:
@nex3 should one of these 2 spec files be removed ?

nex3:
Although it's hard to tell since they aren't documented, I'm guessing the point of these files is to verify the behavior of Sass in the presence of files with glyph-identical names. That said, since it causes infrastructural headaches and there's no reason in principle why that behavior should fail, I'm okay with deleting them. Do you want to send a PR?

PR #1586, opened 2020-10-30
Remove duplicate file creating issues in a case sensitive filesystem

stof:
Closes https://github.com/sass/sass-spec/issues/1551

I kept the other file as it is the one used in the CI for libsass: https://github.com/sass/libsass/blob/87292ae4b2167401b505be1188d3b82861ab3253/appveyor.yml#L93
And having only one of them won't be an issue.


nex3 (approved):

nex3:
Thanks @stof!

PR #1589, opened 2020-11-02
Add a spec enforcing that import don't leak a load path

stof:
I discovered that the handling of imports in https://github.com/scssphp/scssphp/ is flawed. Importing a file leaks its directory as a load (and before configured load paths), effectively allowing to resolve imports relatively to caller files in the import chain.
This adds a spec preventing such behavior.
Neither dart-sass nor libsass are affected by the issue.

### DIFF
--- a/spec/directives/import/error/not_found.hrx
+++ b/spec/directives/import/error/not_found.hrx
@@ -44,3 +44,34 @@ Error: Can't find stylesheet to import.
   |         ^^^^^^^
   '
   input.scss 3:9  root stylesheet
+
+<===>
+================================================================================
+<===> parent_relative/input.scss
+// A file in a subdirectory shouldn't be able to load a URL relative
+// to the importing file.
+// Regression test for scssphp/scssphp#242
+@import "dir/child"
+
+<===> parent_relative/sibling.scss
+a {b: ""}
+
+<===> parent_relative/dir/child.scss
+@import "sibling"
+
+<===> parent_relative/error
+Error: Can't find stylesheet to import.
+  ,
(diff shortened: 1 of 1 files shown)


spec/directives/import/error/import_relative.hrx
```diff
@@ -0,0 +1,27 @@
+<===> input.scss
+// Importing a file must not allow that file to resolve its URL
+// relatively to the importer but only to itself (and to load paths)
```
nex3, line 3: I find this comment a little hard to follow. Maybe rewrite as:

```suggestion
// A file in a subdirectory shouldn't be able to load a URL relative
// to the importing file.
```

spec/directives/import/error/import_relative.hrx
```diff
@@ -0,0 +1,27 @@
...
+@import "child/child"
+
+<===> sibling.scss
+a {b: ""}
+
+<===> child/child.scss
```
nex3, line 9: Naming both path components "child" is a bit confusing. Maybe, `dir/child.scss`?

spec/directives/import/error/import_relative.hrx
nex3, line 3: Also, can you add a note that this is a regression test and a reference to the issue in scssphp that has details about the bug?

nex3 (changes requested):

stof:
@nex3 all comments have been taken into account.

nex3 (approved):

nex3:
Thanks!

PR #1591, opened 2020-11-03
Make the spec runner compatible with the sass executable

### LINKED ISSUE #1540: Allow running specs against a Dart Sass binary
Running specs with `--dart` runs Dart Sass from source through a Dart script that runs all specs in a single VM, avoiding the VM startup time for subsequent specs.

However, with `dart2native`, it's now possible to create a Dart Sass executable with minimal startup time. When running small sets of specs repeatedly while refactoring, the single VM startup time exceeds the cumulative startup time of running the native executable for each test.

It would be nice to be able to do `--impl dart-sass -c /path/to/native/dart/sass`, similar to how running libsass tests work, but Dart Sass's CLI is currently not compatible with what sass-spec expects.

Once support for running Dart Sass in this way is implemented, it would also be useful to run benchmarks to see if running a native executable repeatedly is faster than running all specs from source in a single VM.

stof:
Closes #1540 

To run specs against a sass binary, use this command:

```
bundle exec ./sass-spec.rb --impl dart-sass -c /path/to/sass --cmd-args='--no-unicode --no-color'
```

`--cmd-args` is needed to match the arguments used by `--dart` so that the output matches the expectation.

Regarding performance, this indeed helps when running only a few specs. Here are the result on my machine. The `-c` run uses the Linux x64 binary for dart-sass 1.28.0 downloaded from the github releases.

Full run:
- `--dart`: Finished in 23.162925s, 251.1773 runs/s
- `-c`: Finished in 131.673413s, 44.1851 runs/s

Run filtered on `spec/directives/import/error/not_found`:
- `--dart`: Finished in 0.264161s, 11.3567 runs/s
- `-c`: Finished in 0.069728s, 43.0243 runs/s

### DIFF
--- a/lib/sass_spec/engine_adapter.rb
+++ b/lib/sass_spec/engine_adapter.rb
@@ -28,7 +28,7 @@ class ExecutableEngineAdapter < EngineAdapter
     @command = command
     @timeout = 90
     @description = description || command
-    @version = `#{@command} -v`
+    @version = `#{@command} --version`
   end
 
   def set_description(description)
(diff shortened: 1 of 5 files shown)


jathak (approved):
Thanks for this!

Issue #1581, opened 2020-10-28
Convert the test runner to TypeScript


nex3:
This is a high-level bug to track the work of converting the test runner from Ruby to TypeScript. We want to do this conversion to increase the scope of developer familiarity with the runner—Ruby is no longer as widely-known a language as it once was—and as an opportunity to do away with seldom-used test runner features and add thorough tests for the features that remain.

I consider the following features important enough to support in the new test runner. (If anyone else wants to add to or remove from this list, let's discuss in the comments.)

* [ ] Running all existing specs (including support for all the spec-authoring features described in [Spec Structure](https://github.com/sass/sass-spec/#spec-structure) in the readme).
* [ ] The `--dart` flag, which takes the path to a Dart Sass checkout and spins up a Dart Sass subprocess to efficiently run specs against Dart Sass.
* [ ] The `--impl`, `--command`, and `--cmd-args` flags, which make it possible to run specs against other implementations' command-line interfaces.
* [ ] The `--run-todo` flag, which runs specs even if they have `:todo` annotations.
* [ ] The `--probe-todo` flag, which runs specs even if they have `:todo` annotations but only reports them as failures if they pass (indicating that the `:todo` is no longer necessary).
* The `--interactive` flag, with the following interactions:
  * [ ] Displaying the contents of the test case (as HRX)
  * [ ] Displaying the actual result (output, warning, or error)
  * [ ] Showing the diff between actual and expected result
  * [ ] Replacing the expected result with the actual result
  * [ ] Adding the expected result as an implementation-specific result
  * [ ] Marking the test as `:todo` for the current implementation
  * [ ] Marking the test as `:ignore_for` the current implementation
  * [ ] The "!" operator which repeats the same interaction any time the same context comes up in a given run. (This is lower-priority than the other features.)

The test runner does *not* need to immediately support the `--generate`, `--filter`, `--limit`, `--migrate-impl`, `--silent`, or `--check-annotations` flags, nor does it need to support any of the features of `lib/sass_spec/annotate/cli.rb`.

PR #1594, opened 2020-11-05
Missing separator in hrx file

Cerdic:
Leading to a test merge of `whiteness/low` with `blackness/max` when running the test suite in https://github.com/scssphp/scssphp/


Cerdic:
There are some other cases of missing separator:
https://github.com/sass/sass-spec/blob/master/spec/css/escape.hrx#L16
https://github.com/sass/sass-spec/blob/master/spec/core_functions/color/alpha.hrx#L48
https://github.com/sass/sass-spec/blob/master/spec/core_functions/color/alpha.hrx#L96
https://github.com/sass/sass-spec/blob/master/spec/css/min_max/error.hrx#L23
https://github.com/sass/sass-spec/blob/master/spec/directives/import/load.hrx#L81
https://github.com/sass/sass-spec/blob/master/spec/core_functions/color/change_color/hwb.hrx#L43
https://github.com/sass/sass-spec/blob/master/spec/directives/forward/member/visibility.hrx#L40


And this two are not clear for me, looks like two tests are merged
https://github.com/sass/sass-spec/blob/master/spec/directives/forward/member/visibility.hrx#L46
https://github.com/sass/sass-spec/blob/master/spec/directives/forward/member/visibility.hrx#L134

Or should I consider this as normal?

nex3:
All of those examples should have separators. In the last two, the separators should go after the first `output.css` file.

At some point, I'd like to have a linter that verifies these conventions automatically, but that'll probably need to wait until after #1581 is finished.

Cerdic:
So I completed the PR with the other missing separators now detected by our test suite.

I made a separate commit 3669c64182db8b4c662c6b86b18cb92778eff8d3 for spec/directives/forward/member/visibility.hrx file as it appears to me that separators are just misplaced in this file (and none missing at the end) and a review would be useful

stof:
Note that our scssphp testsuite currently relies on the separator comment to separate tests, because there is no hrx reader for PHP (and we haven't made one) so we "hacked" a way to identify specs based on the convention.

nex3 (approved):

PR #1595, opened 2020-11-11
Disable null safety when running Dart tests

jathak:
The latest dev version of Dart enables null safety by default, so we need to explicitly disable it in the spec running script until we migrate to null safety (which we have to wait for all of our dependencies to migrate first).


jathak:
Travis doesn't run on the Dart dev channel for this repo, but I've confirmed this fixes the spec running with the latest dev SDK locally.

nex3 (approved):

PR #1600, opened 2020-11-18
Tests on escaping char in selectors: number, non alpha chars and non alpha chars in interpolation

Cerdic:
Related to an issue on ScssPHP https://github.com/scssphp/scssphp/pull/274 we are proposing to extend the spec on char escaping in selectors. 

As a matter of fact, the only test on that point for now was in https://github.com/sass/sass-spec/blob/master/spec/non_conformant/extend-tests/escaped_selector.hrx (and ScssPHP was not compliant on it, but due to Selector Parsing, not due to `@extend` compilation)

The following tests are covering 
- escaped number as a first char and as a non first char
- escape sequence following or not by a space
- escaped non alpha char in the selector
- escape sequence in an interpolation in the selector
- and a combo, which is a grid selector from a framework, combining all these together

I add a `todo: sass/libsass` on each test as it seems to be what should be done, and I did not test these specs on libsaas.

The result on dart-sass is the one on the proposal, according to test done with SassMeister.

### DIFF
--- a/spec/css/selector.hrx
+++ b/spec/css/selector.hrx
@@ -362,11 +362,6 @@ Error: unterminated attribute selector for a
 
 <===>
 ================================================================================
-<===> escaping/number_as_first_char_without_space/options.yml
----
-:todo:
-- sass/libsass
-
 <===> escaping/number_as_first_char_without_space/input.scss
 .\31u {clear: left;}
 
@@ -375,13 +370,13 @@ Error: unterminated attribute selector for a
   clear: left;
 }
 
+<===> escaping/number_as_first_char_without_space/output-libsass.css
+.\31u {
+  clear: left;
+}
+
 <===>
 ================================================================================
-<===> escaping/number_as_first_char_with_space/options.yml
----
-:todo:
-- sass/libsass
-
 <===> escaping/number_as_first_char_with_space/input.scss
 .\31 u {clear: left;}
 
@@ -392,11 +387,6 @@ Error: unterminated attribute selector for a
 
 <===>
 ================================================================================
-<===> escaping/number_as_nonfirst_char_without_space/options.yml
----
-:todo:
-- sass/libsass
-
 <===> escaping/number_as_nonfirst_char_without_space/input.scss
 .a\31u {clear: left;}
 
@@ -405,13 +395,13 @@ Error: unterminated attribute selector for a
   clear: left;
 }
 
+<===> escaping/number_as_nonfirst_char_without_space/output-libsass.css
+.a\31u {
+  clear: left;
+}
+
 <===>
 ================================================================================
-<===> escaping/number_as_nonfirst_char_with_space/options.yml
----
-:todo:
-- sass/libsass
-
 <===> escaping/number_as_nonfirst_char_with_space/input.scss
 .a\31 u {clear: left;}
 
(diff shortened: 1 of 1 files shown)


Cerdic:
Looks like the `todo: sass/libsass` mentions are not correctly formated as test on libsass is indeed failing :( 
https://travis-ci.com/github/sass/sass-spec/jobs/442192145

I can correct if I know what I did wrong!

stof:
Given that the libsass output is still valid CSS (with more escaping applied than in dart-sass), maybe you should add a `output-libsass` file.

Regarding the TODO, it should either be `libsass` or something starting with `sass/libsass#` for things to work

Cerdic:
indeed, I added the output-libsass where needed and removed the todo!

spec/css/selector.hrx
```diff
@@ -359,3 +359,119 @@ Error: unterminated attribute selector for a
 >> [a=b ï] {c: d}
 
    ---^
+
+<===>
+================================================================================
+<===> escaping/number_as_first_char_without_space/input.scss
+.\31u {clear: left;}
```
nex3, line 366: Please [use single-letter names for irrelevant placeholders](https://github.com/sass/sass-spec/blob/master/STYLE_GUIDE.md#do-use-single-letter-names-for-irrelevant-placeholders).

spec/css/selector.hrx
```diff
@@ -359,3 +359,119 @@ Error: unterminated attribute selector for a
...
+}
+
+<===> escaping/dollar_char_as_numeric/output-libsass.css
+.u\24 {
+  clear: left;
+}
+
+<===>
+================================================================================
+<===> escaping/parenthesis_in_interpolation/input.scss
```
nex3, line 445: It's unclear to me what [single behavior](https://github.com/sass/sass-spec/blob/master/STYLE_GUIDE.md#do-test-only-one-thing-per-spec) this spec is testing. Is it non-identifier characters in interpolation? If so, why it the interpolation so complex to generate? Why not just write `.u#{"$"}`?

spec/css/selector.hrx
```diff
@@ -359,3 +359,119 @@ Error: unterminated attribute selector for a
...
+
+<===> escaping/parenthesis_in_interpolation/output-libsass.css
+.u\28xlarge\29 {
+  clear: left;
+}
+
+
+<===>
+================================================================================
+<===> escaping/alltogether_in_grid_selector/input.scss
```
nex3, line 463: This test case also doesn't seem to test a single behavior.

nex3 (changes requested):
The libsass output here is not valid. [The specification](https://github.com/sass/sass/blob/master/spec/syntax.md#consuming-an-escaped-code-point) is clear about exactly how escaped code points should be represented in identifiers, and LibSass doesn't follow that behavior. https://github.com/sass/libsass/issues/2700 was the tracking issue for that change, but it's closed now that LibSass is deprecated so you can feel free to just mark specs like these as ignored for LibSass.

If we're adding tests for escapes outside of a SassScript context, I'd like to cover all the edge cases in the specification above. Non-printable characters are notably missing, currently.

stof:
@Cerdic will you finish this or do you want me to take over the work instead ?

Cerdic:
@stof I will try to come back on this today

stof:
@nex3 as far as CSS selectors are concerned, the libsass output is not wrong. The are semantically the same CSS selectors (the libsass parsing which does not convert escape sequences to the actual char is an issue if this appears in equality selectors or in map keys, but that's not an issue for usages of selectors as long as you don't extend them, so this is rather an edge case to cover in the extend specs, and maybe it is already).

nex3:
@stof That's a fair point, although if we're going to encode that behavioral difference in the specs we should also add specs for the cases where that internal representation breaks. In addition to `@extend`, it will also cause the SassScript `&` to produce the wrong value.

stof:
@Cerdic @nex3 this PR should be closed as it has been taken over in #1620 which has been merged

PR #1604, opened 2020-12-15
Add support to ppc64le. Skip dartsass

gururajrkatti:
This PR adds CI support for the IBM Power Little Endian (ppc64le) architecture. The idea is to ensure that the builds on this architecture are continuously tested along with the Intel builds (amd64) as this is part of the ubuntu distro on that architecture as well and detecting (and fixing) any issues or failures early would help to ensure that we are always up to date.
PS: since dart sdk not available for ppc64le omitted that job.


nex3:
Since LibSass is deprecated, I don't think it makes sense to expand the scope of testing for it, especially on a platform that (as far as I know) no core maintainers have experience with or would be able to debug effectively.

PR #1605, opened 2020-12-18
Add specs for plain-CSS clamp()

### LINKED ISSUE #1599: Add specs for plain-CSS clamp()
See https://github.com/sass/sass/issues/2860

nex3:
Closes #1599
See sass/sass#2860

[skip dart-sass]


spec/core_functions/color/hsla/four_args/special_functions.hrx
```diff
@@ -256,3 +256,63 @@ a {
   b: hsla(1, 2%, 3%, max(0.4));
 }
 
+<===>
+================================================================================
+<===> clamp/arg_1/options.yml
```
jathak, line 261: Would it make sense to use a single `clamp/options.yml` file instead of separately disabling each test for libsass?

spec/css/functions/special.hrx
```diff
@@ -0,0 +1,24 @@
...
+a {b: clamp(0, 1, 2)}
+
+<===>
+================================================================================
+<===> clamp/punctuation/input.scss
+a {b: clamp(@#$%^&*({[]})_-+=|\\:""''<>,.?/)}
+
+<===>
+================================================================================
+<===> clamp/punctuation/input.scss
```
jathak, line 18: This is a duplicate of the previous test

jathak (approved):

spec/css/functions/special.hrx
nex3: Not sure I understand, I don't think the punctuation test is duplicated anywhere?

PR #1608, opened 2020-12-29
Add specs for angle unit deprecations

nex3:
Partially addresses #1607
See sass/sass#2904

[skip dart-sass]


spec/core_functions/color/hsl/three_args/units.hrx
```diff
@@ -16,6 +83,38 @@ a {
   b: #bf4040;
 }
 
+<===> saturation/unitless/warning
+WARNING: $saturation: Passing a number without unit % is deprecated.
```
jathak, line 87: Why are some of these deprecation warnings just `WARNING` while others say `DEPRECATION WARNING`?

jathak (approved):

spec/core_functions/color/hsl/three_args/units.hrx
nex3: Because I forgot to pass `deprecation: true` to one of my calls to `warn` :stuck_out_tongue_closed_eyes:

PR #1609, opened 2021-01-06
Migrate the CI setup to Github actions

stof:


stof:
See https://github.com/stof/sass-spec/actions/runs/466185101 for a test run in my own fork (PRs opened from a work won't run a workflow they introduce, due to some security restrictions).

I made 2 changes compared to the Travis setup:
- I added a job running the unit tests of the TS runner (I'm quite sure not having that job was a mistake when introducing the TS runner rather than an explicit choice)
- The skipping of jobs does not support passing a comma-separated list of implementations in the PR description anymore after the `skip` keyword (the github expression syntax for disabling jobs in the clean way does not support regexps). To skip multiple implementations for the same PR, the `skip` keyword needs to be repeated for each of them instead (I considered this trade-off OK since most PRs are only skipping one implementation anyway)

stof:
A future improvement here would be caching the compilation of sassc, given that this is what takes most of the time in this job and libsass does not change often (especially now that it is deprecated).

.github/workflows/ci.yml
```diff
@@ -0,0 +1,106 @@
+name: CI
+
+on:
+  push:
+    branches:
+      - master
+      # Feature branches beginning with "feature."
+      - feature.*
+  pull_request:
```
nex3, line 9:
```suggestion
  pull_request: {branches: ['**']}
```

Otherwise I don't think this will actually run.

nex3 (changes requested):

.github/workflows/ci.yml
stof: it will run. When you don't define `branches` or `tags`, it runs on all references

nex3: That wasn't my experience with https://github.com/sass/sass/pull/2992... it didn't run until I pushed https://github.com/sass/sass/pull/2992/commits/3386ec27dcf3cb726ae3984dfdfdb70d8a75b2a0.

stof: then the github doc would be wrong

stof: also, I have other repos where this works.

nex3: Looks like you're right, https://github.com/sass/sass/pull/2998 is running actions now. No idea why my previous attempt didn't work.

nex3:
Forked to #1612 to get the CI running.

PR #1614, opened 2021-01-09
Use the TypeScript runner for CI tests

nex3:


Awjin (approved):

README.md
```diff
@@ -25,12 +25,10 @@ implementations to ensure that they correctly implement the language.
 
 ## Running Specs
 
-Before running specs, you'll need to install [Ruby][] and [Bundler][]. The spec
-runner is written in Ruby, so it's necessary no matter which implementation
-you're testing. Then, from the root of this repo, run `bundle install`.
+Before running specs, you'll need to install [Node.js]. Then, from the root of
```
stof, line 28: this should document that it requires node 14+. The TS runner does not seem to work on node.js 12.

nex3: I think it specifically requires 14.14 or greater, according to package.json. Can you file a PR to add that note?

Issue #1531, opened 2020-05-02
Explicitly test for negative values in @for


connorskees:
Hello,

I wasn't able to find any tests for negative numbers in `@for`

e.g.
```scss
@for $i from -1 to 3 {
  ...
}

@for $i from 1 to -3 {
  ...
}

@for $i from -1 to -3 {
  ...
}
```

PR #1615, opened 2021-01-11
Refactor specs for the for at-rule to be conformant

stof:
This is the ground work for my PR adding specs covering the proper handling of units in `@for` for the fast-track proposal for https://github.com/sass/sass/issues/2931

As `@for` was only covered by non-conformant specs, this PR adds proper conformant specs for the existing behavior (the changes related to covering the fast-track proposal for units will be submitted in a separate PR).
This adds a few additional cases that were not covered before:
- empty loop
- backward loop
- floats in the loop bounds being an error (ironically, these ones need to be skipped for libsass, even though libsass complains saying that `"foo" is not an integer.` for the other spec covering the behavior for passing something else than a number)


stof:
The second commit adds more specs covering #1531

spec/directives/for/basic.hrx
```diff
@@ -0,0 +1,145 @@
+<===> inclusive_forward/input.scss
```
nex3, line 1: It's a good idea to [avoid placeholder path names](https://github.com/sass/sass-spec/blob/master/STYLE_GUIDE.md#do-use-descriptive-paths) like "basic", and try to find a more descriptive way to divide up the tests instead. In this case, you can just put everything under a single `for.hrx` file since there aren't that many specs yet.

spec/directives/for/basic.hrx
```diff
@@ -0,0 +1,145 @@
...
+<===>
+================================================================================
+<===> to_scope/input.scss
+// Overriding the variable inside the loop will not impact the end of the loop.
+$limit: 4;
+
+@for $i from 1 through $limit {
+  $limit: 2;
+  a {
+    b: $limit c $i;
```
nex3, line 126: This is a test case where the property name *isn't* irrelevant, so using a single-letter placeholder does more harm than good. If you write instead:

```suggestion
    limit: $limit;
    i: $i;
```

then it's clear when looking at the output which number is which.

nex3 (changes requested):

stof:
@nex3 I fixed your remarks.

nex3 (commented):
Thanks!

nex3 (approved):

PR #1617, opened 2021-01-12
Add specs for the handling of unit in for loops

### LINKED ISSUE #1616: Adds specs for the handling of units in for loops
See https://github.com/sass/sass/issues/2931

stof:
Currently includes commits from #1615. This will get cleaned once it is merged.

Closes #1616

skip dart-sass

### DIFF
--- a/spec/directives/for.hrx
+++ b/spec/directives/for.hrx
@@ -169,7 +169,7 @@ a {
 ================================================================================
 <===> unit/to_unitless/options.yml
 ---
-:todo:
+:ignore_for:
   - libsass
 
 <===> unit/to_unitless/input.scss
@@ -190,7 +190,7 @@ a {
 ================================================================================
 <===> unit/from_unitless/options.yml
 ---
-:todo:
+:ignore_for:
   - libsass
 
 <===> unit/from_unitless/input.scss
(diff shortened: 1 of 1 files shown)


spec/directives/for.hrx
```diff
@@ -149,6 +149,86 @@ a {
...
+  b: 4px;
+  b: 5px;
+}
+
+<===>
+================================================================================
+<===> unit/to_unitless/options.yml
+---
+:todo:
+  - libsass
```
stof, line 173: the libsass behavior is to require the same unit rather than requiring compatible units. That's why these 3 specs are marked as TODO. I haven't created an issue on libsass for that given that it is deprecated.

nex3: These should probably be `ignore_for` rather than `todo` since there's on expectation that LibSass will support them in the future.

stof: done.
Maybe the readme should be updated for its explanation of `ignore_for` as it is described as being used for deprecated features removed from modern implementation, not for modern features ignored for deprecated implementations. That's why I chose `todo` initially.

nex3 (approved):

PR #1620, opened 2021-02-10
Tests on escaping char in selectors: number, non alpha chars and non alpha chars in interpolation

stof:
This is taking over https://github.com/sass/sass-spec/pull/1600 to finish it


nex3 (approved):
If you've got time, I'd really like to see some tests that verify that the SassScript `&` for selectors with escapes does produce a consistent value (tests which will fail for LibSass).

PR #1630, opened 2021-03-16
Improve the way to exit from the spec runner

stof:
In node, write to stdout are asynchronous and can happen over multiple ticks of the event loop.
process.exit() forces the program to exit immediately, without waiting for any remaining pending asynchronous work, which can result in lost output if the output is not fully written yet.
Setting the process.exitCode property instead will configure the exit code to be used, while letting node terminate the process normally when nothing is pending in the event loop anymore.

See https://nodejs.org/api/process.html#process_process_exit_code for the doc about it.

### DIFF
--- a/lib-js/compiler.ts
+++ b/lib-js/compiler.ts
@@ -13,44 +13,40 @@ export interface Stdio {
 /**
  * A wrapper around a process that can compile Sass files.
  */
-export interface Compiler {
+export abstract class Compiler {
   /**
    * Run the compiler with the given args, at the path given as the cwd.
    */
-  compile(path: string, args: string[]): Promise<Stdio>
+  abstract compile(path: string, args: string[]): Promise<Stdio>
 
   /**
    * Shutdowns the compiler in case it was long-running
    */
-  shutdown(): void
+  shutdown(): void {}
 }
 
-/**
- * Returns a sass compiler that runs the given command.
- */
-export function executableCompiler(
-  command: string,
-  initArgs: string[] = []
-): Compiler {
-  return {
-    async compile(path, args) {
-      const { error, stdout, stderr, status } = child_process.spawnSync(
-        command,
-        [...initArgs, ...args],
-        {
-          cwd: path,
-          encoding: "utf-8",
-          stdio: ["ignore", "pipe", "pipe"],
-        }
-      )
-      if (error) {
-        throw new Error(`Failed to run executable compiler: ${error}`)
+export class ExecutableCompiler extends Compiler {
+  constructor(
+    private readonly command: string,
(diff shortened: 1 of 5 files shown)


Awjin:
Since `Tabulator` and `Interactor` use `process.stdin` and `process.stdout`, removing `process.exit()` most likely prevents the process from exiting and cleaning up properly. Could you verify this?

stof:
I don't think that's related to those, as the libsass run ended properly.
I think it is more the fact that the DartCompiler spaws a long-running process, which then keeps the event loop active. We need to shut it down cleanly at the end of the testsuite. I'll update my PR to account for that.

stof:
@Awjin here you go

lib-js/compiler.ts
```diff
@@ -43,10 +48,14 @@ export function executableCompiler(
       }
       return { stdout, stderr, status }
     },
+    shutdown() {
```
Awjin, line 51: Thanks for helping with this.

Instead of shimming the interface, consider making `Compiler` an abstract class with a default noop `shutdown()`, and then turning the function `executableCompiler()` into an `ExecutableCompiler` class that extends Compiler. Something like—

```
export abstract class Compiler {
  abstract compile(path: string, args: string[]): Promise<Stdio>;
  shutdown(): void {}
}

```

Awjin (changes requested):

lib-js/compiler.ts
stof: done. but this actually results in more code, not in less.

lib-js/compiler.ts
```diff
@@ -13,40 +13,47 @@ export interface Stdio {
...
-          stdio: ["ignore", "pipe", "pipe"],
-        }
-      )
-      if (error) {
-        throw new Error(`Failed to run executable compiler: ${error}`)
+export class ExecutableCompiler extends Compiler {
+  private readonly command: string
+  private readonly initArgs: string[]
+
+  constructor(command: string, initArgs: string[] = []) {
```
Awjin, line 32: TS lets you instantiate the members here

```
constructor(private readonly command: string, private readonly initArgs: string[] = [])
```

which lets us remove line 29, 30, 34, and 35. Same goes for `MockCompiler`

lib-js/compiler.ts
Awjin: IMO sometimes more code helps readability

Awjin (approved):
Looks good, thanks!

lib-js/compiler.ts
stof: oh, good to know. I didn't know that.

PR #1633, opened 2021-04-15
Fix some bugs in the TypeScript spec runner

nex3:


sass-spec.ts
```diff
@@ -11,36 +11,50 @@ const naughtyDirs = [
 ]
 
 async function runAllTests() {
-  const interactor = new Interactor(process.stdin, process.stdout)
-  const start = Date.now()
-  const args = await parseArgs(process.argv.slice(2))
-  const rootPath = args.root
-  const rootDir = await fromPath(rootPath)
-  const tabulator = new Tabulator(process.stdout, args.verbose)
+  let args_: CliArgs | undefined
```
Awjin, line 14: Might be unneeded? Typescript usually infers these correctly.

Awjin (approved):
Is there a reason we are checking in `package-lock`?

sass-spec.ts
nex3: In this case it didn't seem to (I think because it's referred to from within a callback?)

nex3 (commented):
Nat Allison checked in the lockfile. I don't have a strong opinion one way or the other.

PR #1634, opened 2021-04-19
Add a regression test for sass/dart-sass#1287

nex3:


jathak (approved):

nex3:
Diff view notwithstanding, the test case here does include trailing whitespace on the line in question. I verified that this case fails with 1.32.10 and succeeds with 1.32.11.

Tofandel:
Indeed, I checked the files the spaces are there, strange they are not showing in diff view

PR #1635, opened 2021-04-27
Add test for extending extenders across modules.

Awjin:
[skip dart-sass]


spec/directives/extend/across_modules.hrx
```diff
@@ -0,0 +1,74 @@
+<===> options.yml
```
nex3, line 1: There are already some `@extend` + `@use` specs in `spec/directives/use/extend.hrx`, so maybe these should go there as well?

spec/directives/extend/across_modules.hrx
```diff
@@ -0,0 +1,74 @@
...
+<===> extending_an_extender_multiple_times/extender_of_extender2.scss
+@use "extender";
+.extender_of_extender2 {
+  @extend .extender;
+  b: 2;
+}
+
+<===> extending_an_extender_multiple_times/input.scss
+@use "extender_of_extender1";
+@use "extender_of_extender2";
```
nex3, line 31: In other diamond dependency tests we call these `left`, `right`, and `downstream`... might be good to use the same convention here.

nex3 (approved):

spec/directives/extend/across_modules.hrx
Awjin: Good call, I'll move these specs there and rename them to be more consistent.

PR #1636, opened 2021-04-30
Add specs covering errors for non simple extended selectors

stof:
While working on scssphp, I discovered that these error cases were not covered anywhere in sass-spec (most of `@extend` is still covered by non-conformant specs, but even those don't cover those cases)

### DIFF
--- a/spec/directives/extend/error.hrx
+++ b/spec/directives/extend/error.hrx
@@ -1,0 +1,55 @@
+<===> complex/input.scss
+a b {
+  a: b;
+}
+c {
+  @extend a b;
+}
+
+<===> complex/error
+Error: complex selectors may not be extended.
+  ,
+5 |   @extend a b;
+  |           ^^^
+  '
+  input.scss 5:11  root stylesheet
+
+<===> complex/error-libsass
+Error: complex selectors may not be extended.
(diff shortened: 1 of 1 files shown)


spec/directives/extend/error.hrx
```diff
@@ -0,0 +1,55 @@
...
+================================================================================
+<===> compound/input.scss
+a:hover {
+  a: b;
+}
+b {
+  @extend a:hover;
+}
+
+<===> compound/output-libsass.css
```
jathak, line 34: I probably makes sense to mark this one as TODO for libsass, since the correct behavior is now an error, not a warning.

jathak (changes requested):

spec/directives/extend/error.hrx
stof: Well, we have lots of other cases in sass-spec that have a libsass alternative showing a deprecation warning instead of triggering an error (that one misses the `DEPRECATION` part though)

stof: @jathak but will it stay deprecated ? @mgreter is working on the 4.0 version of libsass in https://github.com/sass/libsass/pull/3135

jathak (approved):

PR #1637, opened 2021-05-03
Add specs cover content-exists in control statements

stof:
Replaces #1361


spec/core_functions/meta/content_exists.hrx
```diff
@@ -16,11 +21,6 @@ b {
   c: false;
 }
 
-<===> false/top_level/options.yml
```
stof, line 19: This change was caused by the interactive mode reordering the files to put `options.yml` first

jathak (approved):

PR #1638, opened 2021-05-03
Fix the delimiter comment

stof:


stof:
@nex3 @Awjin could this be merged ? It prevents me from using the latest sass-spec version in the scssphp testsuite (we have our own runner, to handle some output normalization and the handling of skipped tests or skipped warnings due to our implementation not being an official one supported directly in sass-spec, and our own runner relies on those separators instead of using an actual hrx parser which does not exist in PHP)

jathak (approved):

PR #1642, opened 2021-05-03
Add specs enforcing no units for str-slice

stof:
skip dart-sass


stof:
The reason to skip dart-sass is that I discovered that the dart-sass error message contains the wrong argument name. https://github.com/sass/dart-sass/pull/1303 is fixing it (and runs the specs from this PR)

jathak (approved):

PR #1645, opened 2021-05-10
Fix invalid todo markup (spec was still run for libsass)

mgreter:


jathak (approved):

stof:
this looks like an issue of the libsass runner (or maybe of the ruby runner). This YAML is equivalent (and is the one produced by the interactive mode when writing the file with the ts runner)

nex3:
I believe Ruby's YAML implementation interprets `:foo:` as a Symbol key but `':foo':` as a String key.

stof:
then, the interactive mode for the ts runner should be fixed to dumping it unquoted or the ruby runner should be updated to support string keys

nex3:
Yep: https://github.com/sass/sass-spec/issues/1649

PR #1646, opened 2021-05-11
Enable libsass different prefix in extender

mgreter:
skip libsass -> https://github.com/sass/libsass/pull/3136


jathak:
Going forward, can you make sure to wait to merge the implementation PR until the spec PR is also ready?

PR #1657, opened 2021-05-22
Adjust warning expectations to support sass/dart-sass#1323

nex3:
[skip dart-sass]

### DIFF
--- a/lib-js/compiler.ts
+++ b/lib-js/compiler.ts
@@ -92,7 +92,7 @@ export class DartCompiler implements Compiler {
 
   async compile(path: string, opts: string[]): Promise<Stdio> {
     this.stdin.write(`!cd ${path}\n`)
-    this.stdin.write([...this.initArgs, ...opts].join(" ") + "\n")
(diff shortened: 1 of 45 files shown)


Awjin (approved):

lib-js/compiler.ts
```diff
@@ -92,7 +92,7 @@ export class DartCompiler implements Compiler {
 
   async compile(path: string, opts: string[]): Promise<Stdio> {
     this.stdin.write(`!cd ${path}\n`)
-    this.stdin.write([...this.initArgs, ...opts].join(" ") + "\n")
+    this.stdin.write(["--verbose", ...this.initArgs, ...opts].join(" ") + "\n")
```
stof, line 95: shouldn't it also be added in the dart-sass `implArgs` in cli-args.ts, to ensure that dart-sass also runs in verbose mode when running specs against the dart-sass compiled executable rather than the source code ?

stof:
this broke the CI

PR #1658, opened 2021-05-22
Fix the arguments for running specs against the dart-sass command

stof:
Thet `--verbose` argument is needed by the `dart-sass` implementation for both implementations of the sass-spec Compiler, not only for the DartCompiler. Otherwise, specs are failing when running `npm run sass-spec -- --impl dart-sass -c $(which sass) spec/non_conformant/scss/precision`

### DIFF
--- a/lib-js/compiler.ts
+++ b/lib-js/compiler.ts
@@ -92,7 +92,7 @@ export class DartCompiler implements Compiler {
 
   async compile(path: string, opts: string[]): Promise<Stdio> {
     this.stdin.write(`!cd ${path}\n`)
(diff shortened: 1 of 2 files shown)


lib-js/compiler.ts
```diff
@@ -92,7 +92,7 @@ export class DartCompiler implements Compiler {
 
   async compile(path: string, opts: string[]): Promise<Stdio> {
     this.stdin.write(`!cd ${path}\n`)
-    this.stdin.write(["--verbose", ...this.initArgs, ...opts].join(" ") + "\n")
```
stof, line 95: it is removed here, because it is now part of `this.initArgs`

Awjin (approved):
Good catch, thanks!

PR #1661, opened 2021-05-29
Remove ignored specs covering the subject operator

stof:
As this operator (that was supported by ruby-sass) has been removed from the CSS Selectors spec, it is not supported by dart-sass and libsass. This means that these specs don't run anymore.


nex3 (approved):
Thanks for thinking of this!

PR #1663, opened 2021-06-02
Use gts style

nex3:


lib-js/interactor.ts
```diff
@@ -1,207 +1,207 @@
...
+      await choice.resolve(test);
+      rl.close();
+      return;
     }
 
-    while (true) {
-      this.printLine()
-      this.printLine(`In test case: ${test.dir.relPath()}`)
-      this.printLine(test.result().message)
+    for (;;) {
```
jathak, line 176: Is there a reason why gts prefers this? I can't find any mention of it in the style guide, so I wonder if this is a bug

jathak (approved):

lib-js/interactor.ts
nex3: See https://eslint.org/docs/rules/no-constant-condition. I think it's weird that they specifically choose to disallow `while (true)`, since that's *not* likely to be a bug, but it's definitely intentional.

PR #1665, opened 2021-06-02
Add test for `@for` inside property declaration

### LINKED ISSUE #1602: Specs for the use directive contain a suspicious spec
`directives/use/error/syntax/member/single_dot` looks suspicious to me. This spec is in the `directives/use` folder, but does not rely on `@use` at all: https://github.com/sass/sass-spec/blob/6015399e4b4ef8de2351e88d8acac3935b9ac00b/spec/directives/use/error/syntax.hrx#L78-L91

So either this spec is not covering the intended behavior, or it is in the wrong location

jathak:
[skip dart-sass]

### DIFF
--- a/spec/directives/for.hrx
+++ b/spec/directives/for.hrx
@@ -116,6 +116,24 @@ a {
 
 <===>
 ================================================================================
+<===> in_declaration/input.scss
+a {
+  b: {
+    @for $i from 1 through 5 {c: $i}
+  }
+}
+
+<===> in_declaration/output.css
+a {
+  b-c: 1;
+  b-c: 2;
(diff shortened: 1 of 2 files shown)


spec/directives/for.hrx
```diff
@@ -114,6 +114,24 @@ a {
   b: -1;
 }
 
+<===>
+================================================================================
+<===> in_declaration/input.scss
+a {
+  b: {
+    @for $i from 1 through 5 {c: $i;}
```
nex3, line 122: Nit: unnecessary semicolon.

nex3 (approved):

PR #1666, opened 2021-06-07
Add some specs for implicit dependencies

jathak:
[skip dart-sass]


nex3 (changes requested):
It seems like there are holes in the matrix of tested cases: what about `use_in_first`? `forward_second`? The combination of the two?

PR #1675, opened 2021-06-26
Add spec for configuring multiple configured forwards.

Awjin:
See https://github.com/sass/dart-sass/issues/1343

[skip dart-sass]

### DIFF
--- a/spec/directives/forward/with.hrx
+++ b/spec/directives/forward/with.hrx
@@ -160,6 +160,35 @@ d {
 
 <===>
 ================================================================================
+<===> facade_contains_multiple_configured_forwards/input.scss
+@use "midstream" with (
+  $a: twice-configured a,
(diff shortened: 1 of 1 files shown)


spec/directives/forward/with.hrx
```diff
@@ -158,6 +158,35 @@ d {
...
+================================================================================
+<===> facade_contains_multiple_configured_forwards/input.scss
+@use "midstream" with (
+  $a: twice-configured a,
+  $b: twice-configured b,
+);
+
+<===> facade_contains_multiple_configured_forwards/_midstream.scss
+@forward "upstream1" with ($a: configured a !default);
+@forward "upstream2" with ($b: configured b !default);
```
nex3, line 171: Might be worth also defining `$c` in `_midstream` and overriding it in `input`.

Issue #1678, opened 2021-07-24
Additional test case for `@extend`


connorskees:
Hello,

I found an interesting bug in my implementation of `@extend` that I think would be useful to add here. I'll try to submit a pr for this case sometime this week.

```scss
c b {
  @extend %d;
}

c > b {
  @extend %d;
}

%d {
  color: red;
}
```

PR #1680, opened 2021-07-31
add spec for trimming super selectors in `@extend`

connorskees:
This was a really interesting bug -- the only one I've had to actually clone the `dart-sass` repo for in order to track down. The bug ended up being related to this line
Captura de pantalla de 2021-07-30 23-37-40

https://github.com/sass/dart-sass/blob/6f17b4aa9ca879d79d07f83474bae164617570bd/lib/src/extend/extension_store.dart#L135

`Set.identity` constructs a set that uses _pointer_ equality, rather than true equality. In rust this is hard to model, so this behavior was overlooked. The fix was implemented by giving a unique id to every original. See https://github.com/connorskees/grass/commit/fccf93cd96740c6642fa4eddb623472a90b85b76

Resolves https://github.com/sass/sass-spec/issues/1678

### DIFF
--- a/spec/non_conformant/extend-tests/239_trims_super_selector_without_combinator.hrx
+++ b/spec/non_conformant/extend-tests/239_trims_super_selector_without_combinator.hrx
@@ -1,0 +1,17 @@
+<===> input.scss
+a b {
+  @extend %c;
+}
+
+a > b {
+  @extend %c;
+}
+
+%c {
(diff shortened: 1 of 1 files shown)


jathak:
Thanks for this! Can you move this spec to the proper conformant directory?

nex3:
Closing due to inactivity.

PR #1688, opened 2021-08-26
Add specs for first-class calc

nex3:
See sass/sass#818
[skip dart-sass]


spec/values/calculation/calc/error/known_incompatible/minus.hrx
```diff
@@ -0,0 +1,10 @@
+<===> input.scss
```
Awjin, line 1: Is this file a duplicate of `spec/values/calculation/calc/error/known_incompatible/operators.hrx` ?

spec/values/calculation/calc/simplify.hrx
```diff
@@ -0,0 +1,107 @@
...
+a {b: calc(1px + 2px + 1%)}
+
+<===> plus/left/output.css
+a {
+  b: calc(3px + 1%);
+}
+
+<===>
+================================================================================
+<===> plus/right/input.scss
```
Awjin, line 21: Might be worth adding a comment here explaining this behavior.

spec/values/calculation/max.hrx
```diff
@@ -0,0 +1,215 @@
...
+Error: c is not a number.
+  ,
+1 | a {b: max(c)}
+  |       ^^^^^^
+  '
+  input.scss 1:7  root stylesheet
+
+<===>
+================================================================================
+<===> error/syntax/rest/input.scss
```
Awjin, line 33: Should this be an error?

spec/values/calculation/min.hrx
```diff
@@ -0,0 +1,245 @@
...
+Error: c is not a number.
+  ,
+1 | a {b: min(c)}
+  |       ^^^^^^
+  '
+  input.scss 1:7  root stylesheet
+
+<===>
+================================================================================
+<===> error/syntax/rest/input.scss
```
Awjin, line 33: Should this be an error?

spec/values/calculation/min.hrx
```diff
@@ -0,0 +1,245 @@
...
+a {b: min(1px, 2px, 3%)}
+
+<===> perserved/unit/third/output.css
+a {
+  b: min(1px, 2px, 3%);
+}
+
+<===>
+================================================================================
+<===> preserved/math/first/output.css
```
Awjin, line 208: Here and below: some of these are missing `input.scss` files

Awjin (changes requested):
Overall LGTM, just some minor nits

spec/values/calculation/max.hrx
nex3: No—the rest argument causes this to be parsed as the global `max()` function—but it certainly shouldn't be tested here.

Awjin (approved):

PR #1689, opened 2021-08-27
Add spec for where pseudo-selector nesting

nstringham:
In this PR I attempt to add a spec for sass/sass#3130

I have never worked with the sass spec before so please let me know if I did something wrong

I have not tested this change on my local machine


nstringham:
Thanks @jathak

jathak (approved):

PR #1696, opened 2021-09-03
Add specs covering the CSS4 import conditions

### LINKED ISSUE #1446: There are no specs for CSS4 import supports condition
e.g. `@import url("narrow.css") supports(display: flex) handheld and (max-width: 400px);`
See https://www.w3.org/TR/css-cascade-4/#at-import

stof:
Closes #1446

### DIFF
--- a/spec/css/plain/import/conditions.hrx
+++ b/spec/css/plain/import/conditions.hrx
@@ -1,0 +1,92 @@
+<===> supports/declaration/options.yml
+:ignore_for:
+  - libsass
+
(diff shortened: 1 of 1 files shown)


jathak (commented):
Can you change the `:todo:`s here to `:ignore_for:`? Since libsass is deprecated now, there's no expectation that it will ever actually be implemented.

stof:
@jathak done

jathak (approved):

PR #1698, opened 2021-09-08
Add the ability to run shared JS API tests

nex3:
So far this only has one ultra-simple test, but we can build it out
once the core infrastructure is in place.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,6 +1,7 @@
 name: CI
 
 env:
+  PROTOC_VERSION: 3.x
   NODE_VERSION: 14
 
 
@@ -10,7 +11,7 @@ on:
 
 jobs:
   unit_tests_rb:
-    name: "Unit tests: Ruby"
+    name: "Unit tests | Ruby"
     runs-on: ubuntu-latest
 
     steps:
@@ -21,7 +22,7 @@ jobs:
         run: bundle exec rspec tests/
 
   unit_tests_ts:
-    name: "Unit tests: Typescript"
+    name: "Unit tests | Typescript"
     runs-on: ubuntu-latest
 
     steps:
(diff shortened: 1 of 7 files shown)


.github/workflows/ci.yml
```diff
@@ -95,3 +96,82 @@ jobs:
 
       - name: Run specs
         run: npm run sass-spec -- --impl libsass -c ../sassc/bin/sassc
+
+  dart_sass_js:
+    name: "JS API | Dart Sass | ${{ matrix.os }}"
```
stof, line 101: Does this sass-spec repo really need to test dart-sass on all OS ? The dart-sass repo already does it.

.github/workflows/ci.yml
```diff
@@ -95,3 +96,82 @@ jobs:
...
+    runs-on: ubuntu-latest
+    if: "github.event_name != 'pull_request' || !contains(github.event.pull_request.body, 'skip dart-sass')"
```
stof, line 103: should this be a different skipping tag than for the specs ? A PR will probably need to skip either the JS API test or the spec, depending on what it changes.

.github/workflows/ci.yml
nex3: When we add new tests here we need to verify that they pass on all OSes. It would be pretty bad if we added a test that failed on Windows, but only saw that failure the next time a pull request was sent out against Dart Sass.

.github/workflows/ci.yml
nex3: Most of the time, I expect we'll be updating *either* the language *or* the JS API, so skipping the other one won't matter either way. Even if we're updating both, though, they'll still be paired with one Dart Sass PR that implements both the language and JS API changes, and in that case we will want to skip both.

Plus, this was is much terser :grinning:.

Awjin (approved):

PR #1706, opened 2021-09-14
Fix a parenthesizing bug in calculations

nex3:
See sass/sass#3147
See sass/sass#3148
[skip dart-sass]

### DIFF
--- a/spec/values/calculation/calc/operator.hrx
+++ b/spec/values/calculation/calc/operator.hrx
@@ -28,12 +28,52 @@ a {
 
 <===>
 ================================================================================
-<===> minus/preserved/input.scss
+<===> minus/preserved/number/input.scss
 a {b: calc(1px - 2%)}
 
-<===> minus/preserved/output.css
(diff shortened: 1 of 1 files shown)


spec/values/calculation/calc/operator.hrx
```diff
@@ -28,14 +28,54 @@ a {
...
+<===>
+================================================================================
+<===> minus/preserved/minus/input.scss
+a {b: calc(1px - (2% - var(--c)))}
+
+<===> minus/preserved/minus/output.css
+a {
+  b: calc(1px - (2% - var(--c)));
+}
+
```
Awjin, line 78: Would it make sense to add similar tests for plus/times/divide?

Awjin (approved):

spec/values/calculation/calc/operator.hrx
nex3: We have them for multiplication and division below under "precedence", but I suppose it couldn't hurt to add them for plus as well.

PR #1709, opened 2021-09-15
Avoid using fully saturated colors for hue specs

stof:
Fully saturated colors are `white` for any hue. If the implementation converts the value to RGB instead of keeping the original HSL values, it might rightfully return 0deg in all cases previously, as that's indeed the hue of `white`


stof:
if this gets accepted, I can submit similar PRs for the specs for `saturation` and `lightness`, which suffer from the same issue.

stof:
@nex3  @jathak is anything missing for all the PRs I sent on this repo during the week ?

PR #1710, opened 2021-09-15
Add back support for min/max calculations

nex3:
See sass/sass#3142
See sass/sass#3150
[skip dart-sass]


spec/values/calculation/max.hrx
```diff
@@ -0,0 +1,313 @@
...
+1 | a {b: max(1px, 2px, 3s)}
+  |           ^^^ 1px
+  |                     == 3s
+  '
+  input.scss 1:11  root stylesheet
+
+<===>
+================================================================================
+<===> error/potentially_incompatible_before_unitless/input.scss
+a {b: max(1c, 2d, 3)}
```
Awjin, line 76: I'm not sure I understand this case. How does this differ from `max(1d, 2, 3e)`, which doesn't error?

spec/values/calculation/max.hrx
```diff
@@ -0,0 +1,313 @@
...
+a {b: max(1%, 2.5 + 0.9px)}
+
+<===> preserved/operation/unitless_and_unit/plus/output.css
+a {
+  b: max(1%, 3.4px);
+}
+
+<===>
+================================================================================
+<===> preserved/operation/unitless_and_unit/maxus/input.scss
```
Awjin, line 297:
```suggestion
<===> preserved/operation/unitless_and_unit/minus/input.scss
```

spec/values/calculation/max.hrx
```diff
@@ -0,0 +1,313 @@
...
+a {b: max(1%, 2.5 - 0.9px)}
+
+<===> preserved/operation/unitless_and_unit/maxus/output.css
```
Awjin, line 300:
```suggestion
<===> preserved/operation/unitless_and_unit/minus/output.css
```

spec/values/calculation/min.hrx
```diff
@@ -0,0 +1,373 @@
...
+<===>
+================================================================================
+<===> preserved/unit/third/input.scss
+a {b: min(1px, 2px, 3%)}
+
+<===> preserved/unit/third/output.css
+a {
+  b: min(1px, 2px, 3%);
+}
+
```
Awjin, line 333: Looks like these are duplicated tests

Awjin (approved):

spec/values/calculation/max.hrx
nex3: It has to do with the order the numbers are compared against one another. For `max(1d, 2, 3e)` the algorithm first compares `1d` to `2`, which is valid because it's comparing a number with a unit to a number without, and then compares `2` to `3e`, which is valid for the same reason. Here though it first compares `1c` to `1d`, which is invalid because we don't know how to convert between those units. This is one of those bad cases that arises out of allowing unitless numbers to work with numbers with units.

stof: isn't the algorithm comparing a number to all the following ones in `SassCalculation._verifyCompatibleNumbers` ?

nex3: In Dart Sass's implementation, `_verifyCompatibleNumbers` isn't called until after simplification occurs, and handling the semantics of `max()` is considered part of simplification.

stof: hmm, indeed. I forgot about that.

PR #1711, opened 2021-09-15
Update the specs of the warn directives for the styleguide

stof:


spec/directives/warn.hrx
```diff
@@ -0,0 +1,179 @@
...
+
+<===> position/function/warning-libsass
+WARNING: test
+         on line 2:3 of input.scss, in function `c`
+         from line 7:6 of input.scss
+
+<===>
+================================================================================
+<===> escaped/options.yml
+:todo:
```
Awjin, line 115: This should also be `:ignore_for:` since we are not expecting further libsass development

spec/directives/warn.hrx
```diff
@@ -0,0 +1,179 @@
...
+<===> errors/sass/semicolon/input.sass
+@warn value;
+
+<===> errors/sass/semicolon/error
+Error: semicolons aren't allowed in the indented syntax.
+  ,
+1 | @warn value;
+  |            ^
+  '
+  input.sass 1:12  root stylesheet
```
Awjin, line 179:
```suggestion
```

This looks like a test for the indented syntax, unrelated to `@warn`.

Awjin (changes requested):

spec/directives/warn.hrx
stof: this was just moving the existing spec around, consolidating `warn/*.hrx` into `warn.hrx` to avoid having only one spec per hrx file

Awjin: The consolidation looks good. The issue referenced in the `:todo:` is closed now, so we can safely change it to `:ignore_for:`.

stof: @Awjin updated.

Awjin (approved):

spec/directives/warn.hrx
Awjin: Thanks!

PR #1712, opened 2021-09-15
Remove duplicate non-conformant spec

stof:
Out of range values for the `saturate()` functions are already covered by the conformant specs.


Awjin (approved):
Thanks for the cleanup

PR #1723, opened 2021-09-26
Fix specs covering prefixed :is() for superselector specs

stof:
When reflecting changes from `matches.hrx` to `is.hrx`, the prefixed pseudos have not been replaced properly.


Awjin (approved):
Good eye, thanks!

PR #1727, opened 2021-10-05
Add specs for the new JS importer API

nex3:
See sass/sass#2509

[skip dart-sass]

### DIFF
--- a/js-api-spec/importer.test.ts
+++ b/js-api-spec/importer.test.ts
@@ -1,0 +1,633 @@
+// Copyright 2021 Google Inc. Use of this source code is governed by an
+// MIT-style license that can be found in the LICENSE file or at
+// https://opensource.org/licenses/MIT.
+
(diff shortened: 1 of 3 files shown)


js-api-spec/importer.test.ts
```diff
@@ -0,0 +1,633 @@
...
+            ],
+          });
+        }).toThrowSassException({
+          line: 0,
+          url: pathToFileURL('dir/_other.css'),
+        });
+      });
+    });
+
+    describe('fromImport is', () => {
```
Awjin, line 539: Should this also test `@forward` and `meta.load-css`?

Awjin (approved):

js-api-spec/importer.test.ts
nex3: Since we're already testing those for normal importers, I think it's fine to rely on this behavior being the same.

PR #1733, opened 2021-10-12
Allow js-api-spec to work seamlessly with TypeScript packages

nex3:
Rather than requiring the types to be stripped, this now exports the
TypeScript packages through a plain JS library with new type
annotations.


Awjin:
I see. Having the compiler check multiple directories makes sense.

PR #1739, opened 2021-10-27
Add specs for SassList and SassMap.

Awjin:
See https://github.com/sass/embedded-host-node/pull/68

[skip sass-embedded]

### DIFF
--- a/js-api-spec/value/list.test.ts
+++ b/js-api-spec/value/list.test.ts
@@ -61,15 +61,22 @@ skipForImpl('dart-sass', () => {
         );
       });
 
-      it("doesn't equal a value with different metadata", () => {
+      it("doesn't equal the same list with a different ordering", () => {
         expect(
           list.equals(
             new SassList(
-              [
-                new SassString('a', {quotes: false}),
-                new SassString('b', {quotes: false}),
-                new SassString('c', {quotes: false}),
-              ],
+              [new SassString('c'), new SassString('b'), new SassString('a')],
+              {separator: ','}
+            )
+          )
+        ).toBe(false);
+      });
+
+      it("doesn't equal a list with different metadata", () => {
+        expect(
+          list.equals(
+            new SassList(
+              [new SassString('a'), new SassString('b'), new SassString('c')],
               {separator: ' '}
             )
           )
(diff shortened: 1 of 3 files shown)


js-api-spec/value/list.test.ts
```diff
@@ -0,0 +1,283 @@
...
+                new SassString('b', {quotes: false}),
+                new SassString('c', {quotes: false}),
+              ],
+              {separator: ',', brackets: true}
+            )
+          )
+        ).toBe(false);
+      });
+
+      it("doesn't equal a value with different contents", () => {
```
nex3, line 92: Also test equality with a shorter list and a longer list that have the same contents where they overlap.

js-api-spec/value/list.test.ts
```diff
@@ -0,0 +1,283 @@
...
+        expect(list.hasBrackets).toBe(false);
+      });
+
+      it('supports null separators', () => {
+        const list = new SassList(
+          [new SassString('a'), new SassString('b'), new SassString('c')],
+          {separator: null}
+        );
+        expect(list.separator).toBe(null);
+      });
```
nex3, line 161: A list with more than one element shouldn't allow an undecided separator. If that's not in the protocol/spec yet, it should be.

js-api-spec/value/list.test.ts
```diff
@@ -0,0 +1,283 @@
...
+            string.sassIndexToListIndex(new SassNumber(2))
+          ).toThrow();
+          expect(() =>
+            string.sassIndexToListIndex(new SassNumber(-2))
+          ).toThrow();
+        });
+      });
+    });
+
+    describe('an empty list', () => {
```
nex3, line 244: Test that this equals another call to `SassList.empty()` and a call to `new SassList()` without any elements.

js-api-spec/value/map.test.ts
```diff
@@ -0,0 +1,210 @@
...
+      });
+
+      it('rejects invalid indices', () => {
+        expect(() => map.sassIndexToListIndex(new SassNumber(0))).toThrow();
+        expect(() => map.sassIndexToListIndex(new SassNumber(3))).toThrow();
+        expect(() => map.sassIndexToListIndex(new SassNumber(-3))).toThrow();
+      });
+    });
+
+    describe('an empty map', () => {
```
nex3, line 184: Similar additional equality tests here.

nex3 (changes requested):

js-api-spec/value/list.test.ts
Awjin: Good point. I updated the protocol here: https://github.com/sass/embedded-protocol/pull/81. For the spec, would it be easier to update it here https://github.com/sass/sass/pull/3183?

nex3 (approved):
One more thing: it turns out that initializing values outside of a `beforeEach()` clause breaks the latest version of jest, because it'll try to run them when determining how many tests were skipped. Can you move those into `beforeEach()`es like I've done for the other value tests?

Awjin:
Good catch, done.

PR #1740, opened 2021-11-04
Tests for inspecting nested empty bracketed lists

Awjin:
See https://github.com/sass/sass/issues/3160

[skip dart-sass]


nex3 (changes requested):
It seems like this is missing some important cases:

* Empty bracketed/unbracketed lists without inspect
* Empty unbracketed lists with inspect

Also, should the inspect behavior live in the inspect function's specs? I'm not sure what we do with other values.

spec/values/lists/brackets.hrx
```diff
@@ -1,52 +1,87 @@
...
-  append: [foo, bar, baz];
-  append-with-separator: [foo bar baz];
-  inspect-empty: [];
-  inspect-simple: [foo, bar];
-  inspect-nested-bracketed: [[foo]];
-  inspect-nested-unbracketed: [(foo bar)];
-  inspect-nested-unbracketed-comma: [foo bar,];
-  inspect-nested-unbracketed-singleton: [(foo,)];
+<===> empty/input.scss
+a { b: [] }
```
nex3, line 2:
```suggestion
a {b: []}
```

...and same below

nex3 (approved):

PR #1752, opened 2021-12-22
Add JS API specs for the SassFunction value type

nex3:
[skip sass-embedded]


jathak (approved):
LGTM aside from the static analysis errors

PR #1754, opened 2021-12-23
Fix specs to respect the styleguide

stof:
In scssphp, we rely on the separator comment to separate the specs (we don't detect spec directories through navigating the virtual file system of the hrx archive, because we don't have a proper hrx parser available in PHP) As such, using the wrong comment for the separator breaks those tests for us.

The second commit fixes a case where some specs were nested inside another spec directory by mistake, which does not work (those nested spec directories are not actually spec directories). As our testsuite outputs a warning when it skips a test containing multiple `input.scss` file, I detected those.


stof:
I'm writing a linter for specs (which will be submitted in another PR to run in the CI), which allowed me to find more cases where some specs were nested in an invalid way.

spec/core_functions/color/rgb/four_args/special_functions.hrx
```diff
@@ -282,44 +282,3 @@ a {b: rgb(1, 2, 3, unquote("clamp(0.4, 0.5, 0.6)"))}
 a {
   b: rgb(1, 2, 3, clamp(0.4, 0.5, 0.6));
 }
-
-<===>
-================================================================================
-<===> calc/string/arg_2/args/input.scss
```
stof, line 288: Those specs were not running at all due to being nested in `calc/string/arg_2/` and `var/arg_2/`.

As they are actually covering the case of special functions in the 2-args signature of the Sass rgb function and not in the 4-args one, I moved them to the proper hrx file to solve the nesting.

stof:
@nex3 @jathak has this been missed ?

Note that once it is merged, I'll open my PR adding the spec linter.

jathak:
Sorry for the delay! I'll take a look at this and your other open PRs soon (hopefully tomorrow)

jathak (approved):
Thanks!

PR #1755, opened 2021-12-23
Use non-white colors for the saturation tests

stof:
A color with 100% lightness is white, whatever its hue and saturation. For such colors, a roundtrip to the RGB values will not preserve the original saturation, making them a bad choice for the tests of the saturation function.

This is the same kind of fix than #1709


stof:
@nex3 @jathak has this been missed

jathak (approved):

PR #1759, opened 2021-12-31
Port legacy JS render() tests over from Dart Sass

nex3:


js-api-spec/legacy/render.test.ts
```diff
@@ -2,9 +2,221 @@
...
 // https://opensource.org/licenses/MIT.
 
+import * as p from 'path';
 import * as sass from 'sass';
 
-import {skipForImpl, captureStdioAsync} from '../utils';
+import {sandbox} from '../sandbox';
+import {skipForImpl, captureStdio, captureStdioAsync} from '../utils';
+
+skipForImpl('sass-embedded', () => {
```
jathak, line 11: Does embedded only support the new API, or are we skipping these for some other reason?

jathak (approved):

js-api-spec/legacy/render.test.ts
nex3: It currently only supports the `async()` render method, but eventually it will support the full API. I want to land a full test suite that's working with Dart Sass before I start implementing it in the embedded host.

PR #1772, opened 2022-01-17
Add a linter for specs

stof:
This linter validates that specs follow some rules of the styleguide (not all rules are validated for now. This can be expanded in the future if needed). For now, 3 things are validated:

- it detects invalid nesting of specs that would cause them not to run (see #1754 that fixed existing mistakes that I found while building that tool)
- it detects invalid spec directories having both `input.sass` and `input.scss`
- it enforces the formatting of hrx files to match the styleguide rules (sorting files in spec directories, having a separator comment between them, etc...). Those formatting errors are auto-fixable by passing the `--fix` option (reusing the formatting logic used by the interactive mode)


stof:
@nex3 @jathak any blocker for merging this ?

jathak:
Sorry for the delay on reviewing! We've been backlogged with other tasks, but I'll try to review soon

jathak (approved):
Thanks for this! (and sorry again for taking so long to review)

PR #1781, opened 2022-03-15
Add specs for calcs within supports

jathak:
See sass/sass#3259.

[skip dart-sass]


nex3 (changes requested):
In theory a calculation can appear in the left-hand side as well, so we should test that. Also, we should verify that calculations within interpolations (in either `SupportsDeclaration`s *or* `InterpolatedAnyValue`s) are still simplified.

nex3 (approved):

PR #1785, opened 2022-04-07
Fix and add specs for relative imports with `compileString()`

nex3:
The newly-tested behavior matches the specification as updated by
sass/sass#3278.

[skip sass-embedded]


nex3:
cc @ntkme

jathak (approved):

PR #1787, opened 2022-04-16
Add specs covering CSS imports with layer conditions

stof:
See https://github.com/sass/sass/issues/3285

skip dart-sass

### DIFF
--- a/spec/css/plain/import/conditions.hrx
+++ b/spec/css/plain/import/conditions.hrx
@@ -1,3 +1,50 @@
+<===> layer/options.yml
+:ignore_for:
+  - libsass
+
+<===>
+================================================================================
+<===> layer/default/input.scss
+@import "a.css" layer;
+
+<===> layer/default/output.css
+@import "a.css" layer;
+
+<===>
+================================================================================
+<===> layer/named/input.scss
+@import "a.css" layer(b);
+
+<===> layer/named/output.css
+@import "a.css" layer(b);
+
+<===>
+================================================================================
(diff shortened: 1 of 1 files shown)


spec/css/plain/import/conditions.hrx
```diff
@@ -149,6 +181,30 @@ $a: "a";
...
+<===> media/list/after_ident/input.scss
+@import "a" b, (c: d) and (e: f), g;
+
+<===> media/list/after_ident/output.css
+@import "a" b, (c: d) and (e: f), g;
+
+<===>
+================================================================================
+<===> media/list/after_feature/input.scss
+@import "a" (b: c), (d: e) and (f: g), h;
```
stof, line 195: this looks wrong to me. The `,` separates several ImportArgument, and `(d: e)` is not a valid ImportUrl

nex3: This is valid CSS! An `@import` [may be followed by a `<media-query-list>`](https://www.w3.org/TR/css-cascade-5/#at-import), and `(b: c), (d: e) and (f: g), h` is [a valid `<media-query-list>`](https://www.w3.org/TR/mediaqueries-4/#typedef-media-query-list).

jathak (approved):

spec/css/plain/import/conditions.hrx
stof: The the spec is ambiguous when allowing multiple ImportArgument

PR #1790, opened 2022-06-07
test case for loud comments in file imported twice in a row

Goodwine:
See https://github.com/sass/dart-sass/pull/1713
[skip dart-sass]


lib-js/test-case/compare.ts
```diff
@@ -83,7 +83,7 @@ export function compareResults(
     return failures.MissingOutput();
   } else if (expected.isSuccess) {
     if (!actual.isSuccess) {
-      return failures.UnexpectedError();
+      return failures.UnexpectedError(actual.error);
```
Goodwine, line 86: BTW I should mention for context - if you look at the bug report the error is very cryptic and not easy to read at all, by adding this I found the error to be more useful

spec/css/comment.hrx
```diff
@@ -161,6 +161,25 @@ a {
   b: c e;
 }
 
+<===>
+================================================================================
+<===> imports/input.scss
```
nex3, line 166: I think this should go somewhere in `spec/directives/import`.

PR #1791, opened 2022-06-13
Add a spot check for directories that don't contain tests

### LINKED ISSUE #1433: Add spot-check tests for folders without specs
Occasionally, directories get left without full specs either because a spec wasn't fully deleted or because of typos like the one fixed in https://github.com/sass/sass-spec/pull/1431. We should have a sanity check that verifies that every directory under `spec/` contains either a spec or subdirectories.

nex3:
Closes #1433


Goodwine (approved):
LGTM sans the linter comment, very interesting findings out of the new test

PR #1794, opened 2022-06-16
Add tests for parsing the `var()` function

nex3:
See sass/sass#3245

[skip dart-sass]


spec/css/functions/var.hrx
```diff
@@ -0,0 +1,338 @@
...
+<===> css_function/two_argument/dynamic/output.css
+a {
+  b: var(--c, 3);
+}
+
+<===>
+================================================================================
+<===> css_function/two_argument/rest/input.scss
+$args: --c, d;
+a {b: var($args...)}
```
Goodwine, line 61: If I had
```scss
$args: --c, d;
a {b: var($args..., )}
```
or
```scss
$args: --c;
a {b: var($args...,)}
```

Should those cases pass with `a { b: var(--c, d) }` and `a { b: var(--c, ) }` respectively?
or would both be considered errors?

nex3: That's actually tested on line 275. They're both valid, but they ignore the comma, because Sass generally allows trailing commas in argument lists, maps, and so on.

Goodwine: I see, I did notice that test but assumed that `@function var` would make the special handling to work differently, but if both are meant to be similar then this is fine

Goodwine (approved):

PR #1796, opened 2022-06-22
Repro dart-sass#1716

Goodwine:
Repro https://github.com/sass/dart-sass/issues/1716

Taking example - https://github.com/KaelWD/sass-forward-twice

1. Removed tab directory and `@forward` rules related to them (same error)
2. Renamed variables in `use module with {...}` to non-existing variables (same error)
3. Only used 1 valid variable (passed?)
4. Minimal repro attached (same error)

[skip dart-sass]
See https://github.com/sass/dart-sass/pull/1739

Blocked on
- https://github.com/sass/sass/pull/3357


spec/directives/forward/with.hrx
```diff
@@ -656,3 +656,65 @@ b {
...
   midstream2: overridden 2;
 }
+
+<===>
+================================================================================
+<===> use/multiple/same_file_works/input.scss
+@use 'module' with (
+  $list-variable: 'foo',
+  // $other: 'baz', <- uncommenting this is OK
+  // $whatever: 'bar', <- uncommenting this causes the module loaded twice err
```
Goodwine, line 666: Because of the behavior with these two
I am thinking that when a variable exists in all `@forwarded` declarations then it's fine, but when there is a variable that does not exist somehow it triggers the "loaded twice" error

So whatever is dedupping the module seems to have a quirk with variables

or something like that :)

spec/directives/use/error/with.hrx
```diff
@@ -351,3 +351,109 @@ Error: This variable was not declared with !default in the @used module.
...
+These tests verify the configuration is reused by opaque ID across multiple
+source files preventing unexpected behaviors when variables are declared on
+separate files with validations that occur when a module is used more than once.
+
+
+<===>
+================================================================================
+<===> unused_distributed_vars/single_use/input.scss
+@use 'module' with (
+  $foo: 'foo',
```
nex3, line 370: Nit: [Use single-letter names for irrelevant placeholders](https://github.com/sass/sass-spec/blob/main/STYLE_GUIDE.md#do-use-single-letter-names-for-irrelevant-placeholders).

Goodwine: done

nex3 (approved):

PR #1802, opened 2022-07-06
Fix CI when declaring -dev dependencies

Goodwine:
(initial commit is empty for testing but I'll overwrite it)


Goodwine:
Nothing to do there, see:
https://github.com/sass/embedded-host-node/pull/153
https://github.com/sass/dart-sass-embedded/pull/92
Those should be the fixes

PR #1806, opened 2022-07-19
Add specs for passing non-deg units to hwb's hue

nex3:
See #1607

[skip dart-sass]

### DIFF
--- a/spec/core_functions/color/hwb/three_args/units.hrx
+++ b/spec/core_functions/color/hwb/three_args/units.hrx
@@ -6,3 +6,55 @@ a {b: color.hwb(0deg, 30%, 40%)}
 a {
   b: #994d4d;
 }
+
+<===>
+================================================================================
+<===> hue/rad/input.scss
+@use 'sass:color';
+a {b: color.hwb(1rad, 30%, 40%)}
+
+<===> hue/rad/output.css
+a {
+  b: #994e4d;
+}
+
+<===> hue/rad/warning
+DEPRECATION WARNING: $hue: Passing a unit other than deg (1rad) is deprecated.
+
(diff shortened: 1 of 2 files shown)


spec/core_functions/color/hwb/three_args/units.hrx
```diff
@@ -6,3 +6,55 @@ a {b: color.hwb(0deg, 30%, 40%)}
...
+}
+
+<===> hue/rad/warning
+DEPRECATION WARNING: $hue: Passing a unit other than deg (1rad) is deprecated.
+
+You're passing 1rad, which is currently (incorrectly) converted to 1deg.
+Soon, it will instead be correctly converted to 57.2957795131deg.
+
+To preserve current behavior: $hue * 1deg/1rad
+To migrate to new behavior: 0deg + $hue
```
Goodwine, line 28: Does `0dev + $hue` mean that users will have to always do this until dart-sass 2.0.0?
I wonder if we should give another alternative like doing the calculation to deg and saying
```
To migrate to new behavior, either:
 - force-cast to deg units with: 0deg + $hue
 - manually update to deg units: 57.3 deg
```

Goodwine (approved):

spec/core_functions/color/hwb/three_args/units.hrx
nex3: Once we release Phase 2 non-`deg` angles will be automatically converted to degrees. `0deg + $hue` is just a stop-gap measure to force the conversion before Phase 2 is released.

I'm always wary about suggesting literal values in these, because I expect a lot of the time the argument comes from something that is itself dynamic and so can't be reliably replaced by a single value.

PR #1812, opened 2022-08-09
Improve error messages for incorrect units in color functions

nex3:
See sass/dart-sass#1745

[skip dart-sass]

### DIFF
--- a/spec/core_functions/color/change_color/error/bounds.hrx
+++ b/spec/core_functions/color/change_color/error/bounds.hrx
@@ -40,6 +40,29 @@ Error: argument `$red` of `change-color($color, $red: false, $green: false, $blu
 
 <===>
 ================================================================================
+<===> red/unit/input.scss
+// This test covers sass/dart-sass#1745, but should be removed once units are
+// fully forbidden (sass/sass#3374).
+a {b: change-color(red, $red: 300px)}
+
+<===> red/unit/error
+Error: $red: Expected 300px to be within 0 and 255.
+  ,
+3 | a {b: change-color(red, $red: 300px)}
+  |       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+  '
+  input.scss 3:7  root stylesheet
+
+<===> red/unit/error-libsass
+Error: argument `$red` of `change-color($color, $red: false, $green: false, $blue: false, $hue: false, $saturation: false, $lightness: false, $alpha: false)` must be between -0 and 255
+        on line 3:7 of input.scss, in function `change-color`
+        from line 3:7 of input.scss
+>> a {b: change-color(red, $red: 300px)}
+
+   ------^
+
+<===>
+================================================================================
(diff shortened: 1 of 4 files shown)


spec/core_functions/color/fade_out.hrx
```diff
@@ -187,3 +187,26 @@ Error: argument `$amount` of `fade-out($color, $amount)` must be between -0 and
...
+<===> error/bounds/unit/error
+Error: $amount: Expected 50% to be within 0 and 1.
+  ,
+3 | a {b: fade-out(red, 50%)}
+  |       ^^^^^^^^^^^^^^^^^^
+  '
+  input.scss 3:7  root stylesheet
+
+<===> error/bounds/unit/error-libsass
+Error: argument `$amount` of `fade-out($color, $amount)` must be between -0 and 1
```
Goodwine, line 207: Does this mean libsass handled this somewhat correctly already and the tests are just showing what it looks like currently?

spec/core_functions/color/change_color/error/bounds.hrx
```diff
@@ -206,6 +308,39 @@ Error: argument `$lightness` of `change-color($color, $red: false, $green: false
...
 
+<===>
+================================================================================
+<===> lightness/unit/input.scss
+// This test covers sass/dart-sass#1745, but should be removed once units are
+// fully forbidden (sass/sass#3374).
+a {b: change-color(red, $lightness: 200px)}
+
+<===> lightness/unit/error
+DEPRECATION WARNING: $lightness: Passing a number without unit % (200px) is deprecated.
```
Goodwine, line 319: I feel like "a number without unit %" may sound strange to users who may not realize % is considered a unit in sass

Edit: I guess that's out of scope for this change

nex3: Yeah, I think this is something we just have to expect users to learn at some point.

spec/core_functions/color/fade_out.hrx
```diff
@@ -187,3 +187,26 @@ Error: argument `$amount` of `fade-out($color, $amount)` must be between -0 and
...
+
+<===>
+================================================================================
+<===> error/bounds/unit/input.scss
+// This test covers sass/dart-sass#1745, but should be removed once units are
+// fully forbidden (sass/sass#3374).
+a {b: fade-out(red, 50%)}
+
+<===> error/bounds/unit/error
+Error: $amount: Expected 50% to be within 0 and 1.
```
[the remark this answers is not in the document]
nex3: I think that would imply more strongly that passing a number with a unit would also be an error, which isn't true (yet). Eventually we'll have a proper deprecation warning which will be more explicit about this.

spec/core_functions/color/fade_out.hrx
nex3: Yeah, I guess so!

Goodwine (approved):

PR #1817, opened 2022-08-19
Linter and fixer for HRX files over 500 lines

### LINKED ISSUE #1805: Add a tool to split HRX files into subdirectories
The [style guide](https://github.com/sass/sass-spec/blob/main/STYLE_GUIDE.md#dont-have-hrx-files-longer-than-500-lines-or-so) puts a soft 500-line limit on the size of HRX files, but it's currently a fairly onerous manual process to convert a single file into multiple physical subdirectories. We should add a tool that creates a separate subdirectory for each top-level virtual directory in a given HRX file.

Goodwine:
Also refactored lint-spec reporter to be reusable

Fixes #1805


spec/libsass/bourbon.hrx
```diff
@@ -1,3 +1,11 @@
+<===> options.yml
+
+:todo:
+# This file starts failing when split, sass compiler seems to struggle when
+# `input.scss` is a real file with `@import "a/b/c"` and `a/b` is actually an
+# HRX file containing `c` in its root.
+- lint-hrx
+
```
Goodwine, line 8: For some reason this file could not be split, it started throwing errors, so I figured I should leave it as-is for now which means this tool now has a way to skip HRX that shouldn't get split for whatever reason

Goodwine:
I split the commits so the changes are on the first two, hopefully that makes it easier to review

lib-js/reporter.ts
```diff
@@ -0,0 +1,81 @@
...
+/**
+ * This reporter is used by linters to report findings and add annotations used
+ * by GitHub actions to display enhance the output, for example making paths to
+ * the problematic files link to the file in a GitHub repo.
+ *
+ * This reporter will print out a notice suggesting users to add the flag
+ * `--fix` on their linter to automatically fix the linter findings reported
+ * as being fixable.
+ */
+export class GithubActionsReporter {
```
nex3, line 12: I think the original name, `Reporter`, is better—although this *can* print GitHub Actions annotations, it also runs as just a normal human-friendly reporter if the `GITHUB_ACTIONS` environment variable isn't set.

.github/workflows/ci.yml
```diff
@@ -57,6 +57,7 @@ jobs:
         with: {node-version: "${{ env.NODE_VERSION }}"}
       - run: npm install
       - run: npm run lint-spec
+      - run: npm run lint-hrx
```
nex3, line 60: Why create a separate entrypoint for this rather than just defining it as effectively an additional kind of lint in `lint-spec`?

lint-hrx.ts
```diff
@@ -0,0 +1,161 @@
...
+
+/** The number of lines allowed in HRX files. */
+const LINE_THRESHOLD = 500;
+
+/**
+ * Verifies all HRX files in the project `spec` directory are under the
+ * {@link LINE_THRESHOLD}.
+ *
+ * HRX files are unwrapped and transformed into real files when
+ * {@link reporter.fix} is true.
```
nex3, line 19: It seems like a layering mismatch to have the reporter be responsible for tracking whether the linter fixes errors. I know you inherited that from the previous code, but if possible it would be good to fix rather than carrying it forward.

lint-hrx.ts
```diff
@@ -0,0 +1,161 @@
...
+ *
+ * Skips HRX files if they include a top-level `options.yml` file annotated with
+ * :todo: or :ignore-for: `lint-hrx`. For example:
```
nex3, line 22: Nit:

```suggestion
 * `:todo`: or `:ignore-for:` `lint-hrx`. For example:
```

lint-hrx.ts
```diff
@@ -0,0 +1,161 @@
...
+ *
+ * ```hrx
+ * <===> options.yml
+ * # Ignored because <SOME REASON>
+ * :todo:
+ * - lint-hrx
```
nex3, line 28: Might be a good idea to also allow sass-spec issue references, to match how we do TODOs for implementations.

lint-hrx.ts
```diff
@@ -0,0 +1,161 @@
...
+
+  // No Errors.
+  if (tooBig.length === 0) return;
+
+  if (!reporter.fix) {
+    const tooBigRelative = tooBig.map(f => p.relative(process.cwd(), f));
+    for (const f of tooBigRelative) {
+      reporter.reportError(`HRX file exceeds ${LINE_THRESHOLD} lines`, f, true);
+    }
+    return;
```
nex3, line 43: Nit: since the `else` branch is approximately the same size as the `if` branch here, I'd just use an `else` explicitly rather than an early return.

lint-hrx.ts
```diff
@@ -0,0 +1,161 @@
...
+import {fromPath, SpecDirectory} from './lib-js/spec-directory';
+import RealDirectory from './lib-js/spec-directory/real-directory';
+import VirtualDirectory from './lib-js/spec-directory/virtual-directory';
+import {GithubActionsReporter} from './lib-js/reporter';
+
+/** The number of lines allowed in HRX files. */
+const LINE_THRESHOLD = 500;
+
+/**
+ * Verifies all HRX files in the project `spec` directory are under the
```
nex3, line 15: Should "in the project `spec` directory" be "in `rootPath`"?

lint-hrx.ts
```diff
@@ -0,0 +1,161 @@
...
+ * such as number of lines in an HRX file.
+ */
+async function bigHrxInTree(
+  dir: SpecDirectory,
+  reporter: GithubActionsReporter
+): Promise<string[]> {
+  if (!(dir instanceof RealDirectory)) return [];
+
+  return [
+    ...(await bigHRX(dir, reporter)),
```
nex3, line 109: I don't think you need parens here.

lint-hrx.ts
```diff
@@ -0,0 +1,161 @@
...
+      )
+    ).flat(),
+  ];
+}
+
+/**
+ * Returns the absolute paths to all HRX files in a single {@link dir} that
+ * exceed the {@link LINE_THRESHOLD} into the global set {@link tooBig}.
+ */
+async function bigHRX(
```
nex3, line 122:
```suggestion
async function bigHrx(
```

lint-hrx.ts
```diff
@@ -0,0 +1,161 @@
...
+      await Promise.all(
+        (await dir.subdirs()).map(d => bigHrxInTree(d, reporter))
+      )
+    ).flat(),
+  ];
+}
+
+/**
+ * Returns the absolute paths to all HRX files in a single {@link dir} that
+ * exceed the {@link LINE_THRESHOLD} into the global set {@link tooBig}.
```
nex3, line 120: I think this "into the global set" line is outdated?

nex3 (changes requested):
I'm assuming all the HRX changes are automated fixes and don't need review.

.github/workflows/ci.yml
Goodwine: it was mostly to avoid touching the reporter which had some quirks when used on 2 linters, but I can merge it

lint-hrx.ts
Goodwine: given how the linters using the reporter work, the `fix` isn't even needed within the reporter 😱

lint-hrx.ts
Goodwine: They are added by the auto formatter 🤔

lint-hrx.ts
Goodwine: it is 😅

lint-hrx.ts
Goodwine: looking at how it's done, it just looks at whether it includes a substring, so `sass/libsass#123` doesn't do anything specific about being a github issue, it just happens to contain the `impl` value (libsass in this case)

With that in mind I updated it to contain `sass/sass-spec#`

.github/workflows/ci.yml
```diff
@@ -13,7 +13,7 @@ on:
 
 jobs:
   unit_tests_rb:
-    name: "Unit tests | Ruby"
+    name: 'Unit tests | Ruby'
```
nex3, line 16: Can you get your editor to stop reformatting every YAML file it touches? :laughing:

lib-js/reporter.ts
```diff
@@ -0,0 +1,85 @@
+import {Writable} from 'stream';
```
nex3, line 1: For files that export one primary class, the filename should match that classname (so `lint-reporter.ts`).

lib-js/reporter.ts
```diff
@@ -0,0 +1,85 @@
...
+import p from 'path';
+
+/**
+ * This reporter is used by linters to report findings it can also be used by
```
nex3, line 5:
```suggestion
 * This reporter is used by linters to report findings. It can also be used by
```

lint-spec.ts
```diff
@@ -173,14 +103,163 @@ function getReportedInputLocation(directory: SpecDirectory): string {
...
+ * HRX files are unwrapped and transformed into real files when {@link fix} is
+ * `true`.
+ */
+async function lintHrxSize(
+  rootPath: string,
+  reporter: LintReporter,
+  {fix = false}
+) {
+  const rootDir = await fromPath(rootPath);
+  const tooBig = await bigHrxInTree(rootDir, reporter);
```
nex3, line 132: Does it still make sense for this function to do its own recursive traversal of the test tree separate from `lintDirectory()`?

package.json
```diff
@@ -64,5 +64,8 @@
   },
   "engines": {
     "node": ">=v14.14.0"
+  },
+  "devDependencies": {
+    "file-lines-count": "^1.0.1"
```
nex3, line 69: Since we don't have a distinction between development and deployment, all dependencies should just live in `"dependencies"`.

nex3 (changes requested):

lib-js/reporter.ts
Goodwine: 😅

.github/workflows/ci.yml
Goodwine: 😅

lint-spec.ts
Goodwine: The current implementation of `lintDirectory()` + `lintNestedTestDirectories()` makes it slightly awkward to merge `lintHrxSize()` into the `lintDirectory()` because it stops traversing when the directory `isTestDir` but the HRX size linter wants to keep traversing down

IMO this needs to either be refactored to separate traversal and linting or keep current implementation of each "sub-linter" running their own traversal

I think it makes sense to leave it as-is for now, and think about refactoring when another "linter" comes by

nex3 (approved):

stof:
should we really allow splitting a directory that is actually a test directory into separate files ? This would loose the benefit of co-locating the input and output in the same file.

nex3:
That's a good point. I hadn't thought of that, since a style-guide-compliant test will almost certainly never hit the limit, but for the sake of older tests we should probably add an exception to the linter.

PR #1824, opened 2022-09-14
hrx linter no longer validates or breaks down HRX where the root is a unit test

### LINKED ISSUE #1821: spec linter should allow large HRX files if they are test units
Related: https://github.com/sass/sass-spec/pull/1817

Currently the linter (and fixer) will break up any HRX file longer than 500 lines
This may be too extreme for the sake of unit-testing because it separates inputs and outputs

By doing this we can also merge the directory traversal logic since we'll stop at "test dirs"

Goodwine:
recombines all existing directory unit tests into HRX files reverting files breaking up on #1817

Fixes #1821

### DIFF
--- a/lint-spec.ts
+++ b/lint-spec.ts
@@ -245,9 +245,12 @@ async function bigHrx(
 
   const tooBig: string[] = [];
   for (const f of hrxFiles) {
+    const subDir = await fromPath(f);
+    if (subDir.isTestDir()) continue;
+
     reporter.reportLintedFile(f);
 
-    const options = await (await fromPath(f)).options();
+    const options = await subDir.options();
     const linterAnnotation =
       options.getMode('lint-hrx') ?? options.getMode('sass/sass-spec#');
     // Skip XHR files annotated with :ignore-for: or :todo: containing


nex3 (approved):

stof:
This PR does not seem to recombine directories that were broken up in the previous PR

Goodwine:
The way I generated this PR:
```sh
git checkout main
git pull
git checkout -b hrx-500
git rebase --onto ${PR_SHA}~1 $PR_SHA hrx-500 #basically remove PR_SHA
git checkout main -- package.json tsconfig.json lint-spec.ts lib-js/lint-reporter.ts
# then changes to lint-spec.ts
_changesAfterPR=$(git diff $PR_SHA main)
_newChanges=$(git diff ${PR_SHA}~1 hrx-500)
diff <(echo "$_changesAfterPR") <(echo "$_newChanges")> # only diff was lint-spec.ts
npm run lint-spec -- --fix
git rebase main --interactive # and squashed any commits in between
```
Is there a particular directory that should not have been split? (the only non-splittable directory should have been HRX file that contains at most 1 unit test)

stof:
Look for `input.scss` files in that repo. Most of them should be unsplit.
And the fact that the unsplitting is not done is easy to see: the diff of this PR does not include any modified spec files.

> ```shell
> it rebase main --interactive # and squashed any commits in between
> ```

This might have been the issue: it would have kept the split already done in main in between (as those files would not be modified in your branch).

stof:
I did the recombination work manually in https://github.com/sass/sass-spec/pull/1827

PR #1826, opened 2022-09-15
Add specs for universal floating-point numbers

### LINKED ISSUE #1819: Update tests for floating-point numbers
See https://github.com/sass/sass/issues/2892

nex3:
Closes #1819

[skip dart-sass]


jathak (approved):

spec/core_functions/math/acos.hrx
```diff
@@ -55,7 +55,7 @@ a {b: math.acos(1.000000000001)}
 
 <===> one_fuzzy/output.css
 a {
-  b: 0deg;
+  b: NaNdeg;
```
stof, line 58: shouldn't the case of being fuzzy equal to the bound be handled in `math.acos` instead of returning NaN ?

spec/core_functions/math/round.hrx
```diff
@@ -48,14 +48,18 @@ a {
...
 // This is the smallest number that's representable as a float and in the
-// precision range to be considered equal to 5.
-a {b: round(0.4999999999900001)}
+// precision range to be considered equal to 0.5.
+a {b: round(0.499999999995)}
 
 <===> up/within_precision/output.css
 a {
-  b: 1;
+  b: 0;
```
stof, line 62: This is not consistent with the test name, which says it would round *up*

spec/core_functions/math/acos.hrx
stof: OK, I missed https://github.com/sass/sass/commit/498fda605cb2e872312123a5ca56746d995423c5 so this is actually expected

spec/core_functions/math/round.hrx
nex3: Good call; this test is now redundant with `down/within_precision`, so I'll remove it.

PR #1827, opened 2022-09-17
Recombine big spec tests as hrx

stof:
This recombines spec directories that were broken down to separate files instead of HRX in #1817


stof:
I went one step further, to also recombine test directories that were not using HRX even before #1817 by adding that to the linter itself.
The same `lint-hrx` TODO that is usable to skip the rule about big HRX can be used to skip the rule about not using HRX. I'm relying on it to allow using real files when dealing with non-UTF8, as those cannot be represented as HRX.

lint-spec.ts
```diff
@@ -113,6 +115,37 @@ function getRelativePath(filepath: string): string {
...
+    reporter.reportError(
+      `The test directory "${getRelativePath(
+        directory.path
+      )}" must use the HRX format.`,
+      getReportedInputLocation(directory)
+    );
+  } else {
+    fs.writeFileSync(`${directory.path}.hrx`, await directory.asArchive());
+
+    fs.rmSync(directory.path, {recursive: true});
```
Goodwine, line 145: This is something I actually wanted to do with my previous PR but for some reason it generated HRX files with wrong encodings 🤔 good thing it worked for you by adding the conditional check! 🙂

lint-spec.ts
```diff
@@ -113,6 +115,37 @@ function getRelativePath(filepath: string): string {
   return p.relative(process.cwd(), filepath);
 }
 
+async function lintNonHrxTestDir(
```
Goodwine, line 118: Please add a brief JSDoc comment for this function, make sure it captures a note about using the `:ignore-for: lint-hrx` option for non UTF-8 files

Goodwine (changes requested):
thanks for looking at this! I just have one suggestion before approving

lint-spec.ts
stof: even better would have been to detect files using BOM or invalid UTF-8 to exclude them automatically. But as we have only a few of them anyway, relying on the ignore rule is simpler for now.

lint-spec.ts
stof: I added that

Goodwine (approved):

Issue #1813, opened 2022-08-13
Update tests for strict function units


nex3:
See https://github.com/sass/sass/issues/3374

- [ ] Deprecation
- [ ] Removal

PR #1830, opened 2022-10-28
Update tests for strict function unit deprecations

nex3:
See #1813
See sass/sass#3374

[skip dart-sass]


spec/core_functions/color/mix.hrx
```diff
@@ -82,7 +82,7 @@ a {
 <===>
 ================================================================================
 <===> unitless_weight/firstwards/input.scss
-a {b: mix(#91e16f, #0144bf, 92)}
+a {b: mix(#91e16f, #0144bf, 92%)}
```
Goodwine, line 85: Per the test this should be unitless

spec/core_functions/color/mix.hrx
```diff
@@ -92,7 +92,7 @@ a {
 <===>
 ================================================================================
 <===> unitless_weight/lastwards/input.scss
-a {b: mix(#91e16f, #0144bf, 43)}
+a {b: mix(#91e16f, #0144bf, 43%)}
```
Goodwine, line 95: ditto

spec/core_functions/color/mix.hrx
```diff
@@ -232,6 +232,48 @@ a {
   b: #85d475;
 }
 
+<===>
+================================================================================
+<===> units/weight/unitless/input.scss
```
Goodwine, line 241: These tests should be moved to the `unitless_weight` dir instead of being under `units`

spec/core_functions/color/mix.hrx
```diff
@@ -349,6 +391,16 @@ Error: argument `$weight` of `mix($color1, $color2, $weight: 50%)` must be a num
 a {b: mix(red, blue, -0.001)}
 
 <===> error/bounds/too_low/error
+DEPRECATION WARNING: $weight: Passing a number without unit % (-0.001) is deprecated.
```
Goodwine, line 406: The deprecation warning was already part of the tests above, is testing both warning + error intentional? If not, then I think it's better to add units so we don't emit both

Goodwine (changes requested):
Can you also add a test that would assert the behavior of a number having multiple numerators and denominators?

spec/core_functions/color/mix.hrx
nex3: I'm just going to remove these and let them be covered by the `unitless` test below.

spec/core_functions/color/mix.hrx
nex3: Yeah, good catch

Goodwine (approved):

PR #1834, opened 2022-11-01
Test invalid contents for importer results

Goodwine:
Test invalid arguments for importer results.
    
- Tests contents not being string values
- Tests SassException argument errors have better output
- Adds sandbox directory to .gitignore

See https://github.com/sass/dart-sass/pull/1816

[skip dart-sass]
[skip sass-embedded]

### DIFF
--- a/js-api-spec/importer.test.ts
+++ b/js-api-spec/importer.test.ts
@@ -675,7 +675,7 @@ describe('when importer does not return string contents', () => {
                 return {
                   // Need to force an invalid type to test bad-type handling.
                   // eslint-disable-next-line @typescript-eslint/no-explicit-any
-                  contents: fs.readFileSync(url.pathname) as any,
+                  contents: fs.readFileSync(url) as any,
                   syntax: 'scss',
                 };
               },
@@ -704,7 +704,7 @@ describe('when importer does not return string contents', () => {
                 return {
                   // Need to force an invalid type to test bad-type handling.
                   // eslint-disable-next-line @typescript-eslint/no-explicit-any
-                  contents: fs.readFileSync(url.pathname) as any,
+                  contents: fs.readFileSync(url) as any,
                   syntax: 'scss',
                 };
               },
@@ -732,7 +732,7 @@ it('throws an ArgumentError when the result sourceMapUrl is missing a scheme', (
             canonicalize: url => dir.url(`dir/_${url}.scss`),
             load: url => {
               return {
-                contents: fs.readFileSync(url.pathname, {encoding: 'utf-8'}),
+                contents: fs.readFileSync(url, {encoding: 'utf-8'}),
                 syntax: 'scss',
                 sourceMapUrl: {},
               };
--- a/js-api-spec/importer.test.ts
+++ b/js-api-spec/importer.test.ts
@@ -685,7 +685,7 @@ describe('when importer does not return string contents', () => {
         });
       }).toThrowSassException({
         line: 0,
-        message:
+        includes:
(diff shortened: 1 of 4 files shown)


js-api-spec/importer.test.ts
```diff
@@ -660,6 +661,93 @@ it(
     })
 );
 
+describe('when importer does not return string contents', () => {
+  it('throws an error in sync mode', () =>
+    sandbox(dir => {
+      dir.write({'dir/_other.scss': '// non empty file'});
```
nex3, line 667: It's probably overkill to create a whole sandbox to read from here when you could just construct and return a `Buffer` in memory.

nex3 (changes requested):

nex3 (approved):

ntkme:
@Goodwine This is failing on dart-sass-embedded and cause the test to completely stuck:
```
FAIL js-api-spec/legacy/importer.test.ts (13.351 s)
  ● when importer returns non-string contents › throws an error in sync mode
    expected exception.toString() to contain "Invalid argument (contents): must be a string but was: Buffer: Instance of 'NativeUint8List'", was "Error: expected "{".
      ╷
    1 │ not a string
      │             ^
      ╵
      other 1:13  @import
      - 1:9       root stylesheet"
      875 |         },
      876 |       });
    > 877 |     }).toThrowLegacyException({
          |        ^
      878 |       line: 1,
      879 |       includes:
      880 |         'Invalid argument (contents): must be a string but was: Buffer: ' +
      at Object.<anonymous> (js-api-spec/legacy/importer.test.ts:877:8)
  ● when importer returns non-string contents › throws an error in async mode
    expected exception.toString() to contain "Invalid argument (contents): must be a string but was: Buffer: Instance of 'NativeUint8List'", was "Error: expected "{".
      ╷
    1 │ not a string
      │             ^
      ╵
      other 1:13  @import
      - 1:9       root stylesheet"
      899 |         expect(() => {
      900 |           throw err;
    > 901 |         }).toThrowLegacyException({
          |            ^
      902 |           line: 1,
      903 |           includes:
      904 |             'Invalid argument (contents): must be a string but was: ' +
      at js-api-spec/legacy/importer.test.ts:901:12
      at ../embedded-host-node/lib/src/legacy/index.ts:56:[16](https://github.com/sass/sass-spec/actions/runs/3391409147/jobs/5636555116#step:9:17)
  ● when importer returns non-string contents › throws an error in async mode
    thrown: "Exceeded timeout of 5000 ms for a test.
    Use jest.setTimeout(newTimeout) to increase the timeout value, if this is a long-running test."
      883 |   });
      884 |
    > 885 |   it('throws an error in async mode', done => {
          |   ^
      886 |     sass.render(
      887 |       {
      888 |         data: '@import "other";',
      at js-api-spec/legacy/importer.test.ts:885:3
      at Object.<anonymous> (js-api-spec/legacy/importer.test.ts:863:1) 
```

ntkme:
I tested locally:

- Dart Sass Embedded would throw an error, but the error message is different. The error message can be different across implementations, so that we should probably remove them from the current test.
- When sourceMapUrl is a non-empty string but not absolute, Dart Sass Embedded does throw an error but `span` is `null`, this should get fixed.

Goodwine:
Yeah, I guess it is not necessary to include all data, I'll send a PR unless there is one already

Goodwine:
Hm.. although it looks like the only reason it's failing is because the value was actually returned as a string but just not valid sass 🤔, maybe it's better to not-return a buffer and return other invalid data?

PR #1835, opened 2022-11-02
Add a regression test for sass/sass#3322

nex3:
[skip dart-sass]


Goodwine (approved):

Goodwine (changes requested):
Ah, this is missing output.css

Goodwine (approved):

PR #1836, opened 2022-11-04
Fix runaway promise

ntkme:
This await is actually needed or the promise returned by `expect()` would run away because the promise is created but the funciton returns `undefined` immmedately after constuct the promise. Or alternatively the curly bracket before `expect` needs to be removed, so that the async function returns a promise that will be waited.


ntkme:
@Goodwine This seems to be changed accidentally in #1834. CI Tests are failing also due to #1834.

Goodwine (approved):

PR #1837, opened 2022-11-05
Update importer 'contents' invalid type tests to bogus object rather …

Goodwine:
…than a Buffer

The Buffer seems to be replaced as a string correctly in dart-sass-embedded, making these tests fail for the wrong reason. Instead of failing for the wrong type it was failing for invalid Sass.

See - https://github.com/sass/sass-spec/pull/1834

### DIFF
--- a/js-api-spec/importer.test.ts
+++ b/js-api-spec/importer.test.ts
@@ -670,7 +670,7 @@ describe('when importer does not return string contents', () => {
               return {
                 // Need to force an invalid type to test bad-type handling.
                 // eslint-disable-next-line @typescript-eslint/no-explicit-any
-                contents: Buffer.from('not a string') as any,
+                contents: {whatever: 123} as any,
                 syntax: 'scss',
               };
             },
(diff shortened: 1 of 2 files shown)


Goodwine:
NVM it seems like dart-sass-embedded just renders it like a toString() 🤔

ntkme:
For embedded-host-node, it only sets a string on protobuf messages, and the type check is currently done on the dart side, which means the type checking is actually done via protobuf:
- Whether a type can be converted to string or not, therefore the printed error message is not the same.
- When a type can be converted to string, protobuf will implicitly convert it as string, and then for URL it will be checked on the dart side, thus again producing a different error message.

Finally there is an issue where if an object is implicitly converted to a string sourceMapUrl, and the sourceMapUrl is not absolute, the dart side do thrown an error but `span` is missing.

Goodwine:
I see where to make the fixes now, I will revert this PR and reopen another with updated tests in a bit

Issue #1832, opened 2022-10-28
Add tests for string.split()


nex3:
See https://github.com/sass/sass/issues/1950

PR #1841, opened 2022-11-16
Add tests for string.split()

dvdherron:
Closes: https://github.com/sass/sass-spec/issues/1832
See: https://github.com/sass/sass/issues/1950

[skip dart-sass]


spec/core_functions/string/split.hrx
```diff
@@ -0,0 +1,179 @@
+<===> options.yml
+---
+:ignore_for:
+- libsass
+
+<===>
+================================================================================
+<===> empty/input.scss
+a {b: split("")}
```
nex3, line 9: This function won't be defined globally, so this should load it from `sass:string`.

nex3, line 9: I think passing only one argument is expected to be an error. Should this by `split("", "")`?

spec/core_functions/string/split.hrx
```diff
@@ -0,0 +1,179 @@
...
+
+<===> limit/output.css
+a {
+  b: ["a", "b", "c, d"];
+}
+
+<===>
+================================================================================
+<===> limit_zero/input.scss
+a {b: split("a, b, c", ", ", 0)}
```
nex3, line 59: This seems like it should just be an error.

spec/core_functions/string/split.hrx
```diff
@@ -0,0 +1,179 @@
...
+a {
+  b: ["👭", "a"];
+}
+
+<===>
+================================================================================
+<===> private_use_character/input.scss
+// Dart Sass emits private-use characters as escapes in expanded mode, but it
+// should still treat them as single characters for the purpose of functions.
+a {b: split("\E000"), ""}
```
nex3, line 83: Early `)`

spec/core_functions/string/split.hrx
```diff
@@ -0,0 +1,179 @@
...
+a {b: split($string: "a/b/c", $separator: "/", $limit: 1)}
+
+<===> named/output.css
+a {
+  b: ["a", "b/c"];
+}
+
+<===>
+================================================================================
+<===> error/too_many_args/input.scss
```
nex3, line 102: What about too few args?

spec/core_functions/string/split.hrx
```diff
@@ -0,0 +1,179 @@
...
+
+<===> separator/output.css
+a {
+  b: ["a", "b", "c"];
+}
+
+<===>
+================================================================================
+<===> empty_separator/input.scss
+a {b: split("Helvetica", "")}
```
nex3, line 29: Probably also a good idea to test an empty string with a non-empty separator

spec/core_functions/string/split.hrx
```diff
@@ -0,0 +1,179 @@
...
+- libsass
+
+<===>
+================================================================================
+<===> empty/input.scss
+a {b: split("")}
+
+<===> empty/output.css
+a {
+  b: [""];
```
nex3, line 13: According to the spec as written, an empty input string should always return an empty array, since `string.length($string)` will be `0`.

nex3 (changes requested):

spec/core_functions/string/split.hrx
dvdherron: These are all loading for `sass:string` now.

dvdherron: Fixed. Changed this to both `$string` and `$separator` being empty and returning an empty array.

spec/core_functions/string/split.hrx
dvdherron: Agreed. A `0` limit now triggers an error.

spec/core_functions/string/split.hrx
dvdherron: fixed

spec/core_functions/string/split.hrx
dvdherron: added a test for too few args too.

spec/core_functions/string/split.hrx
dvdherron: Added a test for an empty string and a non-empty separator.

spec/core_functions/string/split.hrx
dvdherron: fixed this

spec/core_functions/string/split.hrx
```diff
@@ -0,0 +1,225 @@
...
+  b: ["a", "b/c"];
+}
+
+<===>
+================================================================================
+<===> error/too_many_args/input.scss
+@use "sass:string";
+a {b: string.split("a/b/c", "/", 1, 3)}
+
+<===> error/too_many_args/error
```
nex3, line 115: Note for future reference: if you're just working on tests in isolation, you can leave out the contents of error tests since they can easily be generated from the implementation using `--interactive` mode. The most important thing is _that_ the tests error—the specifics of the message are largely implementation details.

nex3 (approved):
Looks good, although I may have a couple more comments once I implement this and run the tests against the implementation.

spec/core_functions/string/split.hrx
```diff
@@ -0,0 +1,225 @@
...
+<===>
+================================================================================
+<===> error/type/separator/input.scss
+@use "sass:string";
+a {b: string.split("1/2/3", /)}
+
+<===> error/type/separator/error
+Error: $separator: / is not a string.
+  ,
+1 | a {b: string.split("1/2/3", /)}
```
nex3, line 166: This is a syntax error, since `/` isn't a valid Sass expression.

nex3 (changes requested):
A couple missing tests I only noticed once I started in on the implementation:

* The function should return a comma separator even for empty and single-element lists.
* The function should accept unquoted strings, and should return an unquoted string if `$string` is unquoted.

There are a couple other small things, but I can auto-update them once I have write access to the PR.

spec/core_functions/string/split.hrx
dvdherron: Fixed. Changed this separator to a number.

dvdherron:
Added new tests for unquoted strings and checking the separator type for empty and single-element lists in https://github.com/sass/sass-spec/pull/1841/commits/175359d2c7b8cd7f96c8e63ad9a635bc4af73442. I think this covers the changes requested.

nex3 (approved):

nex3:
Test failures should be fixed when https://github.com/sass/embedded-host-node/pull/175 lands.

PR #1842, opened 2022-11-23
Add tests for color.is-legacy()

SondraE:
Part of tests for color spaces. 


nex3:
We should target these color space PRs to separate branch (`color-4` would match the branch name in the Dart Sass repo). If we land them on `main`, it'll break our CI since the actual implementation of the behavior in question isn't finished yet.

dvdherron:
> We should target these color space PRs to separate branch (`color-4` would match the branch name in the Dart Sass repo). If we land them on `main`, it'll break our CI since the actual implementation of the behavior in question isn't finished yet.

@nex3 I don't think we have the right access here to make a new branch. If you create the branch I can re-target these color PRs.

nex3:
@dvdherron Done :smiley: 

@SondraE You should be able to fix the lint errors by running `npm run lint-spec -- --fix`. I'm not sure why the Dart Sass tests are running, but they may go away if you run the tests again now that you have `[skip dart-sass]`.

spec/core_functions/color/is_legacy.hrx
```diff
@@ -0,0 +1,225 @@
...
+a {b: color.is-legacy(oklch(70% 0.1 200))}
+
+<===> oklch/output.css
+a {
+  b: false;
+}
+
+<===>
+================================================================================
+<===> error/too_few_args/input.scss
```
nex3, line 217: These may not need explicit outputs, but they do need inputs to test against :sweat_smile:

nex3 (changes requested):

spec/core_functions/color/is_legacy.hrx
```diff
@@ -0,0 +1,237 @@
...
+@use "sass:color";
+a {b: color.is-legacy(rgb(0 255))}
```
nex3, line 219: This is passing too few args to `rgb()`, but it should be passing too few args to `is-legacy()`

spec/core_functions/color/is_legacy.hrx
```diff
@@ -0,0 +1,237 @@
...
+<===> hex/output.css
+a {
+  b: true;
+}
+
+<===>
+================================================================================
+<===> named/input.scss
+@use "sass:color";
+a {b: color.is-legacy($color: midnightblue)}
```
nex3, line 65: It looks like this is testing both a named argument _and_ a named color—we should have separate tests for each of these cases (see https://github.com/sass/sass-spec/blob/main/STYLE_GUIDE.md#do-test-only-one-thing-per-spec)

nex3 (changes requested):

nex3:
Removing `[skip dart-sass]` since I don't think this will require any code changes in Dart Sass

nex3:
JS API test failures are unrelated, I'm working on fixing them in https://github.com/sass/sass-spec/pull/1853

nex3 (approved):

PR #1847, opened 2022-12-08
Fix dart-sdk 3.0

ntkme:
```
/tmp/dart-sass-spec:2:1: Error: A library can't opt out of null safety by default, when using sound null safety.
// @dart=2.9
^^^^^^^^^^^^
```


ntkme:
@nex3 We need this fix or dart 3.0 dev channel will fail.

PR #1848, opened 2022-12-08
Drop the non-nullability annotation from the Dart Sass wrapper

nex3:
It seems like we may not need this anymore, and it's breaking the Dart
Dev CI.

### DIFF
--- a/lib-js/compiler.ts
+++ b/lib-js/compiler.ts
@@ -116,7 +116,6 @@ export class DartCompiler implements Compiler {
       throw new Error(`${repoPath} is not a valid Dart Sass repository`);
     }
     const dartFile = `
-// @dart=2.9
 import "dart:convert";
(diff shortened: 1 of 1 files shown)


jathak (approved):

ntkme:
@nex3 I will close my PR, but there is one more annotation at this line: https://github.com/sass/sass-spec/pull/1847/files#diff-050b6130eef520999ce5ee12633ca266033054db303cce6f1933311d2c83d9f0

nex3:
Ah sorry @ntkme, I didn't notice you had a PR for this. The Ruby file is essentially unused at this point and we should probably just delete that runner entirely.

ntkme:
@nex3 No worries. I think in this case we can just rerun failed CI for release without needing to cut another one.

PR #1850, opened 2022-12-09
[Color 4] Update tests for color.grayscale()

dvdherron:
Part of updated/new tests for color spaces.


spec/core_functions/color/grayscale.hrx
```diff
@@ -22,6 +22,11 @@ a {
 a {b: grayscale(#494949)}
 
 <===> no_saturation/gray/output.css
+a {
+  b: rgb(73, 73, 73);
```
dvdherron, line 26: @nex3 You mentioned that we should typically not return a color in a different space than the color's origin space. 

Should the proposal for `grayscale` be updated to say that legacy colors should be converted back to `origin`? [Currently](https://github.com/sass/sass/blob/colors-2/proposal/color-4-new-spaces.md#colorgrayscale) it says to return legacy colors in `hsl`.

nex3: Yes, I think it should.

nex3:
Removing `[skip dart-sass]` since this doesn't have an accompanying Dart Sass PR

PR #1851, opened 2022-12-09
Add tests for color.space()

SondraE:
Part of tests for color spaces.


nex3:
Removing `[skip dart-sass]` since this doesn't have an accompanying Dart Sass PR

spec/core_functions/color/space.hrx
```diff
@@ -165,7 +165,7 @@ a {b: color.space(color(xyz-d65 0.21661 0.14602 0.59452))}
 
 <===> xyz-d65/output.css
 a {
-  b: xyz-d65;
+  b: xyz;
```
nex3, line 168: Note: `xyz-d65` is just an alias for `xyz`, so that's the space we return here.

nex3 (approved):

PR #1858, opened 2022-12-16
Rename `fromPath()` to `fromRoot()`

nex3:
This clarifies that this should only be used for the root directory of
a test suite, since it won't consider any options that come from
parent directories. This also refactors lint-spec.ts to avoid using
this function for subdirectories.

### DIFF
--- a/lint-spec.ts
+++ b/lint-spec.ts
@@ -9,20 +9,20 @@ import RealDirectory from './lib/spec-directory/real-directory';
 import VirtualDirectory from './lib/spec-directory/virtual-directory';
 
 async function lintAllTests(fix: boolean) {
-  //try {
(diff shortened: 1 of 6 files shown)


lint-spec.ts
```diff
@@ -4,25 +4,25 @@ import p from 'path';
...
-    await lintDirectory(rootDir, reporter, {canBeHrxRoot: true, fix});
-    await lintHrxSize(rootPath, reporter, {fix});
-
-    reporter.writeSummary();
-    process.exitCode = reporter.exitCode();
-  } catch (error) {
-    console.log(`${error}`);
-    process.exitCode = 255;
-  }
+  //try {
```
Goodwine, line 12: Remove?

Goodwine (approved):

PR #1859, opened 2022-12-17
Add `SpecDirectory.parent()`

nex3:
This allows us to stop eagerly fetching options to pass to subdirectories,
which helps ensure correctness when parent directory options are modified.
Having access to the parent directory also makes more complex and powerful
automated refactors possible.

### DIFF
--- a/lib/spec-directory/virtual-directory.ts
+++ b/lib/spec-directory/virtual-directory.ts
@@ -91,17 +91,12 @@ export default class VirtualDirectory extends SpecDirectory {
   }
 
   /**
-   * Create a virtual directory from string contents, and an optional path.
-   * If no path is given (e.g. in testing), it is set to an empty string.
+   * Create a virtual directory from string contents, for testing purposes.
    */
-  static async fromContents(
-    contents: string,
-    path = ''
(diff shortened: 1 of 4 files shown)


lib/spec-directory/virtual-directory.ts
```diff
@@ -33,24 +33,28 @@ function createSubdirCache(dir: HrxDirectory): Record<string, HrxDirectory> {
 export default class VirtualDirectory extends SpecDirectory {
   path: string;
   basePath: string;
+
+  // The directory that contains this, or `null` if this is the root directory.
```
Goodwine, line 37: Nit: In order for these comments to be JsDoc they need the format `/** */`, I noticed the rest of the properties use line comments, should all of those be updated here?

Goodwine (approved):

lib/spec-directory/virtual-directory.ts
nex3: I think the idea was that since these are private members, they don't need JSDoc comments—although that may not bear up with how VSCode or whatever handles them. (Note that this code was mostly written by a contractor so I'm not 100% sure of the reasoning behind it.) I'll update them.

PR #1862, opened 2022-12-27
Add tests for removing sourcemap comments from Sass sources

Goodwine:
[skip dart-sass]

See: https://github.com/sass/dart-sass/pull/1860


spec/css/comment.hrx
```diff
@@ -69,7 +69,8 @@ should include their own tests for CR LF newlines.
 <===>
 ================================================================================
 <===> converts_newlines/scss/cr/input.scss
-/* foo * bar */
+/* foo
+ * bar */
```
nex3, line 73: It looks like your editor may have accidentally changed a CR into an LF here.

spec/css/comment.hrx
```diff
@@ -161,6 +163,55 @@ a {
...
+<===> sourcemap/options.yml
+:ignore_for:
+  - libsass
+
+<===>
+================================================================================
+<===> sourcemap/sourcemappingurl/input.scss
+a {
+  b: c;
+}
```
nex3, line 177: Nit: https://github.com/sass/sass-spec/blob/main/STYLE_GUIDE.md#prefer-one-line-rules-when-they-contain-a-single-child-with-no-children

spec/css/comment.hrx
```diff
@@ -161,6 +163,55 @@ a {
...
+/*# sourceMappingURL=whatever */
+
+<===> sourcemap/sourcemappingurl/output.css
+a {
+  b: c;
+}
+
+<===>
+================================================================================
+<===> sourcemap/between_imports/input.scss
```
nex3, line 187: "import" is kind of a misleading term here, since this doesn't actually use `@import`. We usually use "load" to refer to "any at-rule that loads another file".

nex3 (changes requested):

spec/css/comment.hrx
Goodwine: oh, didn't realize, fixed

nex3 (approved):

PR #1866, opened 2023-01-17
Add Tests for color.is-in-gamut()

SondraE:
Part of new/updated tests for color spaces.

See: https://github.com/sass/sass/issues/2831


nex3:
I think we should be able to get these tests passing if we merge `main` into `feature.color-4`. In future, it's probably a good idea to avoid merging directly from `main` into one of these PRs, since the feature branch won't always be up-to-date.

jerivas:
@nex3 @SondraE I cleaned up the commit history to remove the commits from `main`. According to my local diffing the end result should be the same as before, let me know if that's not the case

PR #1872, opened 2023-01-20
add spec for trimming super selectors in @extend (pt. 2)

### LINKED ISSUE #1678: Additional test case for `@extend`

connorskees:
revives #1680, closes #1678


connorskees:
It seems trimming has changed since the original PR was opened. This test case will now generate

```css
a b, a > b {
   color: red;
 }
```

Should the expected output be modified, or would this be considered an issue with `dart-sass`? The `a > b` seems redundant.

nex3:
That does look like a bug in Dart Sass. Can you file an issue there and mark this spec as TODO for now?

nex3 (approved):

Issue #1679, opened 2021-07-30
Ability to ignore error messages/spans and deprecation warnings


connorskees:
Error messages/spans and deprecation warnings create a lot of noise when testing against the spec, but aren't necessarily useful to test against. Especially warnings like `DEPRECATION WARNING: Using / for division is deprecated and will be removed in Dart Sass 2.0.0.`, which is the cause for 50+ errors, but can never reasonably be fixed since the message contains the string "Dart Sass"

nex3:
This is why we have the ability to define different outputs for different implementations. Maybe it's time to add Rust Sass to this repo as an officially-tested implementation?

nex3:
Closing this out for now, let me know if you want to start the process of adding Rust Sass to this repo.

connorskees:
Would having to support a Rust implementation of Sass increase the maintenance burden on this repository?

In my local development, I maintain a forked version of this repository with `trimErrors` and `skipWarnings` set to always be `true`. I think the ability to pass a flag that would set these two values would be sufficient for my use case. 

https://github.com/sass/sass-spec/blob/a5d4b0ce5f8f03312f2f67cc36960ecc94069110/lib/test-case/test-case.ts#L124-L126

I can upstream my changes at some point in the near future if this change makes sense.

nex3:
Tracking Rust Sass would be a bit more effort, but it's not onerous. sass-spec is designed to track multiple implementations.

If you'd prefer not to, I'd also accept a PR to add `--ignore-error-diffs` and `--ignore-warning-diffs` flags to the CLI.

nex3:
@connorskees What are your thoughts here?

PR #1888, opened 2023-03-13
add new flags for comparing warning and error diffs

connorskees:
Adds 3 new CLI flags to configure how warning and error diffs are handled. 

Closes https://github.com/sass/sass-spec/issues/1679

### DIFF
--- a/lib/test-case/test-case.ts
+++ b/lib/test-case/test-case.ts
@@ -32,7 +32,7 @@ export default class TestCase {
     impl: string,
     compiler: Compiler,
     todoMode: TodoMode,
-    private compareOpts?: CompareOptions,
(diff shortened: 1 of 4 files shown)


lib/test-case/test-case.ts
```diff
@@ -185,7 +188,7 @@ export default class TestCase {
     // create a warning file
     if (
       this.dir.hasFile('warning') &&
-      this.dir.readFile('warning') &&
+      await this.dir.readFile('warning') &&
```
connorskees, line 191: unrelated to my changes, but was giving an error-level lint. `readFile` returns a promise, which if not awaited will always be true in this context

nex3 (approved):

PR #1890, opened 2023-03-23
Add tests for calculation constants

### LINKED ISSUE #1885: Add tests for calc constants
See https://github.com/sass/sass/issues/3258

nex3:
Closes #1885
See sass/sass#3258

[skip dart-sass]

### DIFF
--- a/spec/values/numbers/degenerate.hrx
+++ b/spec/values/numbers/degenerate.hrx
@@ -1,0 +1,189 @@
+<===> infinity/unitless/input.scss
+@use 'sass:math';
+a {b: math.div(1, 0)}
+
+<===> infinity/unitless/output.css
+a {
+  b: calc(infinity);
+}
+
+<===>
+================================================================================
+<===> infinity/single_unit/input.scss
+@use 'sass:math';
+a {b: math.div(1px, 0)}
+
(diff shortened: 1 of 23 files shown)


spec/values/numbers/degenerate.hrx
```diff
@@ -0,0 +1,195 @@
...
+  input.scss 2:7  root stylesheet
+
+<===>
+================================================================================
+<===> error/nan/multiple_numerator_units/input.scss
+@use 'sass:math';
+a {b: math.div(0px * 0em, 0)}
+
+<===> error/nan/multiple_numerator_units/error
+Error: calc(NaNpx*em) isn't a valid CSS value.
```
Goodwine, line 162: Nit: I think that for `NaN` and `infinity`, having the units adjacent to the value makes the error difficult to read for me. I wonder if it would be possible to have a space in between them only for the cases of `NaN` and `infitnity`. 🤔

Goodwine (approved):

spec/values/numbers/degenerate.hrx
nex3: I think a better solution here would be to have a more principled behavior for representing multi-unit values in general using calc, but that's outside the scope of the current change.

PR #1892, opened 2023-03-27
Add specs for color functions that can be either Sass or CSS native functions

Goodwine:
[skip dart-sass]

Issue: https://github.com/sass/sass/issues/3507
See: https://github.com/sass/dart-sass/pull/1926


spec/core_functions/color/alpha.hrx
```diff
@@ -1,3 +1,9 @@
+<===> options.yml
+':ignore_for':
+  - libsass
```
nex3, line 3: Ideally we wouldn't ignore the tests that are already passing for libsass.

spec/core_functions/color/alpha.hrx
```diff
@@ -96,6 +102,16 @@ a {
   b: opacity(10%);
 }
 
+<===>
+================================================================================
+<===> opacity/with_css_var/input.scss
+a {b: opacity(var(--c))}
```
nex3, line 108: Also test other special number cases (a calculation or an unquoted string that begins with `calc(`)

spec/core_functions/color/hsla.hrx
```diff
@@ -56,3 +50,89 @@ a {
 a {
   b: rgba(51, 204, 204, 0.4);
 }
+
+<===>
+================================================================================
+<===> with_css_var/first_arg/input.scss
+a {b: hsla(var(--a), 25%, 12%, 0.4)}
```
nex3, line 57: We already test this for `hsl()`, so we don't need to test it for `hsla()` as well. See https://github.com/sass/sass-spec/blob/4016da98530de32c2783545a2fc61e23ba70f4f7/spec/core_functions/color/hsla.hrx#L1-L3.

nex3 (changes requested):

spec/core_functions/color/hsla.hrx
Goodwine: gotcha, I totally missed the `hsl/` directory

spec/core_functions/color/alpha.hrx
```diff
@@ -96,6 +96,62 @@ a {
...
+  - libsass
+
+<===> opacity/with_clamp/input.scss
+a {b: opacity(clamp(1rem, 2.5vw, 2rem))}
+
+<===> opacity/with_clamp/output.css
+a {
+  b: opacity(clamp(1rem, 2.5vw, 2rem));
+}
+
```
nex3, line 154: We probably don't need to test every different type of calculation, especially since we'll probably be adding a bunch of new ones shortly.

nex3 (approved):

PR #1898, opened 2023-04-11
Don't separate out media queries after one has bubbled

nex3:
See sass/dart-sass#777

[skip dart-sass]

### DIFF
--- a/spec/css/media/bubbling.hrx
+++ b/spec/css/media/bubbling.hrx
@@ -23,3 +23,31 @@
     }
   }
 }
+
+<===>
+================================================================================
+<===> preserve_merge_after_bubble/input.scss
+// Regression test for sass/dart-sass#777
+@media (a: b) {
+  @media (c: d) {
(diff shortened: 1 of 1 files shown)


spec/css/media/bubbling.hrx
```diff
@@ -23,3 +23,31 @@
     }
   }
 }
+
+<===>
+================================================================================
+<===> preserve_merge_after_bubble/input.scss
```
Goodwine, line 29: Doesn't the change for `equalsIgnoringChildren` also affect other nodes? I wonder if it would make sense to add tests for other "container" nodes where bubbling may be affected by the change 🤔

nex3: I'm not sure there are actually other situations that trigger this behavior. Nested selectors always get pushed below declarations, and other at-rules aren't merged with their parent and so just end up normally nested rather than bubbled.

PR #1900, opened 2023-04-17
Add specs for :has(+ &)

nex3:
See sass/dart-sass#1937

[skip dart-sass]


tbondwilkinson (commented):
Do other specs now cover the :is() selector? Or are the :is() specs no longer needed?

nex3:
I had assumed there was also an `is.hrx`, but it looks like it was just misnamed. I've added it back but renamed it appropriately.

tbondwilkinson (approved):

PR #1902, opened 2023-04-21
Add support for running JS Spec Tests in browser context.

jgerigmeyer:
TODO:

- [x] @nex3 confirm overall approach
- [x] fix sourceMap errors
- [x] [compile.test.ts](https://github.com/sass/sass-spec/tree/main/js-api-spec/compile.test.ts)
- [x] [value/](https://github.com/sass/sass-spec/tree/main/js-api-spec/value)
- [x] [function.test.ts](https://github.com/sass/sass-spec/blob/main/js-api-spec/function.test.ts)
- [x] [importer.test.ts](https://github.com/sass/sass-spec/blob/main/js-api-spec/importer.test.ts)
- [x] [logger.test.ts](https://github.com/sass/sass-spec/blob/main/js-api-spec/logger.test.ts)

[skip dart-sass]


js-api-spec/browser/compile.test.ts
```diff
@@ -0,0 +1,139 @@
...
+          sourceMap: true,
+          sourceMapIncludeSources: true,
+        });
+        expect(result).toHaveProperty('sourceMap');
+
+        const sourceMap = result.sourceMap!;
+        expect(sourceMap).toHaveProperty('sourcesContent');
+        expect(sourceMap.sourcesContent!).toBeArray();
+        expect(sourceMap.sourcesContent!.length).toBeGreaterThanOrEqual(1);
+      });
```
jgerigmeyer, line 71: Of the tests I've ported over so far, these are the only that are [failing](https://github.com/sass/sass-spec/actions/runs/4767106037/jobs/8474966139?pr=1902):

```
dartException: TypeError: Instance of 'MappedListIterable<String, dynamic>': type 'MappedListIterable<String, dynamic>' is not a subtype of type 'Iterable<dynamic>'
```

@jerivas Do we want to support sourcemaps in the browser context? If so, could you add some tests for that in our `browser_test.dart` file?

.github/workflows/ci.yml
```diff
@@ -151,8 +151,46 @@ jobs:
...
+      - uses: browser-actions/setup-chrome@v1
+      - run: npm install
+      - uses: dart-lang/setup-dart@v1.2
+        with: {sdk: stable}
+      - uses: browser-actions/setup-chrome@v1
+
+      - name: Install Dart Sass
+        run: |
+          # TODO update this to use https://github.com/sass/dart-sass.git
+          git clone https://github.com/oddbird/dart-sass.git ../dart-sass --depth 1 --branch browser
```
jgerigmeyer, line 173: Before merging, this should be updated:

```suggestion
          git clone https://github.com/sass/dart-sass.git ../dart-sass --depth 1
```

js-api-spec.ts
```diff
@@ -54,26 +59,18 @@ del.sync(dir, {force: true});
...
-
 // We want to use the type information from --sassSassRepo even if --sassPackage
 // is written in TypeScript, because the specs may test behavior that's not yet
 // implemented in --sassPackage and we don't want that to cause a compile error.
 // We accomplish this by creating a JavaScript package named "sass" that
 // requires and re-exports --sassPackage, and using the type annotations from
 // --sassSassRepo as that package's annotations.
-const packageRequire = JSON.stringify(p.resolve(argv.sassPackage));
+const packageRequire = argv.browser
+  ? p.resolve(argv.sassPackage, 'sass.default.js')
```
jgerigmeyer, line 69: In the browser context, we want to directly import the compiled browser version.

jgerigmeyer (commented):
@nex3 This is still WIP, but it'd be good to have your initial review on the overall approach here, and how the tests and setup files are re-arranged a bit.

nex3 (commented):
Approach looks good to me!

js-api-spec/browser/compile.test.ts
jgerigmeyer: @nex3 I [added tests in Dart Sass](https://github.com/oddbird/dart-sass/commit/773ed19c32f54fc8b329c1d433fa99cb331df6f6) for sourceMaps, and didn't hit any errors. Any ideas why this test is still throwing an error when run in a browser context? It looks like the error starts with the call to `jsify(sourceMap)` in `node/compile.dart` here: https://github.com/sass/dart-sass/pull/1895/files#diff-5da2f39b0add2ab3649136ea6b10ff7c87690d2f895539c10511e5425c224024R164

jgerigmeyer: @nex3 Digging into this a bit more, I'm thinking it might be a bug in the dart2js `jsify` implementation, specifically [where `_convertDataTree` maps an `Iterable`](https://github.com/dart-lang/sdk/blob/main/sdk/lib/_internal/js_runtime/lib/js_patch.dart#L157). If I change that line from:

```dart
convertedList.addAll(o.map(_convert));
```

to:

```dart
convertedList.addAll(o.map(_convert).toList());
```

then the error goes away and the tests seem to pass in both contexts.

Should I file an issue?

jgerigmeyer:
@nex3 Looking at the remainder of the JS API spec tests, I'd like your feedback on which ones we should be supporting in a browser context. Here are my initial thoughts/questions:

- [legacy/](https://github.com/sass/sass-spec/tree/main/js-api-spec/legacy) -- I assume we can skip these
- [value/](https://github.com/sass/sass-spec/tree/main/js-api-spec/value) -- Do we want to run these? If so, we'll need to expose the various types (e.g. SassArgumentList, SassList, sassNull, etc) for the browser build.
- [function.test.ts](https://github.com/sass/sass-spec/blob/main/js-api-spec/function.test.ts) -- Similar to `value/`; if we want to run these, we'll need to expose CustomFunction, SassString, and sassNull.
- [importer.test.ts](https://github.com/sass/sass-spec/blob/main/js-api-spec/importer.test.ts) -- Do we want to run (some of) these? I assume we'd skip the ones that rely on FS access (via `sandbox()`), but it's possible we could run some of the tests.
- [logger.test.ts](https://github.com/sass/sass-spec/blob/main/js-api-spec/logger.test.ts) -- I assume we want some logging/debug support in the browser build, but given the differences between `console` and `stdio`, it might make sense to right a subset of logging tests specifically for the browser context, and not copy any of these over to run in both contexts?

js-api-spec/browser/compile.test.ts
nex3: Yeah, I think filing that issue is a good idea. You could even send them a pull request for it.

nex3:
> @nex3 Looking at the remainder of the JS API spec tests, I'd like your feedback on which ones we should be supporting in a browser context. Here are my initial thoughts/questions:
> 
> * [legacy/](https://github.com/sass/sass-spec/tree/main/js-api-spec/legacy) -- I assume we can skip these

Yep, totally ignore these.

> * [value/](https://github.com/sass/sass-spec/tree/main/js-api-spec/value) -- Do we want to run these? If so, we'll need to expose the various types (e.g. SassArgumentList, SassList, sassNull, etc) for the browser build.
> * [function.test.ts](https://github.com/sass/sass-spec/blob/main/js-api-spec/function.test.ts) -- Similar to `value/`; if we want to run these, we'll need to expose CustomFunction, SassString, and sassNull.


We should definitely support these. We want people to be able to use custom functions on the browser.

> * [importer.test.ts](https://github.com/sass/sass-spec/blob/main/js-api-spec/importer.test.ts) -- Do we want to run (some of) these? I assume we'd skip the ones that rely on FS access (via `sandbox()`), but it's possible we could run some of the tests.

Yeah, I think we should separate out the ones that test importer/filesystem interactions into a separate file and run the rest on the browser.

> * [logger.test.ts](https://github.com/sass/sass-spec/blob/main/js-api-spec/logger.test.ts) -- I assume we want some logging/debug support in the browser build, but given the differences between `console` and `stdio`, it might make sense to right a subset of logging tests specifically for the browser context, and not copy any of these over to run in both contexts?

We could potentially abstract `captureStdio()` into a function that captures stdio on Node and `console.log()` on the browser. If that's infeasible, then maybe we should have three test files under a `logger/` directory: Node-specific, browser-specific, and platform-agnostic.

js-api-spec/browser/compile.test.ts
jgerigmeyer: Filed an [issue](https://github.com/dart-lang/sdk/issues/52200), and skipped these tests in browser context temporarily. (I could submit a PR, but got bogged down trying to get Dart SDK tests to run locally, so decided to wait on that for now.)

jgerigmeyer: Resolved in https://github.com/oddbird/sass-spec/commit/9cca6820c579d66cbfa60788ce951c9fce89f574

js-api-spec/setup.ts
```diff
@@ -0,0 +1,387 @@
...
+        },
+      }),
+    });
+    jasmine.addAsyncMatchers({
+      toThrowSassException: () => ({
+        compare: toThrowSassException,
+      }),
+    });
+  });
+}
```
jgerigmeyer, line 311: I think one of the main downsides of this current tooling is that while Jest and Jasmine have an almost identical API, some of the matchers are not 100% the same. Similar to the [default](https://jestjs.io/docs/expect) and [extended](https://github.com/jest-community/jest-extended) Jest matchers, this uses the [default](https://jasmine.github.io/api/4.6/matchers.html) and [extended](https://github.com/JamieMason/Jasmine-Matchers) Jasmine matchers -- but there are still occasional divergences.

karma.config.js
```diff
@@ -0,0 +1,62 @@
+const {
+  NodeModulesPolyfillPlugin,
+} = require('@esbuild-plugins/node-modules-polyfill');
```
jgerigmeyer, line 3: This is only needed for the `util.inspect` usage in `setup.ts`.

jgerigmeyer (commented):
@nex3 I think this is ready for another (close to final?) round of review!

jerivas (approved):
LGTM!

jgerigmeyer:
@nex3 I think the CI failures here are unrelated to this PR, but are a result of the recent sass-embedded changes (and the need to run `dart run grinder protobuf`).

jgerigmeyer:
[Dart Sass running as browser JS](https://trello.com/c/pJb7GFdE/12-dart-sass-running-as-browser-js)

js-api-spec/compile.browser.test.ts
```diff
@@ -0,0 +1,137 @@
+// Copyright 2021 Google Inc. Use of this source code is governed by an
```
nex3, line 1:
```suggestion
// Copyright 2023 Google Inc. Use of this source code is governed by an
```

Same with other new files.

js-api-spec/setup.ts
nex3: Would it be possible to just use Jasmine in _all_ the tests?

karma.config.js
nex3: Can you add that as a comment in the file?

nex3 (approved):

js-api-spec/setup.ts
jgerigmeyer: @nex3 In https://github.com/sass/sass-spec/pull/1902/files/7219fb7ce424a9179c46c17d744f00859fa92b6d..ba3a0e2c6003367391d8dc545986e9e251f0e5a9 I converted all of the JS API tests to use Jasmine.

What do you think? (I could convert the `test/` directory as well, but since there's no overlap between `test/` and `js-api-spec/` I figured it didn't hurt to leave those running via Jest.)

nex3 (approved):

jgerigmeyer:
@nex3 Full transparency, I just realized yesterday that Karma is [officially deprecated](https://github.com/karma-runner/karma#karma-is-deprecated-and-is-not-accepting-new-features-or-general-bug-fixes) as of a few weeks ago. 🤦

I think there's potential for switching to either [Web Test Runner](https://modern-web.dev/docs/test-runner/overview/) or [jasmine-browser-runner](https://github.com/jasmine/jasmine-browser-runner), but after trying them out a bit today I ran out of steam. I'd be happy to chat more about pros/cons of those two options, but I think I'd recommend that this is a good start, and we can consider a switch to Web Test Runner sooner or later.

PR #1910, opened 2023-05-17
JS API tests for loading multiple nested paths with renderSync

Goodwine:
Issue: https://github.com/sass/dart-sass/issues/1962

[skip sass-embedded]

See: https://github.com/sass/embedded-host-node/pull/226

### DIFF
--- a/js-api-spec/legacy/importer.node.test.ts
+++ b/js-api-spec/legacy/importer.node.test.ts
@@ -941,14 +941,14 @@ it('compiles multiple nested relative imports and incorrect load path with rende
   await sandbox(dir => {
     dir.write({
       'src/index.scss': `
-        @import "./foo/_a.scss";
+        @import "./_a.scss";
         @import "./_includes.scss";
       `,
-      'include/foo/_a.scss': '/* A */',
-      'include/foo/_b.scss': '/* B */',
+      'include/_a.scss': '/* A */',
+      'include/_b.scss': '/* B */',
       'src/_includes.scss': `
-        @import "./foo/_a.scss";
-        @import "./foo/_b.scss";
+        @import "./_a.scss";
+        @import "./_b.scss";
       `,
     });
 
--- a/js-api-spec/legacy/importer.node.test.ts
+++ b/js-api-spec/legacy/importer.node.test.ts
@@ -937,7 +937,8 @@ describe('when importer returns non-string contents', () => {
   });
 });
 
-it('compiles multiple nested relative imports and incorrect load path with renderSync', async () => {
+// Regression test for sass/dart-sass#1962
+it('compiles multiple nested relative imports loaded multiple times across different files', async () => {
   await sandbox(dir => {
     dir.write({
       'src/index.scss': `
--- a/js-api-spec/legacy/importer.node.test.ts
+++ b/js-api-spec/legacy/importer.node.test.ts
@@ -937,8 +937,8 @@ describe('when importer returns non-string contents', () => {
   });
 });
 
-it('compiles multiple nested relative imports and incorrect load path with renderSync', () => {
-  sandbox(dir => {
+it('compiles multiple nested relative imports and incorrect load path with renderSync', async () => {
+  await sandbox(dir => {
     dir.write({
       'src/index.scss': `
         @import "./foo/_a.scss";
--- a/js-api-spec/legacy/importer.node.test.ts
+++ b/js-api-spec/legacy/importer.node.test.ts
@@ -965,6 +965,6 @@ it('compiles multiple nested relative imports loaded multiple times across diffe
         return null;
       },
     });
-    expect(result.css.toString()).toBe('/* A */\n/* A */ /* B */');
+    expect(result.css.toString()).toBe('/* A */\n/* A */\n/* B */');
   });
 });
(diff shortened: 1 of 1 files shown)


js-api-spec/importer.test.ts
```diff
@@ -729,6 +737,37 @@ it('throws an ArgumentError when the result sourceMapUrl is missing a scheme', (
...
+      `,
+      'include/foo/_a.scss': '/* A */',
+      'include/foo/_b.scss': '/* B */',
+      'src/_includes.scss': `
+        @import "./foo/_a.scss";
+        @import "./foo/_b.scss";
+      `,
+    });
+
+    const result = renderSync({
```
ntkme, line 755: Legacy API test should be put `js-api-spec/legacy/importer.test.ts` instead.

ntkme (changes requested):

js-api-spec/importer.test.ts
Goodwine: no wonder I didn't see other `renderSync` heh

js-api-spec/legacy/importer.test.ts
```diff
@@ -924,3 +925,34 @@ describe('when importer returns non-string contents', () => {
     );
   });
 });
+
+it('compiles multiple nested relative imports and incorrect load path with renderSync', () => {
+  sandbox(dir => {
+    dir.write({
```
Goodwine, line 931: removing any of these files make the test pass

js-api-spec/legacy/importer.test.ts
```diff
@@ -924,3 +925,34 @@ describe('when importer returns non-string contents', () => {
...
+      'src/index.scss': `
+        @import "./foo/_a.scss";
+        @import "./_includes.scss";
+      `,
+      'include/foo/_a.scss': '/* A */',
+      'include/foo/_b.scss': '/* B */',
+      'src/_includes.scss': `
```
Goodwine, line 938: I kinda feel the error is sorta WAI because this file is not inside `./includes`

js-api-spec/legacy/importer.test.ts
```diff
@@ -924,3 +925,34 @@ describe('when importer returns non-string contents', () => {
...
+        @import "./foo/_a.scss";
+        @import "./foo/_b.scss";
+      `,
+    });
+
+    const result = sass.renderSync({
+      file: dir('src/index.scss'),
+      importer: function (url) {
+        const path = dir('./include', url);
```
Goodwine, line 947: This IIUC normalizes the path so it fetches for files under `./include`

but because `includes.scss` is loaded from a different directory then I think this messes up with `renderSync`

And I wonder if it's worth fixing it or whether this can be considered to be a reasonable error because of the use of a deprecated API 🤔

nex3:
What parameters are actually passed to the importer in plain Dart Sass versus the embedded host?

Goodwine:
> What parameters are actually passed to the importer in plain Dart Sass versus the embedded host?

I'm not 100% sure if this is what you're referring to, but:

### dart-sass importer
```
console.log
  ## DEBUG: url= ./foo/_a.scss - prev: /home/goodwine/repos/sass-spec/spec/sandbox/9067981249289676/src/index.scss
console.log
  ## DEBUG: url= ./foo/_a.scss - prev: /home/goodwine/repos/sass-spec/spec/sandbox/9067981249289676/src/_includes.scss
console.log
  ## DEBUG: url= ./foo/_b.scss - prev: /home/goodwine/repos/sass-spec/spec/sandbox/9067981249289676/src/_includes.scss
```

### embedded importer
```
console.log
  ## DEBUG: url= foo/_a.scss - prev: /home/goodwine/repos/sass-spec/spec/sandbox/308212682427188/src/index.scss
console.log
  ## DEBUG: url= file:///home/goodwine/repos/sass-spec/spec/sandbox/308212682427188/src/foo/_b.scss - prev: /home/goodwine/repos/sass-spec/spec/sandbox/308212682427188/src/_includes.scss
<crash>
```

I looked at the implementations, and dart-sass seems to just call the importer without trying to do much wrapping, but embedded-host-node has some fancier wrapping and I think that's where it goes "wrong"

Like if I implement a "modern" importer using `compile()` instead of `renderSync()` I would have to implement a correct `canonalize()` and I would realize that the input Sass is kinda wrong 🤔

js-api-spec/legacy/importer.node.test.ts
```diff
@@ -935,3 +936,35 @@ describe('when importer returns non-string contents', () => {
...
+        const path = dir('./include', url);
+        if (fs.existsSync(path)) {
+          return {
+            contents: fs.readFileSync(path, 'utf8'),
+          };
+        }
+        return null;
+      },
+    });
+    expect(result.css.toString()).toBe('/* A */\n/* A */ /* B */');
```
nex3, line 968: Note: once https://github.com/sass/dart-sass/pull/1989 lands, this will (correctly) become

```suggestion
    expect(result.css.toString()).toBe('/* A */\n/* A */\n/* B */');
```

nex3 (approved):

js-api-spec/legacy/importer.node.test.ts
Goodwine: Even after syncing past that commit I don't see a change in behavior (i.e. whitespace instead of `\n` is still working)

Goodwine: nvm I see this problem happening now :)

PR #1911, opened 2023-05-18
Add instructions about building dart-sass before running tests

Goodwine:


Goodwine:
nvm I think this is not entirely necessary (although synchronize may be)

PR #1912, opened 2023-05-19
Calc functions tests

pamelalozano16:
Context: [Calc functions proposal](https://github.com/sass/sass/blob/main/accepted/calc-functions.md)
[skip dart-sass]


Goodwine:
Nit, you want to add `[skip dart-sass]` in your PR description so the CI checks don't run the tests against the current dart sass version

You can also run `npm run lint-spec -- --fix` to automatically fix linter errors

spec/values/calculation/sqrt.hrx
```diff
@@ -0,0 +1,140 @@
...
+  ,
+1 | a {b: calc(sqrt(5%))}
+  |            ^^^^^^^^
+  '
+  input.scss 1:12  root stylesheet
+
+<===>
+================================================================================
+<===> preserved/operation/unitless_and_unit/first/input.scss
+a {b: calc(sqrt(16))}
```
Goodwine, line 95: Nit: consider adding a test that would produce an a non-integer number such as sqrt(2)

spec/values/calculation/sqrt.hrx
```diff
@@ -0,0 +1,140 @@
...
+
+<===> preserved/operation/unitless_and_unit/first/output.css
+a {
+  b: 4;
+}
+
+<===>
+================================================================================
```
Goodwine, line 103: Add a test for a negative number
I assume a negative number returns an error? or `NaN`? 🤔

Goodwine (changes requested):

spec/values/calculation/sqrt.hrx
pamelalozano16: Done! :)

spec/values/calculation/sqrt.hrx
pamelalozano16: Oh yes, nice.

spec/values/calculation/sqrt.hrx
```diff
@@ -0,0 +1,164 @@
...
+a {b: calc(sqrt(16))}
+
+<===> preserved/operation/unitless/first/output.css
+a {
+  b: 4;
+}
+
+<===>
+================================================================================
+<===> preserved/operation/unitless/second/input.scss
```
Goodwine, line 108: Nit: Rather than `/first/` or `/second/` consider using a more descriptive test name such as `/integer/` and `/float/` (or something similar)

Goodwine (commented):
LGTM but I'll leave it to Jen to do a final approval.

Note that if you get an approval, you should not submit until both PRs are approved since they should be merged roughly at the same time.

jathak (commented):
Looking at the spec (both ours and [the CSS spec](https://drafts.csswg.org/css-values-4/#calc-syntax)), I realize now that I mistakenly had the impression (and then shared that with you) that these new math functions have to be contained within another `calc` to work (MDN was unhelpful here its examples for `sqrt` and most of the other functions _did_ have them nested).

In actuality, code like
```css
a {
  b: sqrt(4);
}
```

is valid and the `sqrt(4)` should simplify to 2.

Given that, most of these tests should just have `sqrt` without a `calc` wrapping them (maybe have one test like that to test the nesting). They otherwise look good (though I'll wait to give final approval until all of the functions are implemented/tested).

Sorry for the confusion!

spec/values/calculation/sqrt.hrx
pamelalozano16: 👍

spec/core_functions/math/abs.hrx
```diff
@@ -68,15 +74,17 @@ Recommendation: math.div(-7px, 4em)
...
-    input.scss 1:11  root stylesheet
+    input.scss 2:16  root stylesheet
 
 <===>
 ================================================================================
 <===> named/input.scss
-a {b: abs($number: -3)}
+@use "sass:math" as math;
+$number: -3;
+a {b: math.abs($number)}
```
jathak, line 87: This isn't actually testing a named argument; it's testing passing a variable called `$number` to `abs`.

To test this with the function in the math module, this should be:
```scss
@use "sass:math";
a { b: math.abs($number: 3)}
```

However, since this is a case where even the global `abs` should still parse as a Sass function, the existing test of the global function (`abs($number: -3)`) should also be kept.

spec/values/calculation/abs.hrx
```diff
@@ -0,0 +1,158 @@
...
+More info and automated migrator: https://sass-lang.com/d/slash-div
+  ,
+2 |   a: 2px / abs(1.5);
+  |      ^^^^^^^^^^^^^^
+  '
+    input.scss 2:6  root stylesheet
+
+<===>
+================================================================================
+<===> preserved/unknown_variable/input.scss
```
jathak, line 150: This is a good test, but it's not actually a variable. CSS variables start with `--` and need to be referenced with `var` (e.g. `var(--test)`). This should be called something like `preserved/unquoted_string` and then there should be a separate `preserved/variable` test with a `var()`.

spec/values/calculation/abs.hrx
```diff
@@ -0,0 +1,158 @@
...
+a {b: abs(-123.456)}
+
+<===> negative/decimal/output.css
+a {
+  b: 123.456;
+}
+
+<===>
+================================================================================
+<===> named/input.scss
```
jathak, line 57: This is testing a Sass variable as an argument, not a named argument (which wouldn't be a calculation anyway, so should be tested with the Sass function). `sass_variable` would be an appropriate name for this test.

spec/values/calculation/abs.hrx
```diff
@@ -0,0 +1,158 @@
+<===> README.md
```
jathak, line 1: Here (and for each of the other new calculation functions), make sure to test each of the following:

- a CSS variable as an argument
- arguments with units (where appropriate for the function in question)
- something with arithmetic that the browser can handle, but can't be simplified at compile time, like `10px + 10%`
- partial simplification (e.g. `abs(1px + 2px - var(--example))` should simplify to `abs(3px - var(--example))`)

spec/values/calculation/acos.hrx
```diff
@@ -0,0 +1,134 @@
...
+terseness' sake isn't tested explicitly.
+
+<===>
+================================================================================
+<===> less_than_negative_one/input.scss
+a {b: acos(-2)}
+
+<===> less_than_negative_one/output.css
+a {
+  b: calc(NaN * 1deg);
```
jathak, line 12: @nex3: Is this behavior intended? This matches the behavior of `math.acos(-2)`, so it's correct by the spec, but it seems odd to have a calculation function simplify to a _different_ calculation function.

spec/values/calculation/round.hrx
```diff
@@ -0,0 +1,371 @@
...
+Error: 3px must be either nearest, up, down or to-zero.
+  ,
+1 | a {b: round(3px, 3px, 2px)}
+  |       ^^^^^^^^^^^^^^^^^^^^
+  '
+  input.scss 1:7  root stylesheet
+
+<===>
+================================================================================
+<===> perserved/unknown_variable/number/input.scss
```
jathak, line 64: Typo: preserved, not perserved

spec/values/calculation/round.hrx
```diff
@@ -0,0 +1,371 @@
...
+a {b: round(4px, nearest)}
+
+<===> perserved/unknown_variable/step/output.css
+a {
+  b: round(4px, nearest);
+}
+
+<===>
+================================================================================
+<===> preserved/step/zero/input.scss
```
jathak, line 117: `preserved` as a directory name should be for tests where the `round()` calculation function is preserved in the output. The test where it does simplify should omit this.

spec/values/calculation/round.hrx
```diff
@@ -0,0 +1,371 @@
...
+c {d: round($a)}
+
+<===> preserved/variable/one_arg/output.css
+c {
+  d: 12;
+}
+
+<===>
+================================================================================
+<===> variables/input.scss
```
jathak, line 336: Each of these four should be a separate test.

See https://github.com/sass/sass-spec/blob/main/STYLE_GUIDE.md#do-test-only-one-thing-per-spec

spec/values/calculation/round.hrx
```diff
@@ -0,0 +1,371 @@
+<===> README.md
```
jathak, line 1: Make sure to test the special cases for round both with and without units

spec/values/calculation/sqrt.hrx
```diff
@@ -0,0 +1,173 @@
+<===> README.md
+Most of the same behavior tested for `calc()` applies to `sqrt()`, but for
+terseness' sake isn't tested explicitly.
+
+<===>
+================================================================================
+<===> error/syntax/no_arg/input.scss
+a {b: calc(sqrt())}
```
jathak, line 8: These do not need to be nested in another calc

jathak (changes requested):

spec/core_functions/math/abs.hrx
```diff
@@ -1,5 +1,6 @@
 <===> zero/input.scss
-a {b: abs(0)}
+@use "sass:math" as math;
```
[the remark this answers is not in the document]
pamelalozano16: Got it, done!

spec/core_functions/math/abs.hrx
pamelalozano16: Nice catch, I was missing the try/catch in the abs function.

spec/values/calculation/round.hrx
pamelalozano16: Thanks!

spec/values/calculation/round.hrx
pamelalozano16: 👍

spec/values/calculation/sqrt.hrx
pamelalozano16: 👍

spec/values/calculation/abs.hrx
pamelalozano16: 👍

spec/values/calculation/round.hrx
pamelalozano16: Thanks, found this mistake in the min.hrx tests too, fixed it. :)

spec/values/calculation/abs.hrx
pamelalozano16: This was a nice checklist, I went through all the new files and added tests. Thanks! :)

spec/values/calculation/abs.hrx
pamelalozano16: I changed this test to a named argument. :)

jathak:
Will fully review this afternoon, but you can get the CI passing by adding an `options.yml` to files that are using `@use`.
```hrx
<===> options.yml
---
:ignore_for:
- libsass
```

spec/values/calculation/acos.hrx
nex3: Yes—it actually simplifies to a Number with value `NaN` and numerator unit `deg`, which is then _serialized_ as `calc(NaN * 1deg)`.

spec/values/calculation/mod.hrx
```diff
@@ -0,0 +1,273 @@
+<===> README.md
+Most of the same behavior tested for `calc()` applies to `mod()`, but for
+terseness' sake isn't tested explicitly.
+
+<===>
+================================================================================
+<===> error/syntax/no_arg/input.scss
+a {b: calc(mod())}
```
jathak, line 8: Other than perhaps for a test that explicitly tests nesting (which I don't think is necessary for each individual function), these tests should just test `mod()` on its own, rather than wrapping any of them in a `calc()`.

Same for `pow()`, `rem()`, and `sign()`.

spec/values/calculation/round.hrx
```diff
@@ -0,0 +1,422 @@
...
+a {b: round(to-zero, -120px, 25px)}
+
+<===> operation/to-zero/negative/output.css
+a {
+  b: -100px;
+}
+
+<===>
+================================================================================
+<===> NaN/simplify/input.scss
```
jathak, line 327: The test directories should probably be all lower case to avoid any potential issues with case-sensitive vs case-insensitive file systems.

@nex3: Is this something we should add to the style guide?

jathak (approved):
LGTM, other than some functions being unnecessarily wrapped in a `calc()` (though wait to merge this at the same time as the implementation)

spec/values/calculation/mod.hrx
pamelalozano16: Done :)

spec/values/calculation/abs.hrx
```diff
@@ -0,0 +1,255 @@
...
+<===> operation/incompatible_units/input.scss
+a {b: abs(17px * 5%)}
+
+<===> operation/incompatible_units/error
+Error: 85px*% isn't a valid CSS value.
+  ,
+1 | a {b: abs(17px * 5%)}
+  |       ^^^^^^^^^^^^^^
+  '
+  input.scss 1:7  root stylesheet
```
nex3, line 70: I don't think we need to test this separately from the above. The previous test already demonstrates that `abs()` passes through units as-is; this error is just Sass complaining about the units we already know `abs()` returns.

spec/values/calculation/abs.hrx
```diff
@@ -0,0 +1,255 @@
...
+
+<===>
+================================================================================
+<===> operation/integer/input.scss
```
nex3, line 74: Again, we don't need separate integer tests.

spec/values/calculation/abs.hrx
```diff
@@ -0,0 +1,255 @@
...
+================================================================================
+<===> preserved/unquoted_string/input.scss
+a {
+  b: abs(test);
+}
+
+<===> preserved/unquoted_string/output.css
+a {
+  b: abs(test);
+}
```
nex3, line 128: Similarly, these aren't testing anything that isn't already covered by the previous operation test. The only relevant question here is "Does `abs()` return a calculation if its argument can't be fully simplified?" and we know that it does. We can leave the question of "Which values can't be simplified?" up to the `calc()` tests.

spec/values/calculation/abs.hrx
```diff
@@ -0,0 +1,255 @@
...
+
+<===>
+================================================================================
+<===> named_argument/input.scss
+a {b: abs($number: -3)}
```
nex3, line 133: It's probably worth another test to verify that there's a strict delineation between "parsing `abs()` as a Sass function" and "parsing `abs()` as a calculation". For example, `abs($number: var(--c))` should fail.

spec/values/calculation/abs.hrx
```diff
@@ -0,0 +1,255 @@
...
+<===> error/incompatible_units/input.scss
+a {b: abs(17px + 4deg)}
+
+<===> error/incompatible_units/error
+Error: 17px and 4deg are incompatible.
+  ,
+1 | a {b: abs(17px + 4deg)}
+  |           ^^^^^^^^^^^
+  '
+  input.scss 1:11  root stylesheet
```
nex3, line 195: As above, this is just testing the simplification logic which is already covered by the `calc()` tests.

spec/values/calculation/acos.hrx
```diff
@@ -0,0 +1,191 @@
+<===> README.md
+Most of the same behavior tested for `calc()` applies to `acos()`, but for
+terseness' sake isn't tested explicitly.
+
+<===>
+================================================================================
+<===> less_than_negative_one/input.scss
```
nex3, line 7: Test negative one as well.

nex3 (changes requested):

spec/values/calculation/abs.hrx
```diff
@@ -0,0 +1,255 @@
...
+a {b: abs(0)}
+
+<===> zero/output.css
+a {
+  b: 0;
+}
+
+<===>
+================================================================================
+<===> positive/integer/input.scss
```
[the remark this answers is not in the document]
pamelalozano16: Got it, thanks! I left both of them but for positive/negative tests.

spec/values/calculation/abs.hrx
nex3: Looks like the first two of these are still here.

spec/values/calculation/acos.hrx
```diff
@@ -0,0 +1,185 @@
...
+<===> error/too_many_args/input.scss
+a {b: acos(0, 0)}
+
+<===> error/too_many_args/error
+Error: expected "+", "-", "*", "/", or ")".
+  ,
+1 | a {b: acos(0, 0)}
+  |             ^
+  '
+  input.scss 1:13  root stylesheet
```
nex3, line 159: Also test too few arguments.

spec/values/calculation/atan2.hrx
```diff
@@ -0,0 +1,204 @@
...
+
+<===>
+================================================================================
+<===> operation/math/input.scss
+a {b: atan2(10 + 6, 8 - 1)}
+
+<===> operation/math/output.css
+a {
+  b: 66.3706222693deg;
+}
```
nex3, line 33: As with abs(), all you need is one test to verify that the normal calculation simplification behavior happens—and it should probably include both an operation that can be simplified and something that can't.

spec/values/calculation/atan2.hrx
```diff
@@ -0,0 +1,204 @@
...
+
+<===>
+================================================================================
+<===> compatible_units/input.scss
+a {b: atan2(1cm, -10mm)}
+
+<===> compatible_units/output.css
+a {
+  b: 135deg;
+}
```
nex3, line 23: Also test:

* a real unit and a percentage
* a complex unit and a percentage (this should fail)
* two percentages
* two fake units
* a real unit and a fake unit

As a style consideration, I'd probably call these directories `units/none`, `units/compatible`, `units/real_and_unknown`, `units/complex_and_unknown`, `units/unknown`, `units/fake`, `units/real_and_fake`.

spec/values/calculation/atan2.hrx
```diff
@@ -0,0 +1,204 @@
...
+a {b: atan2(1, 1px)}
+
+<===> error/incompatible_units/error
+Error: 1 and 1px are incompatible.
+  ,
+1 | a {b: atan2(1, 1px)}
+  |             ^ 1
+  |                === 1px
+  '
+  input.scss 1:13  root stylesheet
```
nex3, line 113: Also test two known-incompatible units.

spec/values/calculation/atan2.hrx
```diff
@@ -0,0 +1,204 @@
...
+
+<===>
+================================================================================
+<===> error/one_arg/input.scss
```
nex3, line 117: We consistently call this `too_few_args` elsewhere. It's important to pay attention to the surrounding style and try to match it whenever possible. It both saves you time that you would otherwise have spent coming up with new conventions, and helps keep the codebase (or in this case test suite) consistent and easy to navigate.

spec/values/calculation/atan2.hrx
```diff
@@ -0,0 +1,204 @@
...
+<===> error/syntax/invalid_arg/input.scss
+a {b: atan2($)}
+
+<===> error/syntax/invalid_arg/error
+Error: Expected identifier.
+  ,
+1 | a {b: atan2($)}
+  |              ^
+  '
+  input.scss 1:14  root stylesheet
```
nex3, line 139: When testing error cases, it's a good idea to make sure that everything _except_ the error you're intending to test is valid so that you can be sure the code is failing due to the correct error. In this case, this is both invalid syntax _and_ only one argument.

spec/values/calculation/atan2.hrx
```diff
@@ -0,0 +1,204 @@
...
+<===> error/zero_args/input.scss
+a {b: atan2()}
+
+<===> error/zero_args/error
+Error: Expected number, variable, function, or calculation.
+  ,
+1 | a {b: atan2()}
+  |             ^
+  '
+  input.scss 1:13  root stylesheet
```
nex3, line 178: You don't need to test _all_ argument numbers below the minimum, just verify that passing one too few args fails.

spec/values/calculation/cos.hrx
```diff
@@ -0,0 +1,247 @@
...
+<===> turn/output.css
+a {
+  b: 1;
+}
+
+<===>
+================================================================================
+<===> negative_infinity/input.scss
+@use "sass:math";
+a {b: cos(math.div(-1, 0))}
```
nex3, line 49:
```suggestion
a {b: cos(-infinity)}
```

Sass supports [calculation constants](https://developer.mozilla.org/en-US/docs/Web/CSS/calc-constant) in calculations.

spec/values/calculation/acos.hrx
```diff
@@ -0,0 +1,185 @@
...
+
+<===>
+================================================================================
+<===> zero_fuzzy/input.scss
+a {b: acos(0.000000000001)}
+
+<===> zero_fuzzy/output.css
+a {
+  b: 89.9999999999deg;
+}
```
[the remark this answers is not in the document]
pamelalozano16: Right, thanks! :)

spec/values/calculation/atan.hrx
```diff
@@ -0,0 +1,195 @@
...
+
+<===>
+================================================================================
+<===> greater_than_one/input.scss
+a {b: atan(2)}
+
+<===> greater_than_one/output.css
+a {
+  b: 63.4349488229deg;
+}
```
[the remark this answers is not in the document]
pamelalozano16: Ok got it, I only left tests for infinities, zero, and one.

spec/values/calculation/abs.hrx
pamelalozano16: Oh I though you mean only the `abs(test)` case. But yes, I see both are tested in the previous operation test.

spec/values/calculation/acos.hrx
pamelalozano16: Since it's a single argument function, isn't it enough with the `too_few_args` test that's [here](https://github.com/sass/sass-spec/blob/0e3fef7a107f236920e480e574c01880a9cf4bb7/spec/values/calculation/acos.hrx#L110)?

spec/values/calculation/acos.hrx
```diff
@@ -0,0 +1,185 @@
...
+<===> error/named_argument/input.scss
+a {b: acos($number: 3)}
+
+<===> error/named_argument/error
+Error: expected "+", "-", "*", "/", or ")".
+  ,
+1 | a {b: acos($number: 3)}
+  |                   ^
+  '
+  input.scss 1:19  root stylesheet
```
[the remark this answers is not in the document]
pamelalozano16: Got it, thanks

spec/values/calculation/acos.hrx
```diff
@@ -0,0 +1,170 @@
...
+================================================================================
+<===> operation/simplification/input.scss
+a {
+  b: acos(3px - 1px + var(--c));
+}
+
+<===> operation/simplification/output.css
+a {
+  b: acos(2px + var(--c));
+}
```
nex3, line 66: Since this is the only test for simplification behavior, and the point is just to test that simplification is happening at all, I think just calling it `simplification` makes the most sense.

spec/values/calculation/exp.hrx
nex3, line 1: This doesn't have any unit-related tests at all.

spec/values/calculation/hypot.hrx
```diff
@@ -0,0 +1,241 @@
...
+<===>
+================================================================================
+<===> simplification/input.scss
+@use "sass:math";
+a {b: hypot(3 + 2, 4 * 10, 5 * 40, math.div(6, 2.54), 7 * math.div(96, 2.54))}
+
+<===> simplification/output.css
+a {
+  b: 334.1051331579;
+}
```
nex3, line 80: To match other simplification tests, this should include both an expression that can simplify and one that can't. Also, I don't think it needs to have so many arguments.

spec/values/calculation/hypot.hrx
```diff
@@ -0,0 +1,241 @@
...
+
+<===>
+================================================================================
+<===> percentage/input.scss
+a {b: hypot(10%, 5%)}
+
+<===> percentage/output.css
+a {
+  b: hypot(10%, 5%);
+}
```
nex3, line 110: This should be with the `units` tests. Generally, different test cases should have as consistent a structure between them as possible (given that they sometimes have different testing needs). Same goes for other unit tests here.

spec/values/calculation/round/one_argument.hrx
```diff
@@ -0,0 +1,200 @@
...
+================================================================================
+<===> interpolation/input.scss
+a {
+  e: round(3px + #{"3"});
+}
+
+<===> interpolation/output.css
+a {
+  e: round(3px + 3);
+}
```
nex3, line 77: Shouldn't this be under `preserved/`?

spec/values/calculation/round/one_argument.hrx
```diff
@@ -0,0 +1,200 @@
...
+  b: round(1px + 2px - var(--c))
+}
+
+<===> error/variables/error
+Error: Single argument 3px - var(--c) expected to be simplifiable.
+  ,
+2 |   b: round(1px + 2px - var(--c))
+  |      ^^^^^^^^^^^^^^^^^^^^^^^^^^^
+  '
+  input.scss 2:6  root stylesheet
```
nex3, line 140: What is this testing that isn't covered by `error/unsimplifiable`?

spec/values/calculation/round/one_argument.hrx
```diff
@@ -0,0 +1,200 @@
...
+Error: Only 1 argument allowed, but 4 were passed.
+  ,--> input.scss
+1 | a {b: round(1, 2, 3, 4)}
+  |       ^^^^^^^^^^^^^^^^^ invocation
+  '
+  ,--> sass:math
+1 | @function round($number) {
+  |           ============== declaration
+  '
+  input.scss 1:7  root stylesheet
```
nex3, line 187: These tests aren't specific to the single-argument round function, so they should probably be in a different directory.

spec/values/calculation/round/three_arguments.hrx
```diff
@@ -0,0 +1,241 @@
+<===> README.md
+Most of the same behavior tested for `calc()` applies to `round()`, but for
+terseness' sake isn't tested explicitly.
+
+<===>
+================================================================================
+<===> preserved/operation/nearest/math/input.scss
```
nex3, line 7: This test path is deeply nested for no clear reason. Nested directories should be used for grouping tests, and otherwise the innermost directory name should be an underscore-separated name for the test.

It's also not clear to me why the test name calls out "nearest" when the test case doesn't need to be varied on rounding strategy, or why it calls out "math" at all.

spec/values/calculation/round/three_arguments.hrx
```diff
@@ -0,0 +1,241 @@
...
+a {b: math.div(1, round(down, -0, infinity))}
+
+<===> down/negative_zero/output.css
+a {
+  b: calc(-infinity);
+}
+
+<===>
+================================================================================
+<===> operation/positive/input.scss
```
nex3, line 51: Why is this in a directory named `operation`?

spec/values/calculation/round/three_arguments.hrx
```diff
@@ -0,0 +1,241 @@
...
+================================================================================
+<===> interpolation/input.scss
+a {
+  e: round(#{"up"}, number, 9px);
+}
+
+<===> interpolation/output.css
+a {
+  e: round(up, number, 9px);
+}
```
nex3, line 188: It's less important to test every kind of calculation argument that doesn't simplify to a number (these are all already covered by the `calc()` tests and we can assume that the same simplification logic is in use), and more important to test that _some_ non-number argument works in every argument position in the function.

spec/values/calculation/round/three_arguments.hrx
```diff
@@ -0,0 +1,241 @@
...
+
+<===>
+================================================================================
+<===> math/input.scss
```
nex3, line 192: First of all, [error specs go in the `error` directory](https://github.com/sass/sass-spec/blob/main/STYLE_GUIDE.md#do-put-error-specs-in-a-separate-error-directory). Second, `math` isn't the term used yb the spec (or other tests for this)—it's `operation`. Third, neither `math` nor `operation` on its own is a good name here. Think of the name of a test as a terse version of you describing to a reader what the test is supposed to verify. In this case, it's verifying that the strategy requires an unquoted string with a given value, so something like `error/strategy/operation` would be a better fit.

spec/values/calculation/round/three_arguments.hrx
```diff
@@ -0,0 +1,241 @@
...
+Error: 12px must be either nearest, up, down or to-zero.
+  ,
+2 |   e: round(10px + 2px, 8px, 9px);
+  |      ^^^^^^^^^^^^^^^^^^^^^^^^^^^
+  '
+  input.scss 2:6  root stylesheet
+
+<===>
+================================================================================
+<===> error/syntax/strategy/number/input.scss
```
nex3, line 206: This isn't a syntax error—syntax errors are thrown at parse time, before the stylesheet is even evaluated.

spec/values/calculation/round/three_arguments.hrx
```diff
@@ -0,0 +1,241 @@
...
+Error: 3px must be either nearest, up, down or to-zero.
+  ,
+1 | a {b: round(3px, 3px, 2px)}
+  |       ^^^^^^^^^^^^^^^^^^^^
+  '
+  input.scss 1:7  root stylesheet
+
+<===>
+================================================================================
+<===> error/variable/number_and_step/input.scss
```
nex3, line 219: This test contains neither a Sass variable nor a CSS variable. (An unquoted string isn't a variable in any context.)

spec/values/calculation/round/two_arguments.hrx
nex3, line 1: This should test the same full range of unit combinations as other two-argument functions like `atan2()`.

nex3 (changes requested):

spec/values/calculation/atan2.hrx
```diff
@@ -0,0 +1,210 @@
...
+  ,
+1 | a {b: atan2(1px*2px, 10%)}
+  |             ^^^^^^^
+  '
+  input.scss 1:13  root stylesheet
+
+<===>
+================================================================================
+<===> error/syntax/invalid_arg/input.scss
+a {b: atan2($c, $)}
```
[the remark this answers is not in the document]
pamelalozano16: Ohhh gotcha thanks

spec/values/calculation/exp.hrx
pamelalozano16: Since the `exp` function doesn't take units, I just added the two error tests for known and unknown units. 🤔 Am I missing something?

spec/values/calculation/round/three_arguments.hrx
pamelalozano16: I used to have all this directories when I had the tests on a single file. I forgot to update the names in this one, thanks!

spec/values/calculation/exp.hrx
nex3: Ah, I must have missed them, sorry

spec/values/calculation/round/one_argument.hrx
pamelalozano16: Does it make more sense to move it to the three-argument round tests?

spec/values/calculation/round/three_arguments.hrx
pamelalozano16: Got it, I replaced `number` with a number with real units. Thank you :)

spec/values/calculation/abs.hrx
```diff
@@ -0,0 +1,182 @@
...
+}
+
+<===> simplification/output.css
+a {
+  b: abs(3px - var(--c));
+}
+
+<===>
+================================================================================
+<===> operation/preserves_units/input.scss
```
pamelalozano16, line 49: Not a 100% sure if we should keep this test, I don't think the simplification test covers this, right?

spec/values/calculation/round/one_argument.hrx
```diff
@@ -0,0 +1,151 @@
...
+a {b: round(-7px / 4em) * 1em}
+
+<===> operation/preserves_units/output.css
+a {
+  b: -2px;
+}
+
+<===>
+================================================================================
+<===> preserved/variable/input.scss
```
pamelalozano16, line 47: Since a single-argument needs simplifiable parameters, I separated this test. (Check line #116)

spec/values/calculation/sign.hrx
```diff
@@ -0,0 +1,130 @@
...
+a {b: sign(0)}
+
+<===> zero/output.css
+a {
+  b: 0;
+}
+
+<===>
+================================================================================
+<===> zero_fuzzy/input.scss
```
pamelalozano16, line 37: Seemed important to keep since 0 is a limit of the function.

spec/values/calculation/log.hrx
```diff
@@ -0,0 +1,250 @@
...
+Error: Number 2px*px isn't compatible with CSS calculations.
+  ,
+1 | a {b: log(1px*2px, 10%)}
+  |           ^^^^^^^
+  '
+  input.scss 1:11  root stylesheet
+
+<===>
+================================================================================
+<===> error/unit/known/input.scss
```
nex3, line 163: `units` to match the other tests

spec/values/calculation/log.hrx
```diff
@@ -0,0 +1,250 @@
...
+  ,
+1 | a {b: log(0, 0, 0)}
+  |               ^
+  '
+  input.scss 1:15  root stylesheet
+
+<===>
+================================================================================
+<===> error/syntax/invalid_arg/input.scss
+a {b: log($, $c)}
```
nex3, line 229: Undefined variable

spec/values/calculation/mod.hrx
nex3, line 1: Also test mixed positive/negative dividend and modulus in cases where they _don't_ produce 0.

spec/values/calculation/pow.hrx
```diff
@@ -0,0 +1,192 @@
...
+  '
+  input.scss 1:7  root stylesheet
+
+<===>
+================================================================================
+<===> error/units/unknown_and_unitless/input.scss
+a {b: pow(10%, 10)}
+
+<===> error/units/unknown_and_unitless/error
+Error: 10% and 10 are incompatible.
```
nex3, line 94: This error is misleading—the issue is that _no_ units are allowed. Making them compatible wouldn't fix the error.

spec/values/calculation/round/one_argument.hrx
```diff
@@ -0,0 +1,151 @@
...
+<===> preserved/variable/output.css
+a {
+  b: round(var(--c));
+}
+
+<===>
+================================================================================
+<===> preserved/interpolation/input.scss
+a {
+  e: round(3px + #{"3"});
```
nex3, line 61: To make it clearer _why_ this case is preserved, consider making it something that produces valid CSS like `round(#{"5.5px, 1px"})`

spec/values/calculation/round/two_arguments.hrx
```diff
@@ -0,0 +1,248 @@
...
+
+<===> number/infinity/output.css
+a {
+  b: calc(infinity);
+}
+
+<===>
+================================================================================
+<===> nan/input.scss
+a {b: round(calc(NaN), calc(NaN))}
```
nex3, line 114: You don't need `calc()`s here, since `round()` is already a calculation context.

spec/values/calculation/sign.hrx
```diff
@@ -0,0 +1,130 @@
...
+a {b: sign(-5.6)}
+
+<===> negative/output.css
+a {
+  b: -1;
+}
+
+<===>
+================================================================================
+<===> zero/input.scss
```
nex3, line 27: Also test negative zero

spec/values/calculation/sqrt.hrx
```diff
@@ -0,0 +1,146 @@
...
+a {b: sqrt(0)}
+
+<===> zero/output.css
+a {
+  b: 0;
+}
+
+<===>
+================================================================================
+<===> result_is_integer/input.scss
```
nex3, line 17: You don't need separate integer/float tests

spec/values/calculation/sqrt.hrx
```diff
@@ -0,0 +1,146 @@
...
+  ,
+1 | a {b: sqrt(1%)}
+  |       ^^^^^^^^
+  '
+  input.scss 1:7  root stylesheet
+
+<===>
+================================================================================
+<===> error/unit/real/input.scss
+a {b: calc(sqrt(16px))}
```
nex3, line 73: Why is this in `calc()`?

nex3 (changes requested):

spec/values/calculation/abs.hrx
```diff
@@ -0,0 +1,182 @@
...
+More info and automated migrator: https://sass-lang.com/d/slash-div
+  ,
+2 |   a: 2px / abs(1.5);
+  |      ^^^^^^^^^^^^^^
+  '
+    input.scss 2:6  root stylesheet
+
+<===>
+================================================================================
+<===> error/sass_script/variable/input.scss
```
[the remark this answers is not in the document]
pamelalozano16: Right, done!

spec/values/calculation/hypot.hrx
```diff
@@ -0,0 +1,235 @@
...
+  ,
+1 | a {b: hypot()}
+  |             ^
+  '
+  input.scss 1:13  root stylesheet
+
+<===>
+================================================================================
+<===> error/syntax/invalid_arg/input.scss
+a {b: hypot($c, $, $e)}
```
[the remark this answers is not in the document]
pamelalozano16: Weird, I thought I changed these (both in hypot and log).

spec/values/calculation/mod.hrx
pamelalozano16: Added 3 tests both to mod and rem :)

spec/values/calculation/sqrt.hrx
pamelalozano16: For `sqrt` specifically I thought it might be useful, but now that I think about it it's the same since Sass treats all numbers as floats.

spec/values/calculation/round/one_argument.hrx
pamelalozano16: Done, I changed this test and moved it to the two-argument directory since the result contains two arguments.

spec/values/calculation/pow.hrx
pamelalozano16: So, I was tracing the error and it seems like the `Expected 10% to have no units.` error is being generated in [this](https://github.com/sass/dart-sass/blob/4d38e3b132dd8b4751c8d1f5d23869593b49a146/lib/src/visitor/async_evaluate.dart#L2385) switch block. The problem is, the block is catching the exception and calling the [`_verifyCompatibleNumbers()`](https://github.com/sass/dart-sass/blob/4d38e3b132dd8b4751c8d1f5d23869593b49a146/lib/src/visitor/async_evaluate.dart#L2413) instead of rethrowing. I can imagine there's a good reason for that so I'm not sure how to proceed. Thoughts? 🤔

spec/values/calculation/atan2.hrx
```diff
@@ -0,0 +1,200 @@
...
+}
+
+<===> units/unknown/output.css
+a {
+  b: atan2(1%, 2%);
+}
+
+<===>
+================================================================================
+<===> units/fake/input.scss
```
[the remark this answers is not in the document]
pamelalozano16: I added this test case for all two-argument functions. In the case of `atan2`, it returns the result of the calculation with the `deg` unit, and the other functions return the result with the `foo` unit. This is happening because we are only checking if the units are _compatible_, so two numbers with the `foo` unit would be compatible.
I do see that the spec specifies `round`, `mod`, and `rem` accept any units that are mutually compatible. However, `atan2` and `hypot` do require known units. Is there a function in to identify real units?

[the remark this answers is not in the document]
nex3: I believe the existing behavior is correct. "Known units" is [defined as](https://github.com/sass/sass/blob/main/accepted/calc-functions.md#known-units) "A number has _known units_ unless it has unit `%`.", so `foo` counts.

spec/values/calculation/pow.hrx
nex3: If you guard that call by checking if the original error message actually mentions the numbers being incompatible, it should avoid changing an unrelated error.

spec/values/calculation/rem.hrx
```diff
@@ -0,0 +1,324 @@
...
+}
+
+<===>
+================================================================================
+<===> positive_and_negative/input.scss
+a {b: rem(2, -5)}
+
+<===> positive_and_negative/output.css
+a {
+  b: -5;
```
nex3, line 90: This output looks wrong. From [the CSS spec](https://drafts.csswg.org/css-values-4/#funcdef-mod), `mod()` and `rem()` "return the difference between A and the nearest integer multiple of B either above or below A." That means `rem(2, -5)` should be either -3 (the difference between 2 and the nearest integer multiple of -5 above 2, which is 5) or 2 (the difference between 2 and the nearest integer multiple of -5 _below_ 2, which is 0). It definitely shouldn't be -5.

Specifically, since `rem()` "chooses the integer multiple of B that puts the value between zero and -B, avoiding changing the sign of the value" when the two arguments have different signs, this should be 2 and the `negative_and_positive` case should be -3.

spec/values/calculation/round/one_argument.hrx
```diff
@@ -0,0 +1,139 @@
...
+More info and automated migrator: https://sass-lang.com/d/slash-div
+  ,
+2 |   a: 2px / round(1.5);
+  |      ^^^^^^^^^^^^^^^^
+  '
+    input.scss 2:6  root stylesheet
+
+<===>
+================================================================================
+<===> error/sass_script/variable_named_argument/input.scss
```
nex3, line 91: Now that you have a separate `round/error.hrx`, all the error tests should be in there (under their own separate `one_argument/` etc subdirectories).

spec/values/calculation/round/three_arguments.hrx
```diff
@@ -0,0 +1,247 @@
...
+}
+
+<===>
+================================================================================
+<===> negative/input.scss
+a {b: round(down, -10, -100)}
+
+<===> negative/output.css
+a {
+  b: 0;
```
nex3, line 22: Shouldn't this be -100, since we're rounding down to the next lowest multiple of -100?

spec/values/calculation/round/three_arguments.hrx
```diff
@@ -0,0 +1,247 @@
+<===> README.md
+Most of the same behavior tested for `calc()` applies to `round()`, but for
+terseness' sake isn't tested explicitly.
+
+<===>
+================================================================================
+<===> positive/input.scss
```
nex3, line 7: It's confusing that the name of this test implies that it's not specific to a given strategy, when in fact all three-argument `round()` calls _have to be_ specific to a given strategy. I think it would make more sense to have all fully-simplified tests be in strategy-specific directories, and test each relevant combination for each strategy.

spec/values/calculation/round/two_arguments.hrx
```diff
@@ -0,0 +1,260 @@
...
+
+<===> step/zero/output.css
+a {
+  b: calc(NaN * 1px);
+}
+
+<===>
+================================================================================
+<===> step/infinity/input.scss
+a {b: round(-0, infinity)}
```
nex3, line 106: Why is this testing -0 specifically? Does it have different behavior than positive zero?

spec/values/calculation/sign.hrx
```diff
@@ -0,0 +1,141 @@
...
+a {b: sign(0)}
```
nex3, line 28: This should divide 1 by the result so we can verify that it's actually _positive_ zero, since this can sometimes return negative zero.

spec/values/calculation/sign.hrx
```diff
@@ -0,0 +1,141 @@
...
+}
+
+<===>
+================================================================================
+<===> preserves_units/input.scss
+a {b: sign(-7px / 4em) * 1em}
+
+<===> preserves_units/output.css
+a {
+  b: -1em;
```
nex3, line 63: Shouldn't this be `-1px` if it preserves units?

nex3 (changes requested):

spec/values/calculation/rem.hrx
jathak: I think this may be due to a bug in [the Sass spec](https://github.com/sass/sass/blob/main/accepted/calc-functions.md?plain=1#L376), which says that if the signs are different, then `rem(dividend, modulus)` should equal `mod(dividend, modulus) - dividend`, when it should equal `mod(dividend, modulus) - modulus`

spec/values/calculation/round/three_arguments.hrx
pamelalozano16: Well, the formula for the `down` strategy is: `(number.value / stepWithNumberUnit).floor() * stepWithNumberUnit` so in this case `(-10 / -100).floor() == 0` which is why it's returning a 0. For it to result in `-100`, it would have to be the `up` strategy.

spec/values/calculation/round/three_arguments.hrx
```diff
@@ -0,0 +1,204 @@
...
+a {b: round(to-zero, 120px, 25px)}
+
+<===> strategy/to-zero/positive/output.css
+a {
+  b: 100px;
+}
+
+<===>
+================================================================================
+<===> strategy/to-zero/negative/input.scss
```
pamelalozano16, line 142: All other edge cases are implicitly tested either in the two-argument file (since `nearest` is the default strategy), or in the `up` / `down` directory, since it's the same formula. (See file [here](https://github.com/sass/dart-sass/blob/8fdd53ebe2b4174e0e31d6c2bf6f98e834875dca/lib/src/value/calculation.dart#L709))

spec/values/calculation/round/three_arguments.hrx
nex3: The important question isn't what the code says, it's what the spec says (ultimately this CSS spec, but in this case I think the Sass spec is correctly in agreement with it). The point of tests is to make sure the code is correct, not just to reify the code's current behavior.

Here's what the spec says:

> * Let `upper` and `lower` be the two integer multiples of `step` which
>   are closest to `number` such that `upper` is greater than `lower`. If
>   `upper` would be 0, it's specifically `-0`; if `lower` would be zero,
>   it's specifically `-0`.
> 
> * ...
>
> * If `strategy`'s value is `"down"`, return `lower`.

The lower integer multiple of -100 which is closest to -10 is -100, not 0.

spec/values/calculation/round/three_arguments.hrx
nex3: I like the current structure you have, where you have strategy subdirectories that fully exercise each strategy but have tests that aren't particularly strategy-specific outside those subdirectories.

spec/values/calculation/round/three_arguments.hrx
nex3: I think it would be clearer to test `nearest` in the `three_arguments` directory and leave them implicit in `two_arguments` instead (with just enough to verify that `nearest` is in fact the strategy being used).

It's probably better to duplicate tests between `up`/`down` and `nearest`, just to be on the safe side.

spec/values/calculation/round/three_arguments.hrx
pamelalozano16: Right! With this I found a bug in the implementation. I tested the round function in Safari just to make sure, and yes `round(down, -10, -100)` should be `-100`. So it turns out that, **if the strategy is down** and the **step is negative**, we should do `.ceil()` instead of `.floor()`. Same with the `up` strategy but the other way around. I updated this in the implementation.

spec/values/calculation/round/three_arguments.hrx
pamelalozano16: Got it! Just added the same tests for `nearest` in the three_argument directory

spec/values/calculation/pow.hrx
pamelalozano16: Okay, I added an `if` statement checking if the message contains the word `compatible`. So this applies to both compatibility error messages: `1px and $min: 1deg have incompatible units` or `Number 2px*px isn't compatible with CSS calculations.`

spec/values/calculation/round/three_arguments.hrx
```diff
@@ -2,344 +2,6 @@
 Most of the same behavior tested for `calc()` applies to `round()`, but for
 terseness' sake isn't tested explicitly.
 
-<===>
```
pamelalozano16: After adding the tests from [wpt](https://github.com/web-platform-tests/wpt/blob/master/css/css-values/round-function.html) this file was too large, so I created a strategies directory. I also added some from [this file](https://github.com/web-platform-tests/wpt/blob/master/css/css-values/round-mod-rem-computed.html#L210) to mod and rem that I thought were missing.

spec/values/calculation/round/strategies/down.hrx
```diff
@@ -0,0 +1,197 @@
+<===> README.md
+Most of the same behavior tested for `calc()` applies to `round()`, but for
+terseness' sake isn't tested explicitly.
+
+<===>
+================================================================================
+<===> strategy/down/step_is_multiple_of_number/input.scss
```
nex3, line 7: You don't need `strategy/down` here; this is already in a `strategy/down` directory.

spec/values/calculation/round/strategies/down.hrx
nex3, line 1: It's worth adding a `round/stragies/README.md` that mentions that some test cases were sourced from the WPT suite.

spec/values/calculation/round/two_arguments.hrx
```diff
@@ -92,34 +92,32 @@ a {
...
-a {b: math.div(1, round(0, infinity))}
+<===> nested_function/input.scss
+a {b: round(round(100, 10), 10)}
 
-<===> step/positive_and_infinity/output.css
+<===> nested_function/output.css
 a {
-  b: calc(infinity);
+  b: 100;
 }
```
nex3: This is just a special case of simplification.

spec/values/calculation/round/two_arguments.hrx
```diff
@@ -134,10 +132,70 @@ a {
 
 <===>
 ================================================================================
-<===> infinity/input.scss
+<===> infinity_and_positive/input.scss
```
nex3, line 135: I don't think you need to be as thorough as this for the two-argument `round()`, since it's effectively already tested by `strategy/nearest`. Just have enough tests to verify that it's using the `nearest` strategy at all.

spec/values/calculation/round/one_argument.hrx
nex3, line 1: Also test `round(1 + 1px)`.

nex3 (changes requested):

spec/values/calculation/mod.hrx
```diff
@@ -322,3 +332,55 @@ Error: expected "+", "-", "*", "/", ",", or ")".
   |             ^
   '
   input.scss 1:13  root stylesheet
+
+<===>
+================================================================================
+<===> math/unknown_units/input.scss
```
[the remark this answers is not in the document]
pamelalozano16: We already have a case for `positive_zero` so I'm just deleting this test case.

spec/values/calculation/round/two_arguments.hrx
pamelalozano16: None of these are repeated, here we test with the `number` as infinity, and in `nearest.hrx` we test `step` as infinity. Since all of these edge cases are being tested in [wpt](https://github.com/web-platform-tests/wpt/blob/master/css/css-values/round-function.html), I think we should keep them, what do you think?

spec/values/calculation/round/strategy/README.md
```diff
@@ -0,0 +1,5 @@
+### Three-argument round strategy testing
+
+Tests the `round` function with the various combinations of rounding modes to cover all possibilities and edge cases to ensure robustness. Some tests are based on [web-platform-tests] for the CSS round function.
+
+[web-platform-tests]=https://github.com/web-platform-tests/wpt/blob/master/css/css-values/round-function.html
```
nex3, line 5: Wrong syntax for markdown link references. There's syntax documentation [here](https://docs.github.com/en/get-started/writing-on-github/getting-started-with-writing-and-formatting-on-github/basic-writing-and-formatting-syntax).

spec/values/calculation/round/strategy/down.hrx
```diff
@@ -0,0 +1,198 @@
...
+
+<===>
+================================================================================
+<===> nested_function/input.scss
+a {b: round(down, round(100, 10) + 2 * 3, 10)}
+
+<===> nested_function/output.css
+a {
+  b: 100;
+}
```
nex3, line 198: I don't think this really tests anything more than the simplification logic that's tested elsewhere.

spec/values/calculation/round/two_arguments.hrx
nex3: It's not clear to me why the WPT test some cases with explicit `nearest` and some without. I think for us, it probably makes more sense to move all of these explicitly into `round/strategy/nearest`. A casual reader won't know what's from WPT and what isn't, so that doesn't make for a great organizing principle.

nex3 (changes requested):

spec/values/calculation/round/one_argument.hrx
```diff
@@ -0,0 +1,97 @@
...
+a {b: round(-7px / 4em) * 1em}
+
+<===> preserves_units/output.css
+a {
+  b: -2px;
+}
+
+<===>
+================================================================================
+<===> legacy_round/preserves_units/input.scss
```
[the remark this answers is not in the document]
pamelalozano16: Ok, changed the name to `preserves_single_unit` to avoid conflict with the previous test `preserves_units`.

spec/values/calculation/round/error.hrx
```diff
@@ -0,0 +1,259 @@
...
+Error: Only 1 argument allowed, but 3 were passed.
+  ,--> input.scss
+1 | a {b: round(nearest, 0, "0")}
+  |       ^^^^^^^^^^^^^^^^^^^^^^ invocation
+  '
+  ,--> sass:math
+1 | @function round($number) {
+  |           ============== declaration
+  '
+  input.scss 1:7  root stylesheet
```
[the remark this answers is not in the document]
pamelalozano16: Ohhh right, got it.

nex3 (approved):

PR #1916, opened 2023-06-03
Report JS API tests as they pass or fail

nex3:
The default reporter only prints tests on a single line as they run,
which makes it much harder to debug failures, particularly on CI.


Goodwine (approved):
🎉

PR #1917, opened 2023-06-03
Allow specific paths to be passed to js-api-spec

nex3:


Goodwine:
Was this not possible before? I recall doing this and it would run only those tests:
```
npm run js-api-spec -- \
  --sassSassRepo $SASS_SASS_PATH \
  --sassPackage $SASS_EMBEDDED_PATH \
  js-api-spec/legacy/importer
```

nex3:
Nope, currently it's hard-coded.

Goodwine (approved):

Goodwine:
> Nope, currently it's hard-coded.

Hm, I wonder why I can filter down to specific test files (not test cases tho) with the pattern I mentioned above 🤔

nex3:
Are you thinking of the language spec runner? Because I just tried the JS API spec runner on main and it definitely didn't restrict itself to the paths I passed on the CLI.

Goodwine:
We figured things out offline, I was in an older version, and I guess a newer version changed things a little so the way I mentioned doesn't work anymore without this PR

PR #1918, opened 2023-06-07
Add JS API tests for calculations

jerivas:
[skip dart-sass]
[skip sass-embedded]

https://github.com/sass/sass/pull/3622
https://github.com/sass/dart-sass/pull/1988
https://github.com/sass/embedded-host-node/pull/237


jerivas:
[I want JS users to have access to the calculation API.](https://trello.com/c/CCIn53GZ/2-i-want-js-users-to-have-access-to-the-calculation-api)

js-api-spec/function.test.ts
```diff
@@ -77,6 +82,129 @@ it('passes a default argument value', () => {
   expect(fn).toHaveBeenCalled();
 });
 
+describe('simplifies', () => {
+  it('operations', () => {
+    const fn = spy(
+      () =>
+        new CalculationOperation(
```
nex3, line 89: We shouldn't allow raw calculation operations (or interpolations) to be returned from functions. They aren't values in and of themselves, only when included within a `SassCalculation`.

js-api-spec/function.test.ts
```diff
@@ -77,6 +82,129 @@ it('passes a default argument value', () => {
   expect(fn).toHaveBeenCalled();
 });
 
+describe('simplifies', () => {
```
nex3, line 85: Nit: Since this is so calculation-specific, I'd expect it to be in the calculation test suite.

nex3 (changes requested):

js-api-spec/value/calculation.test.ts
```diff
@@ -0,0 +1,302 @@
...
+    it('correctly stores name and arguments', () => {
+      const result = SassCalculation.calc(new SassNumber(1));
+      expect(result.name).toBe('calc');
+      expect(result.arguments).toEqualWithHash(List([new SassNumber(1)]));
+    });
+
+    it('rejects invalid arguments', () => {
+      for (const value of invalidCalculationValues) {
+        // @ts-expect-error
+        expect(() => SassCalculation.calc(value)).toThrow();
```
jerivas, line 78: @nex3 Because `invalidCalculationValues` includes non-calculation values like `1` and `"1"`, TS complains about calling `calc()` with them. In this case I'm testing precisely that an exception is thrown at runtime, not just flagged by tsc.

This raises some questions:

- Is it right to have runtime validation of `CalculationValue`? Or should I just rely on TS to flag passing the wrong type of arguments?
- If runtime validation is desired, is this the right way to test for it?

nex3: As far as the spec is concerned, implementations aren't _required_ to have any particular failure state if callers violate the stated type declarations, so I'd leave off checks for anything that violates the TypeScript types.

js-api-spec/function.test.ts
```diff
@@ -226,14 +101,29 @@ describe('gracefully handles a custom function', () => {
...
-  it('returning a non-Value', () => {
-    expect(() =>
-      compileString('a {b: foo()}', {
-        functions: {
-          'foo()': (() => 'wrong') as unknown as CustomFunction<'sync'>,
-        },
-      })
-    ).toThrowSassException({line: 0});
+  describe('returning a non-Value', () => {
+    it('before simplification', () => {
```
nex3, line 105: "Before" and "after" simplification is a very implementation-focused way of describing this. I'd make it more focused on the caller's perspective, and say "directly" versus "in a calculation"

js-api-spec/value/calculation.test.ts
```diff
@@ -192,6 +177,17 @@ describe('SassCalculation', () => {
       }
     });
 
+    /* When `clamp()` is called with less than three arguments, the list of
+       accepted values is much narrower */
```
nex3, line 181: Nit: Use single-line comments for anything that's not JSDoc.

js-api-spec/function.test.ts
```diff
@@ -151,12 +169,51 @@ describe('asynchronously', () => {
...
+  it('simplifies', async () => {
+    const fn = async () =>
+      SassCalculation.calc(
+        new CalculationOperation('+', new SassNumber(1), new SassNumber(2))
+      );
+
+    const result = await compileStringAsync('a {b: foo()}', {
+      functions: {'foo()': fn},
+    });
+    expect(result.css).toBe('a {\n  b: 3;\n}');
```
nex3, line 216: I think this should also be in the calculation test file.

js-api-spec/value/calculation.test.ts
```diff
@@ -216,9 +212,146 @@ describe('SassCalculation', () => {
...
+    for (const [value, output] of primitiveValues) {
+      it(value.toString(), () => {
+        expect(
+          compileString('a {b: foo()}', {
+            functions: {'foo()': () => value},
+          }).css
+        ).toBe(`a {\n  b: ${output};\n}`);
+      });
+    }
+  });
```
nex3, line 314: I'm not sure how much value this test is adding. If it's trying to assert that all non-`Calculation`s are passed through unaltered, this doesn't cover all the Sass value types and would be annoying to keep up-to-date even if it did. Otherwise, it seems like the general principle that "a normal Sass value is returned as-is" should be exercised fairly thoroughly be the existing function tests.

nex3 (changes requested):

js-api-spec/value/calculation.test.ts
```diff
@@ -0,0 +1,424 @@
...
+// https://opensource.org/licenses/MIT.
+
+import {
+  Value,
+  SassCalculation,
+  SassNumber,
+  SassString,
+  CalculationOperation,
+  CalculationOperator,
+  CalculationInterpolation,
```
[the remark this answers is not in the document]
ntkme: @nex3 Any comments?

js-api-spec/value/calculation.test.ts
```diff
@@ -0,0 +1,424 @@
...
+
+    it("isn't any other type", () => {
+      expect(() => calculation.assertBoolean()).toThrow();
+      expect(() => calculation.assertColor()).toThrow();
+      expect(() => calculation.assertFunction()).toThrow();
+      expect(() => calculation.assertMap()).toThrow();
+      expect(calculation.tryMap()).toBe(null);
+      expect(() => calculation.assertNumber()).toThrow();
+      expect(() => calculation.assertString()).toThrow();
+    });
```
nex3, line 53: Can you add `assertCalculation()` to the "isn't any other type" tests for all the other types?

nex3 (approved):

js-api-spec/value/calculation.test.ts
jerivas: I think I did on 17c55ef9c064ccc29ebce1d6e33837b5699edd58, unless I missed something?

jerivas: Oh, but they were missing from the dart test suite. I added them in https://github.com/oddbird/dart-sass/commit/6b3b5b80708ce47fdc5af4b4611aa4239c497ad4 and https://github.com/sass/dart-sass/pull/1988/commits/52ab266749a3e6c3a18301afb3c9549592eace94

PR #1928, opened 2023-07-28
[Color 4] Add tests for `lab()`

nex3:
This still needs error output, but Dart Sass currently doesn't support
missing alpha channels or the is-missing() function so they can't be
generated yet.

[skip dart-sass]

### DIFF
--- a/spec/core_functions/color/lab/error.hrx
+++ b/spec/core_functions/color/lab/error.hrx
@@ -1,0 +1,259 @@
+<===> unit/lightness/input.scss
+a {b: lab(1px 2 3)}
+
+<===> unit/lightness/error
+Error: $lightness: Expected 1px to have no units or "%".
+  ,
+1 | a {b: lab(1px 2 3)}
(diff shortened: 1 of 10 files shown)


spec/core_functions/color/lab/error.hrx
```diff
@@ -0,0 +1,259 @@
+<===> unit/lightness/input.scss
+a {b: lab(1px 2 3)}
+
+<===> unit/lightness/error
+Error: $lightness: Expected 1px to have no units or "%".
```
Goodwine, line 5: Nit: unrelated to the change, but when I read "expected 1px have no units or %" It makes me think that the number shouldn't have units and that it also shouldn't have percent (some users might not think of percent as a unit).

So I wonder if that could be reworded (maybe on a separate PR?) to something like.. `Expected 1px to have no units or to have "%".` or `Expected 1px to have "%" or no units.` or something like that 🤔

Goodwine (approved):

Issue #1899, opened 2023-04-13
Add tests for additional calculation functions


nex3:
- [ ] Deprecation
- [ ] Final

See https://github.com/sass/sass/issues/3504

PR #1932, opened 2023-08-17
Add specs for deprecated function names

nex3:
See #1899
See sass/sass#3504

[skip dart-sass]

### DIFF
--- a/spec/directives/function/deprecated_name/unvendored.hrx
+++ b/spec/directives/function/deprecated_name/unvendored.hrx
@@ -1,0 +1,377 @@
+<===> options.yml
+:ignore_for:
+- libsass
+
+<===>
+================================================================================
+<===> round/input.scss
+@function round() {@return a}
+b {c: round()}
+
+<===> round/output.css
+b {
+  c: a;
+}
+
+<===> round/warning
+DEPRECATION WARNING on line 1, column 1 of input.scss: 
+Naming a Sass function "round" is deprecated because it conflicts with the new
+CSS round() syntax. Either rename it or use the math.round() function.
+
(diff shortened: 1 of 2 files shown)


spec/directives/function/deprecated_name/unvendored.hrx
```diff
@@ -0,0 +1,377 @@
...
+b {
+  c: a;
+}
+
+<===> round/warning
+DEPRECATION WARNING on line 1, column 1 of input.scss: 
+Naming a Sass function "round" is deprecated because it conflicts with the new
+CSS round() syntax. Either rename it or use the math.round() function.
+
+More info: https:/sass-lang.com/d/math-fn-name
```
Goodwine, line 21: Already commented on dart-sass, but this is missing a slash after `https:`

spec/directives/function/deprecated_name/unvendored.hrx
connorskees, line 1: Perhaps this case is considered too exceptional, but is it worth adding tests for functions which are renamed to these newly reserved functions?

For example:

```scss
// foo.scss
@use "bar" as *;

a {
  color: round();
}

// bar.scss
@forward "baz" as r*;

// baz.scss
@function ound() { @return 0 }
```

Which today will compile without error, but in the future will fail (as it currently does for a similar setup with e.g. `calc`). 

I had thought it was easier to rename identifiers for some reason -- i.e. without the use of `@forward`, but I can't think of any concrete way off the top of my head.

nex3:
Closing this out in favor of the approach described in https://github.com/sass/sass/pull/3661

PR #1933, opened 2023-08-25
add tests for first class mixins

### LINKED ISSUE #1927: Add tests for first-class mixins
See https://github.com/sass/sass/issues/626

connorskees:
This is some initial work at adding the baseline tests for first class mixins.

Embedded node host PR: https://github.com/sass/embedded-host-node/pull/253
Tracking issue: https://github.com/sass/sass/issues/626
Protobuf changes: https://github.com/sass/sass/pull/3674
dart-sass changes: sass/dart-sass#2073

[skip dart-sass]
[skip sass-embedded]

Closes #1927


connorskees:
It seems like, from some comments I read, that these tests are typically auto-generated? Is there documentation somewhere on how to generate these?

nex3:
The test cases are written by hand, but for error tests specifically the _output_ can be auto-generated. Just write test cases with empty error output:

```hrx
<===> test/input.scss
a {b: }

<===> test/error
```

then run `npm run sass-spec --interactive`. When it gets to one of these test cases, use `O!` to tell it to overwrite the output for all tests that have similar failures (in this case, "the error doesn't match the expected error output") to match the actual output generated by the test runner.

For bonus points, explicitly document this pattern in the README :slightly_smiling_face:.

spec/core_functions/meta/accepts_content.hrx
```diff
@@ -0,0 +1,77 @@
...
+:todo:
+- sass/libsass#2807
+
+<===>
+================================================================================
+<===> accepts/direct-child/input.scss
+@use "sass:meta";
+
+@mixin foo() {
+    @content;
```
nex3, line 12: [Use two spaces for indentation](https://github.com/sass/sass-spec/blob/main/STYLE_GUIDE.md#do-use-two-spaces-for-indentation)

spec/core_functions/meta/accepts_content.hrx
```diff
@@ -0,0 +1,77 @@
...
+---
+:todo:
+- sass/libsass#2807
+
+<===>
+================================================================================
+<===> accepts/direct-child/input.scss
+@use "sass:meta";
+
+@mixin foo() {
```
nex3, line 11: [Use single-letter names for irrelevant placeholders](https://github.com/sass/sass-spec/blob/main/STYLE_GUIDE.md#do-use-single-letter-names-for-irrelevant-placeholders)

spec/core_functions/meta/accepts_content.hrx
```diff
@@ -0,0 +1,77 @@
...
+<===> accepts/nested-child/output.css
+a {
+  b: true;
+}
+
+<===>
+================================================================================
+<===> accepts/builtin/input.scss
+@use "sass:meta";
+@mixin foo() {}
```
nex3, line 44: Why is this here?

spec/core_functions/meta/apply.hrx
```diff
@@ -0,0 +1,122 @@
+<===> options.yml
+---
+:todo:
+- sass/libsass#2807
+
+<===>
+================================================================================
+<===> use-as-function/input.scss
```
nex3, line 8: [Put error specs in a separate "error" directory](https://github.com/sass/sass-spec/blob/main/STYLE_GUIDE.md#do-put-error-specs-in-a-separate-error-directory)

spec/core_functions/meta/apply.hrx
nex3, line 1: None of these tests cover _successfully_ invoking `meta.apply`. I understand some of them are in the `meta.get-mixin()` tests, but I think anything about the complex application behaviors like argument handling belongs here.

spec/core_functions/meta/accepts_content.hrx
nex3, line 1: This needs error tests.

spec/core_functions/meta/accepts_content.hrx
```diff
@@ -0,0 +1,77 @@
...
+- sass/libsass#2807
+
+<===>
+================================================================================
+<===> accepts/direct-child/input.scss
+@use "sass:meta";
+
+@mixin foo() {
+    @content;
+}
```
nex3, line 13: [Prefer one-line rules when they contain a single child with no children](https://github.com/sass/sass-spec/blob/main/STYLE_GUIDE.md#prefer-one-line-rules-when-they-contain-a-single-child-with-no-children)

spec/core_functions/meta/get_mixin/content.hrx
```diff
@@ -0,0 +1,172 @@
+<===> passes-content/input.scss
+@use "sass:meta";
+
+@mixin foo {
+    @content;
+}
```
nex3, line 6: It's probably a good idea to have a wrapper style rule here to be clear that the `@content` isn't being passed through as-is.

spec/core_functions/meta/get_mixin/content.hrx
```diff
@@ -0,0 +1,172 @@
...
+
+<===>
+================================================================================
+<===> scope/redeclare-vars/input.scss
+@use "sass:meta";
+@mixin foo($a: a) {
+    $b: b;
+    @content;
+    a: $a;
+    b: $b;
```
nex3, line 86: [Use descriptive names for multiple placeholders](https://github.com/sass/sass-spec/blob/main/STYLE_GUIDE.md#do-use-descriptive-names-for-multiple-placeholders). For example, `$param` instead of `$a`, `$in-mixin` instead of `$b`, and so on. Similarly, name the values after where they're set so it's clear where they come from in the output.

spec/core_functions/meta/get_mixin/content.hrx
```diff
@@ -0,0 +1,172 @@
...
+  '
+  ,--> sass:meta
+1 | @mixin load-css($url, $with: null) {
+  |        =========================== declaration
+  '
+  input.scss 4:5  root stylesheet
+
+<===>
+================================================================================
+<===> scope/redeclare-vars/input.scss
```
nex3, line 80: This should probably be nested in the `redeclare/` directory you define below instead of being a sibling directory.

spec/core_functions/meta/get_mixin/different_module.hrx
```diff
@@ -0,0 +1,135 @@
...
+}
+
+<===>
+================================================================================
+<===> chosen_prefix/input.scss
+@use "sass:meta";
+@use "foo" as a;
+b {@include meta.apply(meta.get-mixin("foo", $module: "a"), #abcdef)}
+
+<===> chosen_prefix/foo.scss
```
nex3, line 23: [Prefer making imported files partials](https://github.com/sass/sass-spec/blob/main/STYLE_GUIDE.md#prefer-making-imported-files-partials) and [name single imported or used files "other"](https://github.com/sass/sass-spec/blob/main/STYLE_GUIDE.md#do-name-single-imported-or-used-files-other)

spec/core_functions/meta/get_mixin/equality.hrx
```diff
@@ -0,0 +1,74 @@
...
+a {
+  b: true;
+}
+
+<===>
+================================================================================
+<===> built_in/different/input.scss
+@use "sass:meta";
+@mixin foo() {}
+a {b: meta.get-mixin(load-css, meta) == meta.get-mixin(foo)}
```
nex3, line 28: This should compare one built-in mixin to a different one.

spec/core_functions/meta/get_mixin/error.hrx
```diff
@@ -0,0 +1,262 @@
...
+1 | @function get-mixin($name, $module: null) {
+  |           =============================== declaration
+  '
+  input.scss 2:7  root stylesheet
+
+<===>
+================================================================================
+<===> argument/too_many/input.scss
+@use "sass:meta";
+a {b: meta.get-mixin(c, true, d, e)}
```
nex3, line 66: This should be a valid invocation other than the extra arguments.

spec/core_functions/meta/get_mixin/error.hrx
```diff
@@ -0,0 +1,262 @@
...
+@mixin bar() {}
+a {b: meta.get-mixin(foo) / meta.get-mixin(bar)}
+
+<===> division/error
+Error: get-mixin("foo") isn't a valid CSS value.
+  ,
+4 | a {b: meta.get-mixin(foo) / meta.get-mixin(bar)}
+  |       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+  '
+  input.scss 4:7  root stylesheet
```
nex3, line 154: This should probably be in `spec/values/mixins` since it's really testing the behavior of the mixin type rather than the `meta.get-mixin` function. Also, test the other operations.

spec/core_functions/meta/get_mixin/error.hrx
```diff
@@ -0,0 +1,262 @@
...
+  error: meta.get-mixin($add-two-mixin);
+}
+
+<===> mixin_exists/error
+Error: $name: get-mixin("add-two") is not a string.
+  ,
+9 |   error: meta.get-mixin($add-two-mixin);
+  |          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+  '
+  input.scss 9:10  root stylesheet
```
nex3, line 176: Isn't this the same as `argument/mixin_ref`?

spec/core_functions/meta/get_mixin/meta.hrx
nex3, line 1: These should go in the `meta.inspect` and `meta.type-of` tests.

spec/core_functions/meta/get_mixin/scope.hrx
```diff
@@ -0,0 +1,94 @@
...
+}
+
+<===> captures_inner_scope/output.css
+.scope1 .scope2 .scope3 {
+  a: 12;
+}
+
+<===>
+================================================================================
+<===> scope/doesnt-capture-global/input.scss
```
nex3, line 51: [Prefer underscore-separated paths](https://github.com/sass/sass-spec/blob/main/STYLE_GUIDE.md#prefer-underscore-separated-paths). Also, this isn't really accurate—it does capture the global scope (if you pass the mixin reference to a different module it should still refer to the same `$a` value), it's just that the global variable slot it refers to is mutated after the mixin is defined.

spec/core_functions/meta/get_mixin/scope.hrx
```diff
@@ -0,0 +1,94 @@
...
+}
+
+<===> scope/doesnt-capture-global/output.css
+a {
+  a: y;
+}
+
+<===>
+================================================================================
+<===> scope/doesnt-capture-local/input.scss
```
nex3, line 74: Same issue here.

spec/core_functions/meta/module_mixins.hrx
```diff
@@ -0,0 +1,372 @@
+<===> options.yml
+---
+:todo:
+- sass/libsass#2807
+
+<===> _util.scss
```
nex3, line 6: We usually name these files "utils" (plural)

spec/core_functions/meta/module_mixins.hrx
```diff
@@ -0,0 +1,372 @@
...
+<===> empty/output.css
+a {
+  b: ();
+}
+
+<===>
+================================================================================
+<===> multiple/input.scss
+@use "sass:meta";
+@use "../util";
```
nex3, line 37: It's more reliable to refer to this as `core_functions/meta/utils`, since then you can freely change the path of the test case without updating this relative path.

spec/core_functions/meta/module_mixins.hrx
nex3, line 1: Add a test that verifies that the object returned is actually a map, and not just a list of pairs.

nex3 (changes requested):

spec/core_functions/meta/accepts_content.hrx
connorskees: Resolved. It would be nice if `lint-spec` enforced this or if `lint-spec --fix` fixed this -- it was somewhat tedious to grep for `    ` to find all places with the wrong indentation.

spec/core_functions/meta/apply.hrx
nex3, line 1: Also test passing through rest arguments, especially when they contain keywords.

spec/core_functions/meta/get_mixin/content.hrx
```diff
@@ -0,0 +1,170 @@
...
+<===> denies_content/user_defined/error
+Error: Mixin doesn't accept a content block.
+    ,
+2   | @mixin a {}
+    |        = declaration
+... |
+5   |   @include meta.apply(meta.get-mixin("a")) {}
+    |   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ invocation
+    '
+  input.scss 5:3  root stylesheet
```
nex3, line 53: This and other error tests should be in an `error/` directory.

spec/core_functions/meta/get_mixin/content.hrx
```diff
@@ -0,0 +1,170 @@
...
+
+$global: global;
+
+a {
+  $in-style-rule: in-style-rule;
+  @include meta.apply(meta.get-mixin("a")) {
+    $param: x;
+    $in-mixin: x;
+    $global: x;
+    $in-style-rule: x;
```
nex3, line 97: Instead of `x`, name these `in-include` to indicate where the values came from.

spec/core_functions/meta/get_mixin/content.hrx
```diff
@@ -0,0 +1,170 @@
...
+  @include meta.apply(meta.get-mixin(baz)) using ($a) {
+    a: $a;
+  }
+}
+
+<===> scope/fall_through/output.css
+a {
+  a: x;
+  b: x;
+}
```
nex3, line 170: Use semantic placeholders in these tests as well.

nex3 (changes requested):
Almost there!

spec/core_functions/meta/accepts_content.hrx
```diff
@@ -0,0 +1,140 @@
...
+---
+:todo:
+- sass/libsass#2807
+
+<===>
+================================================================================
+<===> accepts/direct_child/input.scss
+@use "sass:meta";
+
+@mixin a() {@content;}
```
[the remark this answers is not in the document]
connorskees: I made this change, but I'll mention that I had intentionally left this semicolon because syntax highlighting (at least in vscode) breaks for Sass files -- and hrx files containing Sass -- when `@content` isn't followed by a semicolon.

image

image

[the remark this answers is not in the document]
nex3: That sounds like a bug in the tooling.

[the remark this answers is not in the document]
connorskees: It is, and it's quite annoying. I know it's bad to change the contents of the spec to workaround the bad tooling, but I assume most people use vscode to edit these files, and in some cases the syntax highlighting gets so bad that it's better to just turn it off. There's a number of different cases like this in vscode for all Sass files,

image
image

The syntax highlighting error shown here is pretty minimal, but it demonstrates the bug. `@content` without a semicolon isn't the only similar issue where there's valid Sass syntax and the editor will either red underline as a syntax error or just completely break down while doing syntax highlighting for the entire file.

I started a bit on an LSP implementation for Sass using my compiler and some of rust-analyzer's internals.

spec/core_functions/meta/get_mixin/content.hrx
```diff
@@ -0,0 +1,168 @@
...
+Error: Mixin doesn't accept a content block.
+  ,--> input.scss
+4 |   @include meta.apply(meta.get-mixin(load-css, meta), "a") {
+  |   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ invocation
+  '
+  ,--> sass:meta
+1 | @mixin load-css($url, $with: null) {
+  |        =========================== declaration
+  '
+  input.scss 4:3  root stylesheet
```
nex3, line 74: Error tests should be at the end of HRX files.

spec/core_functions/meta/accepts_content.hrx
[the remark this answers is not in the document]
connorskees: I wonder if this particular bug could be fixed by changing [this regex](https://github.com/microsoft/vscode/blob/d9c889a0f762cb954875eb29f5c9df2a7df8da30/extensions/scss/syntaxes/scss.tmLanguage.json#L108) to recognize `[;\}]` instead of just `;`.

nex3 (changes requested):

nex3 (approved):

spec/core_functions/meta/accepts_content.hrx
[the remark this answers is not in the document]
connorskees: The above change seems to work. https://github.com/microsoft/vscode/pull/194363

ntkme:
Can we add a js-api test `can round-trip a mixin reference from Sass` similar to this: https://github.com/sass/sass-spec/blob/dff3578d546da96fa536e06f96066c9edcd3612c/js-api-spec/value/function.test.ts#L9-L31

ntkme:
Also js-api tests for `assertMixin`.

connorskees:
Do we have JS API tests today that read a value from Sass? The function tests (and all the tests I've looked at so far) seem to involve declaring a function/another value in JavaScript and sending it to Sass, which is not possible with mixins.

ntkme:
> Do we have JS API tests today that read a value from Sass? The function tests (and all the tests I've looked at so far) seem to involve declaring a function/another value in JavaScript and sending it to Sass, which is not possible with mixins.

The sass function test I linked declare a sass function `plusOne` in sass code, pass it to the host via a host function `foo` and then the host function `foo` return the compiler function `plusOne` back to compiler, which then get invoked.

js-api-spec/value/mixin.test.ts
```diff
@@ -0,0 +1,32 @@
...
+    return args[0];
+  });
+
+  expect(
+    compileString(
+      `
+      @use 'sass:meta';
+
+      @mixin a() {}
+      a {b: meta.inspect(foo(meta.get-mixin('a')))}
```
ntkme, line 23: Maybe we can add content to `@mixin a()`, call `meta.apply()`, and then validate the content is in the output?

ntkme:
You should add `[skip sass-embedded]` to this PR's description, and then trigger test from dart-sass or embedded-host-node repo if needed.

ntkme:
Existing sass values has some test regarding `assert` like:

https://github.com/sass/sass-spec/blob/dff3578d546da96fa536e06f96066c9edcd3612c/js-api-spec/value/argument-list.test.ts#L98-L107
https://github.com/sass/sass-spec/blob/dff3578d546da96fa536e06f96066c9edcd3612c/js-api-spec/value/boolean.test.ts#L27-L36

We should add `assertMixin` to these tests.

js-api-spec/value/boolean.test.ts
```diff
@@ -32,6 +32,7 @@ describe('Sass boolean', () => {
       expect(value.tryMap()).toBe(null);
       expect(value.assertNumber).toThrow();
       expect(value.assertString).toThrow();
+      expect(value.assertMixin).toThrow();
```
ntkme, line 35: Nit: put `assertMixin` after `tryMap` and before `assertNumber` so that these are ordered alphabetically.

nex3:
What was wrong with the assertion in 55fe148bc52489a67691114729b629831ef07c5c?

connorskees:
I'm not sure -- I was somewhat hoping for a subsequent CI run to validate that the later checks are true to get a better idea.

This test failed in https://github.com/sass/dart-sass/actions/runs/6399530354/job/17372158557, only on the browser environment. I am expecting reproducing the run in the browser environment locally to be hard.

connorskees:
It seems to pass with the assertion removed, so it _does_ correctly execute `assertMixin`. It just isn't an `instanceof SassMixin` in the browser? I'm not familiar enough with the end-to-end process here to know what would be causing this and why it only fails in the browser environment. My assumption at least is that it's not a bug.

nex3:
Every Sass value object is expected to be `instanceof` its type—this is part of the API contract we provide to the user, so if it's not passing somewhere that's a bug that needs to be fixed.

nex3:
We use `isInstanceOf()` everywhere else, why wouldn't it work here?

connorskees:
In this particular case, I changed the check from `toBeInstanceOf` to a bare `instanceof` check to get a better sense of the issue from the CI run. I would like for the check to eventually be `toBeInstanceOf`. I meant proper in the commit message in the sense of "`instanceof` proper".

I was hoping to be able to debug without having to download the chrome executable, but I suppose I will have to.

PR #1934, opened 2023-09-01
Re-enable calc function specs

### LINKED ISSUE #1899: Add tests for additional calculation functions

nex3:
See sass/sass#3504
See #1899

[skip dart-sass]


spec/values/calculation/calc/operator.hrx
```diff
@@ -328,42 +328,52 @@ a {
...
-<===> precedence/interpolation/parens/rhs/input.scss
-a {b: calc(1 + (#{c}))}
+<===> precedence/interpolation/calculation/plain/input.scss
+a {b: calc(calc(#{c}))}
 
-<===> precedence/interpolation/parens/rhs/output.css
+<===> precedence/interpolation/calculation/plain/output.css
 a {
-  b: calc(1 + (c));
+  b: calc(c);
```
Goodwine, line 346: I find it slightly confusing that `calc(calc(#{c}))` produces `calc(c)` without parentheses.
`calc(#{1px + 2px})` produces parentheses `calc((3px))`. Same with the test immediately above.

It's not a big deal, but as a user I might be tempted to write a pre-processing postcss script that wraps all calcs with another calc to produce smaller code 😅

Goodwine (approved):
tests LGTM

spec/values/calculation/calc/operator.hrx
nex3: `calc(#{1px + 2px})` returns `calc(3px)`. Parentheses only get added for interpolation if there were parentheses in the original syntax.

If people don't like this behavior, the best way to handle it is to avoid using interpolation in calculations at all.

Goodwine: I see, I misunderstood the test before, now I understand the parenthesis are added in the test above this one because `c` is an unquoted string :)

Goodwine (approved):

PR #1940, opened 2023-09-27
Add specs for the support of relative CSS colors

stof:
https://github.com/sass/sass/pull/3673

[skip dart-sass]

### DIFF
--- a/spec/core_functions/color/hsl/one_arg/relative_color.hrx
+++ b/spec/core_functions/color/hsl/one_arg/relative_color.hrx
@@ -1,0 +1,140 @@
+<===> static/alpha/input.scss
+a {b: hsl(from #aaa h s l / 25%)}
+
+<===> static/alpha/output.css
+a {
+  b: hsl(from #aaa h s l/25%);
+}
+
+<===>
+================================================================================
+<===> static/no_alpha/input.scss
+a {b: hsl(from #aaa h s l)}
+
+<===> static/no_alpha/output.css
+a {
+  b: hsl(from #aaa h s l);
+}
+
+<===>
+================================================================================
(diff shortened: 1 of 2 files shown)


spec/core_functions/color/hsl/one_arg/relative_color.hrx
```diff
@@ -0,0 +1,125 @@
...
+<===> var/alpha/output.css
+a {
+  b: hsl(from var(--c) h s l/25%);
+}
+
+<===>
+================================================================================
+<===> var/no_alpha/input.scss
+a {b: hsl(from var(--c) h s l)}
+<===> var/no_alpha/output.css
```
nex3, line 53: Nit: [end each file with a newline](https://github.com/sass/sass-spec/blob/main/STYLE_GUIDE.md#do-end-each-file-with-a-newline)

nex3 (commented):
Please also add tests for slash lists created with `list.slash()`.

spec/core_functions/color/hsl/one_arg/relative_color.hrx
stof: looks like the spec linter does not enforce this right now.

stof:
@nex3 I added a test for them for each function.

spec/core_functions/color/hsl/one_arg/relative_color.hrx
nex3: The spec linter is unfortunately not as comprehensive as we would like it to be, but it's hard to allocate time to it instead of more pressing concerns.

nex3 (approved):

PR #1949, opened 2023-10-13
[Color 4] JS API specs

jamesnw:
Spec: https://github.com/sass/sass/blob/main/proposal/color-4-new-spaces-js.d.ts.md

- https://github.com/sass/sass/pull/3718
- https://github.com/sass/dart-sass/pull/2117

[skip dart-sass]
[skip sass-embedded]


jgerigmeyer:
[I want JS users to have access to new color spaces.](https://trello.com/c/2R7Mnj6s/1-i-want-js-users-to-have-access-to-new-color-spaces)

jgerigmeyer:
@nex3 A number of the tests seem to be failing on CI because `dart run grinder pkg-npm-dev` pulls in the `main` branch of the `sass/sass` repo -- is there a way to tell CI to build using the dart-sass and sass PRs linked in this PR description?

jgerigmeyer:
> @nex3 A number of the tests seem to be failing on CI because `dart run grinder pkg-npm-dev` pulls in the `main` branch of the `sass/sass` repo -- is there a way to tell CI to build using the dart-sass and sass PRs linked in this PR description?

We should add `[skip dart-sass]` and `[skip dart-sass]` to this PR.

js-api-spec/value/color.test.ts
```diff
@@ -26,22 +49,1008 @@ function hsl(
...
+    ],
+  },
+};
+
+type InterpolationExample = [
+  {weight: number; method?: string},
+  [number, number, number]
+];
+
+const interpolations: {[space: string]: InterpolationExample[]} = {
```
nex3, line 573: Where are you sourcing these examples? If they're from a w3c source, we should link it.

nex3 (commented):
The general approach here of declaratively defining information about each space and using that to reduce duplication looks good to me. I'd prefer to have all of the looping through the declarations _outside_ test cases, so each test case still ends up testing only one thing.

js-api-spec/value/color.test.ts
jamesnw: I generated them using Color.js, with the code living in a [branch in Oddbird's fork](https://github.com/oddbird/color.js/blob/sass-interpolations/apps/mixes/mixes.js) of Color.js due to some internal changes to Color.js I needed to make at the time to support the Hue interpolation method. 

I didn't write the generation code planning for it to be incorporated into sass-spec, but I'd be open to that if you have ideas on how to do that.

jamesnw:
> I'd prefer to have all of the looping through the declarations _outside_ test cases, so each test case still ends up testing only one thing.

I made this change in some places, but didn't do it for things like loops through channels. As is, I went from 580 cases to 1800+, but there may be other places where specificity is worth it.

js-api-spec/value/color.test.ts
nex3: I meant more, what caused you to choose these specific colors to test against rather than any others?

jamesnw: For starting colors, I picked 2 of Oddbird's brand colors, not for any particular reason about the colors themselves. The interpolation examples are basically between those colors with weights [0, 0.5, 1], and then for polar spaces, each hue interpolation method. My thought was to verify a range of options, as opposed to checking specific edge cases, but the checks could easily be expanded.

js-api-spec/value/color.test.ts
nex3, line 1: This file is getting very large—could you split it up into a `color/` directory, maybe on color space boundaries?

nex3 (changes requested):
Just a few style changes and I think we're ready to go!

js-api-spec/value/color.test.ts
```diff
@@ -431,3 +1674,536 @@ describe('SassColor', () => {
...
+              expect(res.space).toBe(destinationSpace.name);
+
+              const expected = destinationSpace.constructor(
+                ...destinationSpace.pink
+              );
+              expect(res).toEqualWithHash(expected);
+            });
+          });
+        });
+      });
```
jgerigmeyer, line 1758: @nex3 Using ColorJS to do color conversions works, but the math isn't precise enough to pass these tests, e.g.:

```
expected color(xyz-d50 0.66406985168935 0.5367266828656 0.43459579340121) to be .equal() to color(xyz-d50 0.6640698533004 0.53672666252811 0.43459582467203)
```

I'm not sure if we want to relax the precision requirements, or have different precision requirements for different platforms, or...?

js-api-spec/value/color.test.ts
jamesnw: I split this out in https://github.com/sass/sass-spec/pull/1949/commits/dd6d11b9e5f0249664ff1d4bcc35000cbd2a28d3, opting to still loop through the color spaces, but each file has a more limited scope.

js-api-spec/value/color.test.ts
jgerigmeyer: Resolved in https://github.com/sass/sass-spec/pull/1949/commits/d5d18b6a3d704f1277ebcbff14161dc22456f82c

js-api-spec/value/color/color-4-channels.test.ts
```diff
@@ -0,0 +1,295 @@
...
+          const pinkCases = channelCases(...space.pink);
+          pinkCases.forEach(channels => {
+            const _color = space.constructor(...channels);
+            expect(_color.channelsOrNull).toFuzzyEqualList(
+              channels.slice(0, 3)
+            );
+          });
+        });
+      });
+      describe('channels', () => {
```
nex3, line 43: Style nit: please leave an empty line between test cases and groups.

nex3 (approved):

js-api-spec/value/color/color-4-channels.test.ts
jamesnw: Addressed in https://github.com/sass/sass-spec/pull/1949/commits/64d225a9e2ff5d5c669cfba09fd341d898bb82f7.

PR #1952, opened 2023-10-30
[Package Importer] Specs

jamesnw:
https://github.com/sass/sass/issues/2739

Blocked until [proposal](https://github.com/sass/sass/blob/main/proposal/package-importer.d.ts.md) is accepted.

- [ ] sass - https://github.com/sass/sass/pull/3728
- [ ] dart-sass - https://github.com/sass/dart-sass/pull/2130
- [ ] embedded-host-node - https://github.com/sass/embedded-host-node/pull/260

[skip dart-sass]
[skip sass-embedded]


js-api-spec/legacy/importer.node.test.ts
```diff
@@ -35,7 +35,29 @@ it('an empty object means an empty file', () =>
 
 describe('import precedence:', () => {
   describe('in sandbox dir', () => {
-    it('relative file is #1', () =>
+    it('packageImporter:node is #1', () =>
```
nex3, line 38: This isn't relevant to the import precedence that this `describe()` is testing. These tests verify what happens when there are multiple possible interpretations of the same load. Because the package importer is always absolute (since it has the `pkg:` scheme), it's never ambiguous with relative, load path, or working-directory loads.

js-api-spec/importer.test.ts
```diff
@@ -762,6 +763,15 @@ it('throws an ArgumentError when the result sourceMapUrl is missing a scheme', (
     includes: 'Invalid argument (sourceMapUrl): must be absolute',
   });
 });
+runOnlyForImpl('browser', () => {
```
nex3, line 766: Empty line between test cases.

js-api-spec/sandbox.ts
```diff
@@ -49,13 +49,21 @@ export async function sandbox(
...
-        chdir: (callback: () => unknown) => {
+        chdir: (
+          callback: () => unknown,
+          options?: {changeEntryPoint: string}
+        ) => {
           const oldPath = process.cwd();
           process.chdir(testDir);
+          const oldEntryPoint = require.main?.filename;
+          if (options?.changeEntryPoint)
+            require.main!.filename = `${testDir}/${options.changeEntryPoint}`;
```
nex3, line 60: Style nit: Use curly braces if the conditional doesn't fit on one line.

js-api-spec/node-package-importer.node.test.ts
```diff
@@ -0,0 +1,681 @@
...
+  nodePackageImporter,
+  NodePackageImporter,
+  LegacyException,
+  LegacyResult,
+  FileImporter,
+} from 'sass';
+
+import {sandbox, SandboxDirectory} from './sandbox';
+
+const fileImporter = (dir: SandboxDirectory): FileImporter => ({
```
nex3, line 21: Document all members.

js-api-spec/node-package-importer.node.test.ts
```diff
@@ -0,0 +1,681 @@
...
+  return JSON.stringify({
+    name: 'foo',
+    version: '1.0.0',
+    description: 'Verifying proposed pkg: loading algorithm',
+    scripts: {
+      test: 'echo "Error: no test specified" && exit 1',
+    },
+    author: 'Package Author',
+    license: 'ISC',
+    ...overrides,
```
nex3, line 35: Are all of these fields really necessary for the test?

js-api-spec/node-package-importer.node.test.ts
```diff
@@ -0,0 +1,681 @@
...
+  FileImporter,
+} from 'sass';
+
+import {sandbox, SandboxDirectory} from './sandbox';
+
+const fileImporter = (dir: SandboxDirectory): FileImporter => ({
+  findFileUrl: file => dir.url(file),
+});
+
+function manifestBuilder(overrides?: {[key: string]: string | Object}) {
```
nex3, line 25: Functions should generally have verb names, like `buildManifest()`

js-api-spec/node-package-importer.node.test.ts
```diff
@@ -0,0 +1,681 @@
...
+      '.': {
+        [key]: './src/sass/_styles.scss',
+      },
+      './colors/index.scss': {
+        [key]: './src/sass/colors/index.scss',
+      },
+      './_variables.scss': {
+        [key]: './src/sass/_variables.scss',
+      },
+    },
```
nex3, line 51: It feels weird to me that this is exporting files that don't necessarily exist in all (or most) tests. I think it might be clearer to just write the necessary exports by hand. Less indirection can make the tests easier to understand as well.

js-api-spec/node-package-importer.node.test.ts
```diff
@@ -0,0 +1,681 @@
...
+            () => {
+              const result = compileString('@use "pkg:foo";', {
+                importers: [nodePackageImporter],
+              });
+              expect(result.css).toEqualIgnoringWhitespace('a {b: c;}');
+            },
+            {changeEntryPoint: 'index.js'}
+          );
+        }));
+      it(`${key} at root without non-pathed exports`, () =>
```
nex3, line 76: The double negative of "without non-pathed" confuses me a bit. What exactly is a "pathed export"? Is the example below a "pathed export"? Why?

js-api-spec/node-package-importer.node.test.ts
```diff
@@ -0,0 +1,681 @@
...
+            },
+            {changeEntryPoint: 'index.js'}
+          );
+        }));
+
+      it(`${key} with subpath`, () =>
+        sandbox(dir => {
+          dir.write({
+            'node_modules/foo/src/sass/_styles.scss': 'd {e: f}',
+            'node_modules/foo/src/sass/_variables.scss': 'a {b: c}',
```
nex3, line 101: Calling this `_variables` when it doesn't contain any variables is a bit confusing. We often use "other" in tests when there are two different stylesheets that might be loaded.

js-api-spec/node-package-importer.node.test.ts
```diff
@@ -0,0 +1,681 @@
...
+            },
+            {changeEntryPoint: 'index.js'}
+          );
+        }));
+      it(`${key} with index`, () =>
+        sandbox(dir => {
+          dir.write({
+            'node_modules/foo/src/sass/_styles.scss': 'd {e: f}',
+            'node_modules/foo/src/sass/_variables.scss': 'g {h: i}',
+            'node_modules/foo/src/sass/colors/index.scss': 'a {b: c}',
```
nex3, line 121: Same issue as above with "colors" here. Maybe `subdir` instead to communicate the aspect of the difference that's relevant to this test.

js-api-spec/node-package-importer.node.test.ts
```diff
@@ -0,0 +1,681 @@
...
+              expect(result.css).toEqualIgnoringWhitespace('a {b: c;}');
+            },
+            {changeEntryPoint: 'index.js'}
+          );
+        }));
+
+      it(`${key} with subpath`, () =>
+        sandbox(dir => {
+          dir.write({
+            'node_modules/foo/src/sass/_styles.scss': 'd {e: f}',
```
nex3, line 100: Why even include styles here? It's not realistically going to be loaded for an import with "variables" in the name.

js-api-spec/node-package-importer.node.test.ts
```diff
@@ -0,0 +1,681 @@
...
+            {changeEntryPoint: 'index.js'}
+          );
+        }));
+    });
+    it('compiles with first conditional match found', () =>
+      sandbox(dir => {
+        dir.write({
+          'node_modules/foo/src/sass/_styles.scss': 'd {e: f}',
+          'node_modules/foo/src/sass/_variables.scss':
+            'a {from: sassCondition}',
```
nex3, line 142: I like giving this a meaningful `from:` property, but you should also give one to the other file—otherwise you don't get any useful debugging information from the file context if the test starts failing.

js-api-spec/node-package-importer.node.test.ts
```diff
@@ -0,0 +1,681 @@
...
+          });
+          dir.chdir(
+            () => {
+              const result = compileString('@use "pkg:foo";', {
+                importers: [nodePackageImporter],
+              });
+              expect(result.css).toEqualIgnoringWhitespace('a {b: c;}');
+            },
+            {changeEntryPoint: 'index.js'}
+          );
```
nex3, line 74: This block gets repeated many times with very few modifications. I recommend factoring it into a helper function.

js-api-spec/node-package-importer.node.test.ts
```diff
@@ -0,0 +1,681 @@
...
+            ),
+          });
+          dir.chdir(
+            () => {
+              const result = compileString('@use "pkg:foo";', {
+                importers: [nodePackageImporter],
+              });
+              expect(result.css).toEqualIgnoringWhitespace('a {b: c;}');
+            },
+            {changeEntryPoint: 'index.js'}
```
nex3, line 73: Should chdir just change the entry point by default?

js-api-spec/node-package-importer.node.test.ts
```diff
@@ -0,0 +1,681 @@
...
+              expect(
+                compileString('@use "pkg:foo/sass/variables";', {
+                  importers: [nodePackageImporter],
+                }).css
+              ).toEqualIgnoringWhitespace('a {b: c;}');
+            },
+            {changeEntryPoint: 'index.js'}
+          );
+        }));
+      it('throws if multiple wildcard export match', () =>
```
nex3, line 266: "exports"

js-api-spec/node-package-importer.node.test.ts
```diff
@@ -0,0 +1,681 @@
...
+            expect(
+              compileString('@use "pkg:foo";', {
+                importers: [nodePackageImporter],
+              }).css
+            ).toEqualIgnoringWhitespace('a {b: c;}');
+          },
+          {changeEntryPoint: 'index.js'}
+        );
+      }));
+    describe('wildcards', () => {
```
nex3, line 208: Probably also worth testing that something like `exports: {'./*': {sass: './src/sass/*'}}` works.

js-api-spec/node-package-importer.node.test.ts
```diff
@@ -0,0 +1,681 @@
...
+          },
+          {changeEntryPoint: 'index.js'}
+        );
+      }));
+    it('style key in package.json', () =>
+      sandbox(dir => {
+        dir.write({
+          'node_modules/foo/src/sass/_styles.scss': 'a {b: c}',
+          'node_modules/foo/package.json': manifestBuilder({
+            sass: 'src/sass/_styles.scss',
```
nex3, line 315: Is this supposed to be `style:`?

js-api-spec/node-package-importer.node.test.ts
```diff
@@ -0,0 +1,681 @@
...
+        dir.chdir(
+          () => {
+            const result = compileString('@use "pkg:bah";', {
+              importers: [nodePackageImporter],
+            });
+            expect(result.css).toEqualIgnoringWhitespace('a {b: c;}');
+          },
+          {changeEntryPoint: 'index.js'}
+        );
+      }));
```
nex3, line 402: Isn't this the same as one of the tests above?

js-api-spec/node-package-importer.node.test.ts
```diff
@@ -0,0 +1,681 @@
...
+    it('resolves most proximate node_module', () =>
+      sandbox(dir => {
+        dir.write({
+          'node_modules/bah/index.scss': '@use "pkg:bar";',
+          'node_modules/bah/package.json': manifestBuilder(),
+          'node_modules/bar/index.scss': 'e {from: root}',
+          'node_modules/bar/package.json': manifestBuilder(),
+          'node_modules/bah/node_modules/bar/index.scss': 'a {b: c}',
```
nex3, line 445: As above, give this a meaningful `from:`

js-api-spec/node-package-importer.node.test.ts
```diff
@@ -0,0 +1,681 @@
...
+          () => {
+            const result = compileString('@use "pkg:bar";', {
+              importers: [nodePackageImporter],
+            });
+            return expect(result.css).toEqualIgnoringWhitespace('a {b: c;}');
+          },
+          {changeEntryPoint: 'deeply/nested/file/index.js'}
+        );
+      }));
+    it('continues if no match found', () => {
```
nex3, line 492: It's odd that this test name says "continues" but it's actually verifying that the test fails.

nex3 (changes requested):

js-api-spec/legacy/importer.node.test.ts
jamesnw: That makes sense. I've removed this test here. I added this as the spec specifies that Package Importers are resolved first for the legacy API. Do we need to test that, or perhaps, is there really a way to test that, as there can't be conflicts?

js-api-spec/node-package-importer.node.test.ts
jamesnw: I realized this abstraction wasn't very used or useful, so I removed it.

js-api-spec/node-package-importer.node.test.ts
jamesnw: I did this in f96ba8dea03b2bd0197cde036582c3df1dbbd09e. I found an object with named parameters was slightly more verbose, but much more parseable than positional arguments.

js-api-spec/node-package-importer.node.test.ts
```diff
@@ -0,0 +1,580 @@
...
+        output: 'a {b: c;}',
+        files: {
+          'node_modules/bah/index.scss': '@use "pkg:bar";',
+          'node_modules/bah/package.json': JSON.stringify({}),
+          'node_modules/bar/index.scss': 'a {b: c}',
+          'node_modules/bar/package.json': JSON.stringify({}),
+        },
+      }));
+
+    it('relative import in package', () =>
```
jamesnw, line 326: This is a failing test that reflects the setup that exists in Bootstrap, and exposes an issue that's puzzling me a bit. 

It is able to find the package `foo`, but from `node_modules/foo/scss/foo.scss`, `@use "mixins/banner";` is unable to find the path `node_modules/foo/scss/mixins/_banner.scss`. 

I tested to confirm this isn't due to partials or extension matching. The Node Package Importer seems to be impacting the subsequent resolution of relative paths. I suspect there's a breaking issue in dart-sass around the ImporterCache.

js-api-spec/node-package-importer.node.test.ts
```diff
@@ -0,0 +1,580 @@
...
+          },
+        }));
+
+      it(`${key} with index`, () =>
+        testPackageImporter({
+          input: '@use "pkg:foo/subdir";',
+          output: 'a {b: c;}',
+          files: {
+            'node_modules/foo/src/sass/_styles.scss': 'd {e: f}',
+            'node_modules/foo/src/sass/_other.scss': 'g {h: i}',
```
nex3, line 94: I don't think these inputs are necessary here.

js-api-spec/node-package-importer.node.test.ts
```diff
@@ -0,0 +1,580 @@
...
+          output: 'a {b: c;}',
+          files: {
+            'node_modules/foo/src/sass/_other.scss': 'a {b: c}',
+            'node_modules/foo/package.json': JSON.stringify({
+              exports: {'./*.scss': './src/sass/*.scss'},
+            }),
+          },
+        }));
+
+      it('resolves with full wildcard path', () =>
```
nex3, line 173: I'd mention that this is about the combination of a wildcard _and_ a `sass:` key.

js-api-spec/node-package-importer.node.test.ts
```diff
@@ -0,0 +1,580 @@
...
+            }),
+          },
+        }));
+
+      it(`${key} with subpath`, () =>
+        testPackageImporter({
+          input: '@use "pkg:foo/other";',
+          output: 'a {b: c;}',
+          files: {
+            'node_modules/foo/src/sass/_other.scss': 'a {b: c}',
```
nex3, line 77: For tests with only one stylesheet, it probably makes sense to just name it "styles"

js-api-spec/node-package-importer.node.test.ts
nex3: That definitely seems like a bug, glad you caught it in a test. Here's what _should_ be happening:

* `pkg:foo` gets passed to `NodePackageImporter.canonicalize()`, which returns `file://.../node_modules/foo/scss/foo.scss`.
* `foo.scss` gets loaded with that canonical URL and `NodePackageImporter` as the current importer to handle relative URLs.
* The import cache resolves `mixins/banner` relative to that canonical URL to return `file://.../node_modules/foo/scss/mixins/banner`.
* That resolved URL then gets passed to `NodePackageImporter.canonicalize()`, which sees that it's a `file:` URL and passes it to `FilesystemImporter.canonicalize()`.
* `FilesystemImporter.canonicalize()` returns `file://.../node_modules/foo/scss/mixins/_banner.scss`.
* `_banner.scss` is loaded and everything is great.

js-api-spec/node-package-importer.node.test.ts
```diff
@@ -0,0 +1,580 @@
...
+              importers: [nodePackageImporter],
+            })
+          ).toThrowSassException({
+            includes: "Can't find stylesheet to import",
+          });
+        });
+      });
+    });
+  });
+  it('faked Node Package Importer fails', () =>
```
nex3, line 422: Empty line between test cases.

js-api-spec/node-package-importer.node.test.ts
nex3: I think a better way to do it would be to pass in a custom importer and verifying that that importer is passed a `pkg:` URL when the Node importer can't find anything.

js-api-spec/node-package-importer.node.test.ts
nex3: I'd prefer to remove everything but the `name` and the fields that are actually used, I think. Otherwise a reader has to do too much digging to figure out what's actually plausibly relevant.

nex3 (changes requested):

js-api-spec/node-package-importer.node.test.ts
jamesnw: Removed in 551b5b758c7e595b89635fba372dd013af447e97

nex3 (changes requested):

nex3 (approved):

jgerigmeyer (approved):

js-api-spec/sandbox.ts
jamesnw, line 1: @nex3 I needed to make this change after merging Shared Resources into `node-embedded-host`, as all async tests were failing. I think this is correct now (the callback can be async, so we need to await it, so that the `finally` doesn't happen before the callback resolves). 

However, it seemed worth flagging with you because the async `dart-sass` tests did not fail, and the `node-embedded-host` tests did not fail until after merging in Shared Resources, and we weren't able to track down any cause for this difference.

ntkme: It could be that functions with `FutureOr<T>` return type in dart is not really an async function in JS, but a sync function that may return a Promise or T. This allows the some synchronous code path to be dispatched synchronously, therefore may lead to different execution order than how a true asynchronous function is dispatched.

js-api-spec/importer.test.ts
```diff
@@ -763,6 +764,16 @@ it('throws an ArgumentError when the result sourceMapUrl is missing a scheme', (
   });
 });
 
+runOnlyForImpl('browser', () => {
+  it('node package loader throws error in browser', () => {
+    expect(() =>
+      compileString('@use "pkg:foo";', {
+        importers: [new NodePackageImporter()],
+      })
```
nex3, line 772: This should just test `new NodePackageImporter()` without passing it to `compileString()` now.

nex3 (approved):

js-api-spec/importer.test.ts
jgerigmeyer: Addressed in https://github.com/oddbird/sass-spec/commit/b18ff2ca5bb3dfbdc33928729ba3abda226cd553

nex3 (approved):

PR #1954, opened 2023-11-14
[Shared Resources] Add JS API tests

jerivas:
[Issue](https://github.com/sass/sass/issues/3296)

Blocked until [proposal](https://github.com/sass/sass/blob/main/proposal/shared-resources.d.ts.md) is accepted.

[skip dart-sass]
[skip sass-embedded]

- https://github.com/sass/sass/pull/3744
- https://github.com/sass/dart-sass/pull/2134
- https://github.com/sass/embedded-host-node/pull/261

### DIFF
--- a/js-api-spec/compiler.node.test.ts
+++ b/js-api-spec/compiler.node.test.ts
@@ -1,4 +1,4 @@
-// Copyright 2023 Google Inc. Use of this source code is governed by an
+// Copyright 2024 Google Inc. Use of this source code is governed by an
 // MIT-style license that can be found in the LICENSE file or at
 // https://opensource.org/licenses/MIT.
 
--- a/js-api-spec/compiler.node.test.ts
+++ b/js-api-spec/compiler.node.test.ts
@@ -9,7 +9,7 @@ import {
   asyncImporters,
   functions,
   getLogger,
-  getSlowImporter,
+  getTriggeredImporter,
   importers,
 } from './compiler.test';
 import {sandbox} from './sandbox';
@@ -122,11 +122,17 @@ describe('AsyncCompiler', () => {
       sandbox(async dir => {
         let completed = false;
         dir.write({'input.scss': '@import "slow"'});
+        const {importer, triggerComplete} = getTriggeredImporter(
+          () => (completed = true)
+        );
         const compilation = compiler.compileAsync(dir('input.scss'), {
-          importers: [getSlowImporter(() => (completed = true))],
+          importers: [importer],
         });
+        const disposalPromise = compiler.dispose();
         expect(completed).toBeFalse();
-        await compiler.dispose();
+        triggerComplete();
+
+        await disposalPromise;
         expect(completed).toBeTrue();
         await expectAsync(compilation).toBeResolved();
       }));
--- a/js-api-spec/compiler.node.test.ts
+++ b/js-api-spec/compiler.node.test.ts
@@ -1,0 +1,142 @@
+// Copyright 2023 Google Inc. Use of this source code is governed by an
+// MIT-style license that can be found in the LICENSE file or at
+// https://opensource.org/licenses/MIT.
+
+import type {AsyncCompiler, Compiler, CompileResult} from 'sass';
+import {initAsyncCompiler, initCompiler} from 'sass';
+
+import {
+  asyncImporters,
+  functions,
+  getLogger,
+  importers,
+  sortCompiled,
+} from './compiler.test';
+import {sandbox} from './sandbox';
+import {URL} from './utils';
+
+describe('Compiler', () => {
+  let compiler: Compiler;
+
+  beforeEach(() => {
+    compiler = initCompiler();
+  });
+
+  afterEach(() => {
+    compiler.dispose();
+  });
+
+  describe('compile', () => {
+    it('performs complete compilations', () =>
+      sandbox(dir => {
+        const logger = getLogger();
+        dir.write({'input.scss': '@import "bar"; .fn {value: foo(bar)}'});
+        const result = compiler.compile(dir('input.scss'), {
+          importers,
+          functions,
+          logger,
+        });
+        expect(result.css).toEqualIgnoringWhitespace(
+          '.import {value: bar;} .fn {value: "bar";}'
+        );
+        expect(logger.debug).toHaveBeenCalledTimes(1);
+      }));
+
+    it('performs compilations in callbacks', () =>
+      sandbox(dir => {
+        dir.write({'input-nested.scss': 'x {y: z}'});
+        const nestedImporter = {
+          canonicalize: (url: string) => new URL(`u:${url}`),
+          load: (url: typeof URL) => ({
+            contents: compiler.compile(dir('input-nested.scss')).css,
+            syntax: 'scss' as const,
+          }),
+        };
+        dir.write({'input.scss': '@import "nested"; a {b: c}'});
+        const result = compiler.compile(dir('input.scss'), {
+          importers: [nestedImporter],
+        });
+        expect(result.css).toEqualIgnoringWhitespace('x {y: z;} a {b: c;}');
+      }));
+
+    it('throws after being disposed', () =>
+      sandbox(dir => {
+        dir.write({'input.scss': '$a: b; c {d: $a}'});
+        compiler.dispose();
+        expect(() => compiler.compile(dir('input.scss'))).toThrowError();
+      }));
+  });
+});
+
+describe('AsyncCompiler', () => {
+  let compiler: AsyncCompiler;
+  const runs = 1000; // Number of concurrent compilations to run
+
+  // A slow importer that executes a callback after a delay
+  const getSlowImporter = (callback: () => void) => ({
+    canonicalize: async (url: string) => new URL('foo:bar'),
+    load: async (url: typeof URL) => {
+      await new Promise(resolve => setTimeout(resolve, 100));
+      callback();
+      return {contents: '', syntax: 'scss' as const};
+    },
+  });
+
+  beforeEach(async () => {
+    compiler = await initAsyncCompiler();
+  });
+
+  afterEach(async () => {
+    await compiler.dispose();
+  });
+
+  describe('compileAsync', () => {
+    it('handles multiple concurrent compilations', () =>
+      sandbox(async dir => {
+        const logger = getLogger();
+        const compilations = Array(runs)
+          .fill(0)
+          .map((_, i) => {
+            const filename = `input-${i}.scss`;
+            dir.write({
+              [filename]: `@import "${i}"; .fn {value: foo(${i})}`,
+            });
+            return compiler.compileAsync(dir(filename), {
+              importers: asyncImporters,
+              functions,
+              logger,
+            });
+          });
+        Array.from(await Promise.all(compilations))
+          .map((result: CompileResult) => result.css)
+          .sort(sortCompiled)
+          .forEach((result, i) => {
+            expect(result).toEqualIgnoringWhitespace(
+              `.import {value: ${i};} .fn {value: "${i}";}`
+            );
+          });
+        expect(logger.debug).toHaveBeenCalledTimes(runs);
+      }));
+
+    it('throws after being disposed', async () => {
+      sandbox(async dir => {
+        dir.write({'input.scss': '$a: b; c {d: $a}'});
+        await compiler.dispose();
+        expect(() => compiler.compileAsync(dir('input.scss'))).toThrowError();
+      });
+    });
+
+    it('waits for compilations to finish before disposing', () =>
+      sandbox(async dir => {
+        let completed = false;
+        dir.write({'input.scss': '@import "slow"'});
+        const compilation = compiler.compileAsync(dir('input.scss'), {
+          importers: [getSlowImporter(() => (completed = true))],
+        });
+        expect(completed).toBeFalse();
+        await compiler.dispose();
+        expect(completed).toBeTrue();
+        await expectAsync(compilation).toBeResolved();
+      }));
+  });
+});
--- a/js-api-spec/compiler.node.test.ts
+++ b/js-api-spec/compiler.node.test.ts
@@ -9,8 +9,8 @@ import {
   asyncImporters,
   functions,
   getLogger,
+  getSlowImporter,
   importers,
-  sortCompiled,
 } from './compiler.test';
 import {sandbox} from './sandbox';
 import {URL} from './utils';
@@ -72,16 +72,6 @@ describe('AsyncCompiler', () => {
   let compiler: AsyncCompiler;
   const runs = 1000; // Number of concurrent compilations to run
 
-  // A slow importer that executes a callback after a delay
-  const getSlowImporter = (callback: () => void) => ({
-    canonicalize: async (url: string) => new URL('foo:bar'),
-    load: async (url: typeof URL) => {
-      await new Promise(resolve => setTimeout(resolve, 100));
-      callback();
-      return {contents: '', syntax: 'scss' as const};
-    },
-  });
-
   beforeEach(async () => {
     compiler = await initAsyncCompiler();
   });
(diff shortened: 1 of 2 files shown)


jerivas:
[I want to share resources across multiple invocations of the Sass compiler's JS API.](https://trello.com/c/BT4VgchI/4-i-want-to-share-resources-across-multiple-invocations-of-the-sass-compilers-js-api)

js-api-spec/compiler.test.ts
```diff
@@ -0,0 +1,106 @@
...
+      );
+      expect((await compiler.compileStringAsync('$a: 1; c {d: $a}')).css).toBe(
+        'c {\n  d: 1;\n}'
+      );
+    });
+    it('throws after being disposed', async () => {
+      await compiler.dispose();
+      await expectAsync(
+        compiler.compileStringAsync('$a: b; c {d: $a}')
+      ).toThrowSassException({line: 0});
```
jerivas, line 67: Even though I'm expecting the throw here, the tests are still failing with:

```
AsyncCompiler compileStringAsync throws after being disposed
  - Error: Compiler caused error: Async compiler has already been disposed..
```

Am I doing something wrong here? https://github.com/oddbird/embedded-host-node/blob/ebfddd959e77c98a0a8ee347a40efb24ee2a24e2/lib/src/async-compiler.ts#L62-L66

js-api-spec/compiler.test.ts
```diff
@@ -0,0 +1,106 @@
...
+        );
+      });
+    });
+    it('throws after being disposed', async () => {
+      await sandbox(async dir => {
+        dir.write({'input.scss': '$a: b; c {d: $a}'});
+        await compiler.dispose();
+        await expectAsync(
+          compiler.compileAsync(dir('input.scss'))
+        ).toBeRejected();
```
jerivas, line 94: I also tried the more generic `toBeRejected()`, but the test still fails:

```
AsyncCompiler compileAsync throws after being disposed
  - Error: Compiler caused error: Async compiler has already been disposed..
```

jerivas (commented):
I added some basic tests to verify the behavior described in the proposal. I think we should also run all the tests from [compile.node.test.ts](https://github.com/oddbird/sass-spec/blob/e955667bccc6fd5bc2abc815644c537941cf305b/js-api-spec/compile.node.test.ts) and [compile.test.ts](https://github.com/oddbird/sass-spec/blob/e955667bccc6fd5bc2abc815644c537941cf305b/js-api-spec/compile.test.ts). What do you think of parametrizing these tests to accept both the standalone `compile*` functions as well as the compiler methods?

js-api-spec/compiler.test.ts
jerivas: I switched to a regular `expect().toThrow()` in https://github.com/sass/sass-spec/pull/1954/commits/6145b4e73263402227b91140c94d9067b257906f. That makes the test pass, but I'm curious if that's the correct behavior

jgerigmeyer: > I switched to a regular `expect().toThrow()` in [6145b4e](https://github.com/sass/sass-spec/commit/6145b4e73263402227b91140c94d9067b257906f). That makes the test pass, but I'm curious if that's the correct behavior

@jerivas Looking at similar tests in `compile.test.ts`, yes I think this is the correct behavior. I updated them to use `toThrowError`, but that's minor.

nex3 (approved):

nex3 (changes requested):
Actually, thinking about this more, it's probably a good idea to verify that the async compiler in particular gracefully handles multiple concurrent compilations, especially when those compilations are "chatty" (calling custom functions, importers, and loggers). Probably just one integration test case is sufficient as long as it handles a large number of chatty compilations at once and verifies that the right compilation calls get the right events.

js-api-spec/compiler.test.ts
```diff
@@ -0,0 +1,125 @@
...
+  });
+
+  afterEach(() => {
+    compiler.dispose();
+  });
+
+  describe('compileString', () => {
+    it('performs multiple compilations', () => {
+      expect(compiler.compileString('$a: b; c {d: $a}').css).toBe(
+        'c {\n  d: b;\n}'
```
jamesnw, line 24: You can also use `toEqualIgnoringWhitespace` so you don't need to worry about the line breaks, for instance- `toEqualIgnoringWhitespace('c {d: b;}')`

js-api-spec/compiler.test.ts
```diff
@@ -0,0 +1,125 @@
...
+    it('waits for compilations to finish before disposing', async () => {
+      const compilation = compiler.compileStringAsync('$a: b; c {d: $a}');
+      await compiler.dispose();
+      await expectAsync(compilation).toBeResolved();
+    });
+  });
+
+  describe('compileAsync', () => {
+    it('performs multiple compilations', async () => {
+      await sandbox(async dir => {
```
jamesnw, line 96: This call back can be slightly simplified by returning sandbox directly from the arrow function instead of wrapping it in an async function-

```
it('performs multiple compilations', () => 
  sandbox(async dir => {
```

(Side note, if the promise chain isn't quite right with the sandbox, I found it was very easy to have false positives. All of these look good, but I got into the habit of changing an expected output to verify that the test can actually break)

jerivas: Unfortunately, If I don't await for sandbox I get:

```
**************************************************
*                    Failures                    *
**************************************************

1) AsyncCompiler compileAsync performs multiple compilations
  - Unhandled promise rejection: Error: Compiler caused error: Async compiler has already been disposed..

2) Non-spec failure
  - Spec "AsyncCompiler compileAsync waits for compilations to finish before disposing" ran a "toBeResolved" expectation after it finished.
  1. Did you forget to return or await the result of expectAsync?
  2. Was done() invoked before an async operation completed?
  3. Did an expectation follow a call to done()?
```

js-api-spec/compiler.test.ts
jerivas: Thanks! Fixed in https://github.com/sass/sass-spec/pull/1954/commits/89ef7e5775b6a89e13e36c03f253e3bef169ccc8

js-api-spec/compiler.test.ts
jerivas: Fixed in 95fd55894208bcd297024c2ba99fafb91cfec8f0

js-api-spec/compiler.test.ts
```diff
@@ -0,0 +1,176 @@
...
+    it('handles multiple concurrent compilations', async () => {
+      const logger = getLogger();
+      const compilations = Array(5)
+        .fill(0)
+        .map((_, i) =>
+          compiler.compileStringAsync(
+            `@import "${i}"; .fn {value: foo(${i})}`,
+            {importers: asyncImporters, functions, logger}
+          )
+        );
```
jerivas, line 106: > verify that the async compiler in particular gracefully handles multiple concurrent compilations, especially when those compilations are "chatty" (calling custom functions, importers, and loggers)

@nex3 this will fire up 5 "chatty" compilations. Let me know what you think

js-api-spec/compiler.test.ts
```diff
@@ -0,0 +1,176 @@
...
+    compiler.dispose();
+  });
+
+  describe('compileString', () => {
+    it('performs complete compilations', () => {
+      const logger = getLogger();
+      const result = compiler.compileString(
+        '@import "bar"; .fn {value: foo(baz)}',
+        {importers, functions, logger}
+      );
```
jerivas, line 47: > we should also verify that at least two concurrent compilations work for the sync compiler, since it's theoretically possible to start a second one in a callback from the first.

I'm not sure I 100% understand the request, but this is my attempt to do it by using importers and functions

js-api-spec/compiler.test.ts
nex3: I think we should probably run more like 1000 of them. We really want to pressure-test the concurrency here.

js-api-spec/compiler.test.ts
nex3: I want to testing running multiple synchronous compilations at the same time. This means you'll need to start a second compilation _within a callback_ for the first one.

js-api-spec/compiler.test.ts
jerivas: Updated to 1000 in https://github.com/sass/sass-spec/pull/1954/commits/36db21b77999bb27cf7085cde9e7b88b8c5c0252. In local testing I was able to go to about 2000 before Jasmine's 5 second timeout kicked in

js-api-spec/compiler.test.ts
jerivas: Added in https://github.com/sass/sass-spec/pull/1954/commits/7c17ae557364320231b793c9296c004127de15a8

js-api-spec/compiler.test.ts
jerivas: It seems like all Mac OS runs (even on other branches) are taking around double the time to complete when compared to Ubuntu. @nex3 would you be open to increasing the timeout for this test only on macs?

js-api-spec/compiler.node.test.ts
```diff
@@ -0,0 +1,142 @@
...
+            });
+            return compiler.compileAsync(dir(filename), {
+              importers: asyncImporters,
+              functions,
+              logger,
+            });
+          });
+        Array.from(await Promise.all(compilations))
+          .map((result: CompileResult) => result.css)
+          .sort(sortCompiled)
```
jamesnw, line 112: @jerivas What's the purpose of sorting here? I think that `Promise.all` will maintain the order, and sorting could hide if the first compilation is returning the results of the second compilation.

js-api-spec/compiler.node.test.ts
```diff
@@ -0,0 +1,142 @@
...
+      }));
+  });
+});
+
+describe('AsyncCompiler', () => {
+  let compiler: AsyncCompiler;
+  const runs = 1000; // Number of concurrent compilations to run
+
+  // A slow importer that executes a callback after a delay
+  const getSlowImporter = (callback: () => void) => ({
```
jamesnw, line 76: Could you reuse the getSlowImporter exported from `compiler.test.ts`

js-api-spec/compiler.node.test.ts
jerivas: Huh, I recall getting errors in the test at some point due to ordering issues, but you are correct; I just tried it and it's not needed. Thanks for catching that!

jerivas: Fixed in a2e6f0cd1f141522ad9a592953d9cd68931ad44f

js-api-spec/compiler.node.test.ts
jerivas: Whoops I even exported the function but didn't use it here 😅 Fixed in a2e6f0cd1f141522ad9a592953d9cd68931ad44f

js-api-spec/compiler.test.ts
jerivas: I increased the timeouts to 15 seconds for string tests and 40 seconds for node tests. See https://github.com/sass/sass-spec/pull/1954/commits/286e7fafd73d3428b8e51d85beb10a097d1bed18

js-api-spec/compiler.test.ts
```diff
@@ -0,0 +1,140 @@
...
+  },
+];
+
+export const getLogger = () => ({debug: spy(() => {})});
+
+/* A slow importer that executes a callback after a delay */
+export const getSlowImporter = (callback: () => void) => ({
+  canonicalize: async () => new URL('foo:bar'),
+  load: async () => {
+    await new Promise(resolve => setTimeout(resolve, 100));
```
nex3, line 38: Rather than using a timeout when testing that concurrent systems wait for a given event, it's usually better to have a promise whose resolution you control explicitly so you can guarantee it won't finish until you want it to (and then you don't end up writing tests that take a lot longer than they need to because of delays). You can run `await Promise(resolve => setTimeout(resolve, 0))` to force all outstanding _non_-time-based tasks to complete, then resolve the promise and verify that the behavior happens as expected.

js-api-spec/compiler.test.ts
```diff
@@ -0,0 +1,140 @@
...
+    canonicalize: (url: string) =>
+      Promise.resolve(importers[0].canonicalize(url)),
+    load: (url: typeof URL) => Promise.resolve(importers[0].load(url)),
+  },
+];
+
+export const getLogger = () => ({debug: spy(() => {})});
+
+/* A slow importer that executes a callback after a delay */
+export const getSlowImporter = (callback: () => void) => ({
```
nex3, line 35: Declare functions as `export function`, and don't forget to include a return type.

nex3 (changes requested):

js-api-spec/compiler.test.ts
jamesnw: Resolved in  6dcdd3824f575e495c25bd5e21d2d406453fe2f2

js-api-spec/compiler.test.ts
jamesnw: Resolved in  6dcdd3824f575e495c25bd5e21d2d406453fe2f2

js-api-spec/compiler.node.test.ts
```diff
@@ -0,0 +1,140 @@
...
+      }));
+
+    it('performs compilations in callbacks', () =>
+      sandbox(dir => {
+        dir.write({'input-nested.scss': 'x {y: z}'});
+        const nestedImporter = {
+          canonicalize: () => new URL('foo:bar'),
+          load: () => ({
+            contents: compiler.compile(dir('input-nested.scss')).css,
+            syntax: 'scss' as const,
```
nex3, line 52: Nit: Rather than `as const`, I think it's cleaner to add a type declaration to the variable itself (`const nestedImporter: Importer`) so that it infers the correct types for all its components. Similarly for the top-level fields in `compiler.test.ts`.

js-api-spec/compiler.node.test.ts
```diff
@@ -0,0 +1,140 @@
...
+        dir.write({'input.scss': '$a: b; c {d: $a}'});
+        compiler.dispose();
+        expect(() => compiler.compile(dir('input.scss'))).toThrowError();
+      }));
+  });
+});
+
+describe('AsyncCompiler', () => {
+  let compiler: AsyncCompiler;
+  const runs = 1000; // Number of concurrent compilations to run
```
nex3, line 73: Declare this in the test where it's used

nex3 (changes requested):

js-api-spec/compiler.node.test.ts
jamesnw: Moved in 0a14b6fe9c3b41bc7a7bc5bbf11710843648ad0b

js-api-spec/compiler.node.test.ts
jamesnw: Updated in 0a14b6fe9c3b41bc7a7bc5bbf11710843648ad0b.

js-api-spec/compiler.test.ts
```diff
@@ -0,0 +1,174 @@
...
+import {spy, URL} from './utils';
+
+export const functions = {
+  'foo($args)': (args: unknown) => new SassString(`${args}`),
+};
+
+export const importers: Array<Importer> = [
+  {
+    canonicalize: (url: string) => new URL(`u:${url}`),
+    load: (url: typeof URL) => ({
```
nex3, line 23: I think now that `importers` itself is typed you don't need to type the parameters here (or in `asyncImporters`).

js-api-spec/compiler.test.ts
```diff
@@ -0,0 +1,174 @@
...
+      });
+      expect(result.css).toEqualIgnoringWhitespace('x {y: z;} a {b: c;}');
+    });
+
+    it('throws after being disposed', () => {
+      compiler.dispose();
+      expect(() => compiler.compileString('$a: b; c {d: $a}')).toThrowError();
+    });
+  });
+  it('errors if constructor invoked directly', () => {
```
nex3, line 104: Blank line before test case

nex3 (changes requested):

js-api-spec/compiler.test.ts
jamesnw: Added in https://github.com/sass/sass-spec/pull/1954/commits/76e676aaddc313aa5a53835cd0c883f54e530314

js-api-spec/compiler.test.ts
jamesnw: Updated

nex3 (approved):

jgerigmeyer (approved):

PR #1957, opened 2023-12-14
Use lts/* and lts/-n for node version

ntkme:
This PR helps us avoid manually updating node versions to test in the future.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -17,8 +17,9 @@ jobs:
 
     steps:
       - uses: actions/checkout@v3
+      # TODO(jathak): Update this to "${{ env.NODE_VERSION }}" (the default lts version) once unit tests are fixed.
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -148,13 +144,15 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest, windows-latest, macos-latest]
-        node_version: [18]
+        node_version: ['lts/*']
```
nex3, line 147: It seems inconsistent that we're using `lts/*` directly here and `env.NODE_VERSION` elsewhere. I'd be fine with removing the variable entirely and using `lts/*` everywhere, but we should be consistent.

ntkme: Done.

nex3:
Thanks!

PR #1966, opened 2024-03-29
[Color 4] Update existing specs

nex3:
This covers changes from sass/sass#3819 and sass/sass#3823.

[skip dart-sass]
[skip sass-embedded]


spec/core_functions/color/adjust_color/hwb.hrx
```diff
@@ -73,7 +73,7 @@ a {b: adjust-color(#993333, $blackness: 100%)}
 
 <===> blackness/max/output.css
 a {
-  b: rgb(42.5, 42.5, 42.5);
+  b: rgb(31.875, 31.875, 31.875);
```
Goodwine, line 76: Why does the `rgb()` value change by more than 10 points on each channel? I don't think the `adjust-color` function changed in a way that would cause this much of a change. Is it intended? I'm not sure what I missed 😅

Goodwine (approved):

nex3:
I'll be adding tests for those as part of `color.to-space` and `color.change` tests, since as it stands there isn't a way with currently-tested functions to _make_ an out-of-gamut clamped color.

PR #1967, opened 2024-04-01
Add tests for the deprecations API

jathak:
See https://github.com/sass/sass/pull/3826

[skip dart-sass]
[skip sass-embedded]

### DIFF
--- a/js-api-spec/deprecations.test.ts
+++ b/js-api-spec/deprecations.test.ts
@@ -12,30 +12,76 @@ import {
 
 import {captureStdio, URL} from './utils';
 
-// TODO(jathak): Add tests for obsolete deprecations once any actually exist.
+/**
+ * Map from obsolete deprecation IDs to version pairs.
+ *
+ * The first version is the version this deprecation type was deprecated in,
+ * while the second version is the version it was made obsolete in.
+ */
+const obsoleteDeprecations: {[key: string]: [string, string]} = {};
 
-describe('deprecation types', () => {
-  const activeDeprecations = {
-    'call-string': '0.0.0',
-    elseif: '1.3.2',
-    'moz-document': '1.7.2',
-    'relative-canonical': '1.14.2',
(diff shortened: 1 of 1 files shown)


nex3 (approved):

js-api-spec/deprecations.test.ts
```diff
@@ -0,0 +1,254 @@
...
+    it(`${id} is a future deprecation`, () => {
+      const deprecation = deprecationsMap[id];
+      expect(deprecation?.id).toBe(id);
+      expect(deprecation?.status).toBe('future');
+    });
+  }
+});
+
+describe('deprecation flags', () => {
+  it('warning emitted with no flags', done => {
```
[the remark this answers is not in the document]
jathak: Done

js-api-spec/deprecations.test.ts
```diff
@@ -0,0 +1,308 @@
...
+  const expectedDeprecations = [
+    ...Object.keys(obsoleteDeprecations),
+    ...Object.keys(activeDeprecations),
+    ...futureDeprecations,
+    'user-authored',
+  ];
+  const extraDeprecations = Object.keys(deprecations).filter(
+    deprecation => !expectedDeprecations.includes(deprecation)
+  );
+  expect(extraDeprecations).toBeEmptyArray();
```
nex3, line 66: We should check that everything in `expectedDeprecations` is in `deprecations` as well. Also, we should _probably_ have some sort of test for the deprecation list in the JS API spec as well.

nex3 (changes requested):

js-api-spec/deprecations.test.ts
jathak: Added an additional check that there are no missing deprecations either.

I also added some types which should ensure that there aren't any extra deprecations in these tests that aren't in the spec.

nex3 (approved):

PR #1968, opened 2024-04-01
[Color 4] Add better tests for legacy color space constructor bounds

nex3:


nex3:
Test failures should be fixed by https://github.com/sass/dart-sass/pull/2214 and https://github.com/sass/sass-spec/pull/1973

Goodwine (approved):

PR #1971, opened 2024-04-09
Add the importer-without-url deprecation

### LINKED ISSUE #1970: Add tests for base importer without URL
~~We can't test this deprecation directly because it's against the TypeScript types anyway, but we do need to update the deprecation list.~~ Although this behavior has always worked for JS users, we haven't been able to test it because it's against the TypeScript types.

nex3:
Closes #1970
See sass/sass#3832

[skip dart-sass]
[skip sass-embedded]


nex3:
Per https://github.com/sass/dart-sass/issues/2208#issuecomment-2046013595, we've decided not to deprecate this after all.

PR #1972, opened 2024-04-09
Add a test for an importer without a base URL

### LINKED ISSUE #1970: Add tests for base importer without URL

nex3:
Closes #1970
See sass/sass#3832


nex3:
CI should be fixed by https://github.com/sass/dart-sass/pull/2214 and https://github.com/sass/sass-spec/pull/1973

Goodwine (approved):

PR #1979, opened 2024-04-22
Add tests for `@keyframes` nesting errors

nex3:
[skip dart-sass]

### DIFF
--- a/spec/css/keyframes.hrx
+++ b/spec/css/keyframes.hrx
@@ -253,14 +253,14 @@ $a: b;
 ================================================================================
 <===> error/in_keyframe_block/style_rule/input.scss
 @keyframes a {
-  to {b {c: d}}
+  to {to {c: d}}
 }
 
 <===> error/in_keyframe_block/style_rule/error
 Error: Style rules may not be used within keyframe blocks.
   ,
-2 |   to {b {c: d}}
-  |       ^^^^^^^^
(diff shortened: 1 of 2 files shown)


spec/css/keyframes.hrx
```diff
@@ -248,3 +248,48 @@ $a: b;
...
+2 |   to {@b}
+  |       ^^
+  '
+  input.scss 2:7  root stylesheet
+
+<===>
+================================================================================
+<===> error/in_keyframe_block/known_at_rule/input.scss
+@keyframes a {
+  to {@media {b: c}}
```
Goodwine, line 286: It seems like the error here is that `@media` is missing a value like `print`. Please change this test to use `@media print` or other similar value. [Currently](https://sass-lang.com/playground/#eJwzNHTITq1MK0rMTS1WSFSo5lJQKMlXqHbITU3JTFQoKMrMK1GoTrJSSK6t5aoFAFWBDxY=) Sass throws the same error because of the missing value for `@media` even without the changes in https://github.com/sass/dart-sass/pull/2226

Goodwine (changes requested):

spec/css/keyframes.hrx
nex3: Ah, good catch!

Goodwine (approved):

PR #1984, opened 2024-05-03
[Color 4] Add tests for color equality with `same()` and `==`

nex3:
[skip dart-sass]


spec/values/colors/equality.hrx
```diff
@@ -0,0 +1,251 @@
+<===> true/legacy/same_space/rgb/no_none/input.scss
+a {b: #abcdef == #abcdef}
```
Goodwine, line 2: When Sass parses `#abc`, does it internally store it as if it parsed `#aabbcc`?

Currently `#abc == #aabbccff` returns `true`, per the spec changes this will now return `false`, or no?

And if it now returns `false`, is this covered by the `false/legacy/same_space/rgb/one_none` test?

spec/core_functions/color/same.hrx
```diff
@@ -0,0 +1,244 @@
+<===> true/same_space/identical/input.scss
+@use 'sass:color';
+a {b: color.same(#abcdef, #abcdef)}
```
Goodwine, line 3: Out of curiosity, is `color.same(#abc, #abcf)` covered by `true/same_space/none_and_zero` or by `true/different_space/no_none/input`?

Goodwine (commented):
LGTM, but I have a couple of questions due to my lack of understanding on how hex colors are parsed 😅

Goodwine (approved):
I forgot to select "Approve" 😅

spec/core_functions/color/same.hrx
nex3: `#abcf` isn't a color. Do you mean `#abcdef`? Those would be different colors, so they'd be under `false/`.

spec/values/colors/equality.hrx
nex3: All colors are stored as four floating-point numbers (the three color space channels and the alpha channel) as well as the color space name. `#abc`, `#aabbcc`, `#aabbccff`, and `rgb(170 187 255)` all share the same internal representation, and so are mutually `==` to one another. We test this in the tests for color parsing and constructor functions; here we can safely assume that the internal representations are the same, and that we only need to verify checks for that expected representation.

PR #1989, opened 2024-05-07
[Color 4] Update tests for `color.scale()`

nex3:


spec/core_functions/color/scale/a98_rgb.hrx
```diff
@@ -0,0 +1,225 @@
...
+
+<===>
+================================================================================
+<===> red/high/input.scss
+@use 'sass:color';
+a {b: color.scale(color(a98-rgb 0.2 0.5 0.7), $red: 86%)}
+
+<===> red/high/output.css
+a {
+  b: color(a98-rgb 0.888 0.5 0.7);
```
Goodwine, line 51: I see.. I thought `scale()` was a simple multiplication, but I get it now... It's moving to the point that is 86% of the way from the original value to the maximum value

Goodwine (approved):

spec/core_functions/color/scale/a98_rgb.hrx
nex3: Exactly :slightly_smiling_face:. The idea is to approximate the colloquial idea of "10% lighter" (or whatever) without blowing out colors that are already fairly light.

PR #1993, opened 2024-05-10
Add relative color syntax tests for the color() function

nex3:


spec/core_functions/color/color/relative_color.hrx
```diff
@@ -0,0 +1,88 @@
...
+}
+
+<===>
+================================================================================
+<===> calc/alpha/input.scss
+a {b: color(from #aaa srgb calc(l + 0.2) a b / 25%)}
+
+<===> calc/alpha/output.css
+a {
+  b: color(from #aaa srgb calc(l + 0.2) a b/25%);
```
Goodwine, line 26: is `calc()` not evaluated to `1.2` because it's considered to be a string in a slash-separated list?

Goodwine (approved):

spec/core_functions/color/color/relative_color.hrx
nex3: `calc()` isn't evaluated to `1.2` because the first character is U+006C LATIN SMALL LETTER L and not U+0031 DIGIT ONE :slightly_smiling_face:.

PR #1996, opened 2024-05-21
Update deprecations tests to use YAML list

jathak:
See https://github.com/sass/sass/pull/3872.

### DIFF
--- a/js-api-spec/deprecations.node.test.ts
+++ b/js-api-spec/deprecations.node.test.ts
@@ -1,0 +1,65 @@
+// Copyright 2024 Google LLC. Use of this source code is governed by an
+// MIT-style license that can be found in the LICENSE file or at
+// https://opensource.org/licenses/MIT.
+
+import fs from 'fs';
+import yaml from 'js-yaml';
+import {deprecations, Deprecation, Version} from 'sass';
+
+describe('deprecation type', () => {
+  const deprecationsMap = deprecations as unknown as {
+    [key: string]: Deprecation;
+  };
+  const obsoleteDeprecations: {[key: string]: [string, string]} = {};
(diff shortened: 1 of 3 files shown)


js-api-spec/deprecations.node.test.ts
```diff
@@ -0,0 +1,65 @@
...
+import yaml from 'js-yaml';
+import {deprecations, Deprecation, Version} from 'sass';
+
+describe('deprecation type', () => {
+  const deprecationsMap = deprecations as unknown as {
+    [key: string]: Deprecation;
+  };
+  const obsoleteDeprecations: {[key: string]: [string, string]} = {};
+  const activeDeprecations: {[key: string]: string} = {};
+  const futureDeprecations: {[key: string]: true} = {};
```
nex3, line 15: It's probably cleaner to use a `Set` for this.

js-api-spec/deprecations.node.test.ts
```diff
@@ -0,0 +1,65 @@
...
+  }
+
+  for (const [id, versions] of Object.entries(obsoleteDeprecations)) {
+    if (!versions) continue;
+    const [deprecatedIn, obsoleteIn] = versions;
+    it(`${id} deprecated in ${deprecatedIn} and obsolete in ${obsoleteIn}`, () => {
+      const deprecation = deprecationsMap[id];
+      expect(deprecation?.id).toBe(id);
+      expect(deprecation?.status).toBe('obsolete');
+      expect(deprecation?.deprecatedIn).toEqual(Version.parse(deprecatedIn));
```
nex3, line 44: Might be worth adding a comment here to the effect that if we ever have a JS API backed by a non-Dart-Sass implementation, it may have different versions for these.

nex3 (approved):

PR #2000, opened 2024-06-17
Add tests for comments in more expression-like contexts

nex3:
See sass/dart-sass#2263

[skip dart-sass]

### DIFF
--- a/spec/css/functions/special.hrx
+++ b/spec/css/functions/special.hrx
@@ -25,3 +25,187 @@ a {b: -a-calc(#{0})}
 a {
   b: -a-calc(0);
 }
+
+<===>
+================================================================================
+<===> comment/calc/after_open_paren/loud/input.scss
+a {b: -a-calc(/**/ c)}
+
+<===> comment/calc/after_open_paren/loud/output.css
+a {
+  b: -a-calc(/**/ c);
+}
+
+<===>
+================================================================================
+<===> comment/calc/after_open_paren/silent/input.scss
+a {
+  b: -a-calc(//
+    c);
+}
+
+<===> comment/calc/after_open_paren/silent/output.css
+a {
+  b: -a-calc( c);
+}
+
+<===>
+================================================================================
+<===> comment/calc/before_close_paren/loud/input.scss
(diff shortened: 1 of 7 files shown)


ntkme:
I wonder if we should omit the whitespace in between silent comment and sibling tokens during output?

nex3:
It's not generally safe to omit that whitespace entirely—for example, `foo//\nbar` is very different than `foobar`. There are some cases where we could fold more aggressively than we do now, but that's generally true with newlines adjacent to trailing spaces, not just silent comments.

Goodwine (approved):

spec/css/functions/special.hrx
```diff
@@ -25,3 +25,187 @@ a {b: -a-calc(#{0})}
...
+<===> comment/progid/before_close_paren/loud/output.css
+a {
+  b: progid:c(d /**/);
+}
+
+<===>
+================================================================================
+<===> comment/progid/before_close_paren/silent/input.scss
+a {
+  b: progid:c(d //
```
Goodwine, line 204: If there are multiple lines with `//`, are they considered a single silent comment, or multiple silent comments?

For example, would this block:
```scss
a {
  b: progid:c(d // foo
                // bar
    );
}
```
result in this?:
```css
a {
  b: progid:c(d

    );
}
```
And if so, is that OK?

nex3: They are not considered a single comment, but the whitespace around them is still considered "adjacent" so you won't end up with multiple newlines.

ntkme:
> There are some cases where we could fold more aggressively than we do now

If we treat silent comment as a kind of whitespace when processing AST, I think this can potentially allow us to fold it more aggressively based on context, even under the current logic.

For example, if silent comment is treated as whitespace: 
`a { b: calc(c<whitespace><silent-comment><newline><whitespace>) }` can be parsed as `a { b: calc(c<whitespace>) }`, and this whitespace can be completely striped based on its context.

Not sure if there is any corner case, and I haven't looked at the implementation details so not sure how much work it is. If it's too much work we can just leave it as is.

PR #2003, opened 2024-07-01
Migrate most tests to the module system

jathak:
This migrates most tests that aren't explicitly testing `@import` or global built-in functions to instead use `@use` and the `sass:` modules.

This was done using a combination of a new `migrate.ts` script that can run the Sass migrator over a spec directory and some manual changes and refactorings. There are probably some tests that could have been migrated that this misses because there were some directories I avoided running the migrator over to avoid accidentally changing a test that's actually explicitly testing `@import`, but this should at least minimize additional migration and new warnings we need to add when we actually deprecate `@import` and global built-in functions.

The `spec/core_functions/module` directory has been replaced with `spec/core_functions/global`, which similarly contains minimal tests for the global versions of each built-in function now that their full functionality is tested using the module system versions.


spec/core_functions/color/darken.hrx
nex3, line 1: For functions like this that are just translated into other, more general functions, it probably makes sense to omit them from the new module system tests entirely and have a small global test to verify that the name still works.

nex3 (changes requested):

spec/core_functions/color/darken.hrx
jathak: Removed the detailed tests for the `color.adjust` shorthands. The basic tests for the global versions are already in `spec/core_functions/global/color` and `spec/core_functions/color/adjust_color/error/missing_globals` tests that the functions that were removed or changed names in the module system don't exist in the `color` module.

nex3 (approved):

PR #2004, opened 2024-07-17
Update specs for @import/global builtins deprecations

jathak:
See https://github.com/sass/sass/pull/3898
[skip dart-sass]
[skip sass-embedded]


nex3 (approved):

nex3:
Looks like this also needs `[skip sass-embedded]`

ntkme:
The following test in `importer.test.ts` hasn't been updated:

``` js
  it("is converted to a file: URL if it's an absolute Windows path", () => {
    const result = compileString('@import "C:/orange";', {
      importers: [
        {
          canonicalize(url: string) {
            expect(url).toEqual('file:///C:/orange');
            return new URL(`u:${url}`);
          },
          load: () => ({contents: 'a {b: c}', syntax: 'scss'}),
        },
      ],
    });
    expect(result.css).toBe('a {\n  b: c;\n}');
  });
```

What's weird is that this test works with `@import`, but breaks with `@use` on ruby host. Not sure what's the intended behavior here.

nex3:
That's intended to be import-only behavior. As you surmise, it's a legacy behavior from when we were much more wishy-washy about the distinction between paths and URLs.

nex3 (approved):

PR #2014, opened 2024-09-13
Update mixed-decl tests to reflect the improved implementation

nex3:
[skip dart-sass]


spec/css/style_rule.hrx
```diff
@@ -215,3 +167,217 @@ a {b: c //
...
   b: c;
 }
+
+<===>
+================================================================================
+<===> declaration/mixed/after_style_rule/higher_specificity/input.scss
+a {
+  b {c: d}
+  e: f;
+}
```
Goodwine, line 177: IIUC this is OK because `a` will never be `b`, so their relative order doesn't matter, right?

It would have been different if these were `.a` and `.b`. Is there a test for this case already?

Goodwine (approved):
lgtm, but only one question. I think it's covered by the `:where()` test, but just trying to make sure

spec/css/style_rule.hrx
nex3: Good question. This actually doesn't take into account whether the selectors can overlap; it's considered okay because `a b` has a different specificity than `a`, so it'll always resolve in specificity order rather than document order. But it would be better to use class selectors to make that explicit.

Goodwine: oh, it's due to specificity!

PR #2018, opened 2024-09-18
Update tests for fixing unifyCompound() and unifyComplex() ordering of pseudo class/element selectors

Goodwine:
See https://github.com/sass/dart-sass/issues/2335
See https://github.com/sass/dart-sass/pull/2350

[skip dart-sass]

### DIFF
--- a/spec/core_functions/selector/unify/compound.hrx
+++ b/spec/core_functions/selector/unify/compound.hrx
@@ -94,3 +94,66 @@ a {b: selector.unify(":c", "::d")}
 a {
   b: :c::d;
 }
+
+<===>
+================================================================================
+<===> order/do_not_cross_pseudo_element/pseudo_class_and_element/into_simple/input.scss
+@use "sass:selector";
+a {b: selector.unify(".x::scrollbar:horizontal", ".y")}
+
+<===> order/do_not_cross_pseudo_element/pseudo_class_and_element/into_simple/output.css
+a {
+  b: .x.y::scrollbar:horizontal;
+}
+
+<===>
+================================================================================
+<===> order/do_not_cross_pseudo_element/pseudo_class_and_element/into_pseudo_element/input.scss
+@use "sass:selector";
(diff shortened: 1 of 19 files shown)


nex3 (approved):

Goodwine:
I added some changes to the tests to catch more potentially erroneous behaviors (or maybe those behaviors are intended haha), PTAL

spec/core_functions/selector/unify/compound.hrx
```diff
@@ -94,3 +132,77 @@ a {b: selector.unify(":c", "::d")}
...
+<===> order/do_not_cross_pseudo_element/pseudo_element/into_pseudo_class_and_element/output.css
+a {
+  b: :foo::bar:baz;
+}
+
+<===>
+================================================================================
+<===> order/cross_pseudo_element/input.scss
+@use "sass:selector";
+a {b: selector.unify(".x", "::y.z")}
```
nex3, line 203: I think we shouldn't test this case, because it's functionally meaningless to begin with.

Goodwine: ok, then I think it would make sense to remove the tests where on argument contains two pseudo selectors like `::x::y` since those don't work either IIUC. Let me know if I should add them back tho

spec/core_functions/selector/unify/compound.hrx
```diff
@@ -20,15 +20,37 @@ a {
...
+<===>
+================================================================================
+<===> full_overlap/pseudo/different_order/input.scss
+@use "sass:selector";
+a {b: selector.unify("::c.d", ".d::c")}
+
+<===> full_overlap/pseudo/different_order/output.css
+a {
+  b: ::c.d;
+}
```
nex3, line 52: Both of these involve meaningless input selectors, so are probably not worth testing. Unifying `.c::d` with itself is worth testing, though, as well as cases with pseudo-selectors

nex3 (approved):

PR #2023, opened 2024-09-24
Add tests for parsing missing channels from Sass

### LINKED ISSUE #2022: Test passing missing channels through the embedded host
See https://github.com/sass/sass/issues/3950

nex3:
See https://github.com/sass/sass/pull/3951
Closes #2022

[skip dart-sass]
[skip sass-embedded]


ntkme:
Looks like I misread the protobuf doc - it says when serializing proto object to JSON, NaN and infinity are not supported.

PR #2030, opened 2024-10-17
Update specs for an emergency workaround for sass/dart-sass#2390

nex3:
These are not intended to reflect the correct behavior, but rather the
fact that we need to limit the amount of repetition univerally until
we have time to find a proper solution to the underlying bug.

[skip dart-sass]


nex3:
I'm planning to land this TBR and cut a release tonight so we don't flood users with deprecation warnings for `@import`.

ntkme:
I suppose the dart-sass code change should only impact `@import` or related syntax, but I'm seeing this PR having some differences in seemingly unrelated deprecations e.g. `strict-unary`.

nex3:
That's expected; the original refactor affected how _all_ parse-time deprecations were handled, so this workaround affects them all as well.

ntkme (approved):

jathak (commented):
LGTM

PR #2034, opened 2024-10-24
Add tests for the indented syntax

jamesnw:
This adds minimal coverage to the indented syntax, with focus on the areas that may change with upcoming improvements, specifically line breaks and white space parsing.


nex3 (commented):
Please take a look at the [style guide](https://github.com/sass/sass-spec/blob/main/STYLE_GUIDE.md) for HRX tests, in particular the sections on irrelevant placeholders and testing one thing per spec. I think a fair number of these are more complex than they need to be in an attempt to test realistic CSS, but here terse is better than realistic.

I recommend adding some error tests as well to cover cases that are invalid, especially ones that may differ in the new system.

jamesnw:
@nex3 Do you have a preference between adding tests to a new `indented-syntax` folder vs. adding .sass variations of tests to files like [sass-spec/spec/directives/for.hrx](https://github.com/sass/sass-spec/blob/main/spec/directives/for.hrx)?

jamesnw:
@nex3 I've made changes per the style guide, and merged in with the existing file structure. I also found https://github.com/sass/sass-spec/tree/main/spec/non_conformant/sass and wasn't sure if it would make sense to have a `syntax/sass` folder to hold some of these, and maybe some of the non_conformant ones?

spec/css/custom_properties/trailing_whitespace.hrx
```diff
@@ -7,10 +7,28 @@
...
 .trailing-whitespace {
   --space: value ;
   --tab: value	;
   --newline: value ;
   --before-closing-brace: value ;
 }
+
+<===>
+================================================================================
+<===> sass/input.sass
```
nex3, line 20: If we're adding new tests, we should make them style-guide-conformant even if they're largely copies of old tests. Bonus points to make the old tests conformant as well.

spec/arguments/invocation.hrx
```diff
@@ -55,3 +55,57 @@ Error: Positional arguments must come before keyword arguments.
   |              ^
   '
   input.scss 3:14  root stylesheet
+
+<===>
+================================================================================
+<===> function/error/multi-line-arguments/declaration/input.sass
```
nex3, line 61: Anything that's specifically testing the indented syntax should probably have `/sass/` somewhere in the path.

nex3 (changes requested):
You can generally ignore the non-conformant tests, although if you notice that you're writing tests that cover all the same cases go ahead and delete the redundant non-conformant tests.

spec/directives/if/indented.hrx
nex3, line 1: Rather than adding `sass/` at the beginning of these HRX entries, rename the file to `sass` to match the other directories. (The logical directory name for each test is the physical name followed by the HRX name, so `indented/sass` is redundant.)

nex3 (approved):

spec/css/custom_properties/trailing_whitespace.hrx
```diff
@@ -1,16 +1,101 @@
...
+
+<===>
+================================================================================
+<===> sass/tab/input.sass
+a
+  --b: c	
+
+<===> sass/tab/output.css
+a {
+  --b: c;
```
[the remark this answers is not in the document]
nex3: There's a design conflict here between "trailing whitespace in custom property values is observable in CSS" and "end-of-line trailing whitespaces is invisible and generally shouldn't matter", complicated further by the fact that custom property trailing whitespace *nearly* never matters in practice. I think given that we're planning to allow `--b: c ;` in the indented syntax eventually anyway, preserving the existing behavior is probably the right way forward.

PR #2037, opened 2024-11-13
Add tests for CSS `round()` with one argument

### LINKED ISSUE #2036: Add tests for `round()` with no modulus
See https://github.com/sass/sass/issues/3803

nex3:
This also updates tests expect global function warnings when
calculations are falling back on global Sass functions.

See sass/sass#3803
Closes #2036

[skip dart-sass]

### DIFF
--- a/spec/values/calculation/abs.hrx
+++ b/spec/values/calculation/abs.hrx
@@ -64,6 +64,17 @@ a {
   b: 2px;
 }
 
+<===> preserves_single_unit/warning
+DEPRECATION WARNING: In future versions of Sass, abs() will be interpreted as the CSS abs() calculation. This doesn't allow unitless numbers to be mixed with numbers with units. If you want to use the Sass function, call math.abs() instead.
+
+See https://sass-lang.com/d/import
+
+  ,
+1 | a {b: abs(1 + 1px)}
+  |           ^^^^^^^
+  '
+    input.scss 1:11  root stylesheet
+
 <===>
 ================================================================================
 <===> sass_script/input.scss
--- a/spec/values/calculation/round/one_argument.hrx
+++ b/spec/values/calculation/round/one_argument.hrx
@@ -42,6 +42,17 @@ a {
   b: -2px;
 }
 
+<===> preserves_units/warning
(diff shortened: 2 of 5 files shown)


spec/values/calculation/abs.hrx
```diff
@@ -64,6 +64,17 @@ a {
...
 }
 
+<===> preserves_single_unit/warning
+DEPRECATION WARNING: In future versions of Sass, abs() will be interpreted as the CSS abs() calculation. This doesn't allow unitless numbers to be mixed with numbers with units. If you want to use the Sass function, call math.abs() instead.
+
+See https://sass-lang.com/d/import
+
+  ,
+1 | a {b: abs(1 + 1px)}
+  |           ^^^^^^^
```
pamelalozano16, line 74: Nit: Shouldn't these point to the function name instead?

spec/values/calculation/round/one_argument.hrx
```diff
@@ -52,6 +63,27 @@ a {
...
+
+See https://sass-lang.com/d/import
+
+  ,
+1 | a {b: round(1 + 1px)}
+  |             ^^^^^^^
+  '
+    input.scss 1:13  root stylesheet
+
+DEPRECATION WARNING: In future versions of Sass, round() will be interpreted as a CSS round() calculation. This requires an explicit modulus when rounding numbers with units. If you want to use the Sass function, call math.round() instead.
```
pamelalozano16, line 77: Is a double warning expected here?

pamelalozano16 (approved):
Lgtm

spec/values/calculation/abs.hrx
nex3: Since the deprecation warning specifically refers to mixing unitless and unit numbers, I think it makes sense to point to the specific location where that mixing occurs.

spec/values/calculation/round/one_argument.hrx
nex3: More or less. It's both the case that `round(1 + 1px)` will throw an error *and* that `round(2px)` (which it resolves to) will throw an error.

PR #2041, opened 2024-12-16
[Indented syntax improvements] Specs

jamesnw:
Blocked until [proposal](https://github.com/sass/sass/blob/main/proposal/indented-syntax-improvements.md) is accepted.

- [ ] dart-sass https://github.com/sass/dart-sass/pull/2467

[skip dart-sass]
[skip sass-embedded]


spec/directives/function/whitespace.hrx
```diff
@@ -0,0 +1,172 @@
...
+<===> return/before_value/sass/input.sass
+@function a()
+  @return
+    b
+
+<===> return/before_value/sass/output.css
+
+<===>
+================================================================================
+<===> nested_at_rule/debug/scss/input.scss
```
nex3, line 42: I'm not sure there's a point in testing multiple different nested at rules. That just seems like combinatorially many tests for little value.

spec/directives/in_declarations.hrx
nex3, line 1: I'm not sure why these need their own distinct whitespace tests

spec/directives/return.hrx
nex3, line 1: Doesn't this need whitespace tests?

spec/directives/while.hrx
nex3, line 1: This directory name doesn't include "whitespace" anywhere.

spec/expressions/comments.hrx
```diff
@@ -0,0 +1,156 @@
+<===> loud/sass/inline-closed/input.sass
```
nex3, line 1: Why use a hyphen here instead of a directory separator?

spec/expressions/comments.hrx
```diff
@@ -0,0 +1,156 @@
...
+================================================================================
+<===> silent/sass/indented-interpolation/input.sass
+// #{a 
+  + b} 
+
+<===> silent/sass/indented-interpolation/output.css
+
+<===>
+================================================================================
+<===> whitespace-comment/sass/inline/input.sass
```
nex3, line 101: I think "comment" here is redundant. If you're trying to emphasize that the comment is being parsed as a whitespace, I'd call this `as_whitespace`.

spec/expressions/variable.hrx
nex3, line 1: A semicolon doesn't make sense for a variable *expression*.

spec/operators/boolean.hrx
nex3, line 1: I'm concerned that these tests have no output verification, which means they could compile for unexpected reasons with unexpected semantics and still pass.

spec/operators/boolean.hrx
```diff
@@ -0,0 +1,75 @@
...
+$a: b not 
+  c
+
+<===> whitespace/not/after/sass/output.css
+
+<===>
+================================================================================
+<===> whitespace/not/before/sass/input.sass
+$a: b 
+not c
```
nex3, line 36: Unary operators shouldn't be tested in the same way as binary operators, since they aren't parsed in the same way.

spec/operators/boolean.hrx
nex3, line 1: I don't think you need to test whitespace separately for each separate operator type. All binary operators are parsed essentially the same.

spec/parser/interpolation.hrx
nex3, line 1: I don't think you need to test this separately for every place an interpolation can appear.

spec/parser/selector.hrx
```diff
@@ -48,6 +48,27 @@ b {
   c: d;
 }
 
+<===>
+================================================================================
+<===> multiline-no-comma/input.sass
```
nex3, line 53: This should probably go under `whitespace/`.

spec/parser/selector.hrx
nex3, line 1: What about cases like `a\n, b` and `a +\nb`?

spec/parser/semicolon.hrx
nex3, line 1: Aren't these redundant with the tests you have in individual statement directories?

nex3 (changes requested):
Other expressions that should probably have whitespace tests include parentheses and function calls.

jamesnw (commented):
@nex3 I think I resolved all the issues.

> Other expressions that should probably have whitespace tests include parentheses and function calls.

For parentheses, I do have tests in spec/values/maps/sass.hrx and spec/values/lists/sass.hrx, and also expanded tests in functions and parentheses in 271999543df62b56589f07437689a374e32b0651. 

This also caught a bug in parsing `@function a($b: c\nd)`. I fixed this and changed `expressionUntilComma` to always consume whitespace (because it is currently only called in places that can consume whitespace) here-
https://github.com/sass/dart-sass/pull/2467/commits/e1df2ad69f4f4c93cecc1bbe2f2a8e6c12721b0f

spec/core_functions/global/newlines.hrx
nex3, line 1: The `core_functions` directory is for testing the semantics of functions. Function call syntax doesn't differ between built-in, user-defined, and plain-CSS functions (except that the latter allows interpolation and not namepsaces) so syntactic tests for functions should go in a single, syntax-oriented location.

spec/css/media/whitespace.hrx
```diff
@@ -18,6 +18,42 @@ Error: Expected identifier.
   '
   input.sass 1:7  root stylesheet
 
+<===>
+================================================================================
+<===> error/logic_sequence/before_operator/sass/input.sass
```
nex3, line 23: The error directory should go at the end of the HRX file.

spec/directives/extend/whitespace.hrx
```diff
@@ -49,3 +49,22 @@ Error: Expected newline.
   |            ^
   '
   input.sass 2:12  root stylesheet
+
+<===>
+================================================================================
+<===> error/after_arg/sass/input.sass
```
nex3, line 55: Also test `@extend a\nb`.

spec/directives/mixin/sass.hrx
```diff
@@ -0,0 +1,5 @@
+<===> content/semicolon/input.sass
+@mixin a
+  @content;
```
nex3, line 3: Why a semicolon?

nex3 (commented):
Pretty close!

spec/core_functions/global/newlines.hrx
jamesnw: Moved to spec/expressions/functions in 5a5ff4562c6088ef998101b60150acca9bc1b5e2, and also switched to using `list.append` so I'm not using a deprecated function.

spec/directives/mixin/sass.hrx
jamesnw: This was porting over a nonconformant test, originally in [spec/non_conformant/sass/semicolon/content.hrx](https://github.com/sass/sass-spec/blob/main/spec/non_conformant/sass/semicolon/content.hrx#L5). Should it be removed?

spec/css/media/whitespace.hrx
jamesnw: Moved in cd4eced4e5e1125de252490046e5eff6d498da20

spec/directives/extend/whitespace.hrx
jamesnw: Tested in cd4eced4e5e1125de252490046e5eff6d498da20

spec/directives/extend/whitespace.hrx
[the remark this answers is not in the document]
nex3: This still doesn't have a test for `@extend a\nb`

nex3 (changes requested):

spec/directives/extend/whitespace.hrx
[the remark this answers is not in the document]
jamesnw: 🤦 Sorry. Added in 3af370809457f5eb8927fd78bfc69af2abd942bb.

nex3 (approved):

PR #2053, opened 2025-03-11
Disallow passing functions/mixins across compilations

ntkme:
https://github.com/sass/dart-sass/issues/2542
https://github.com/sass/dart-sass/pull/2544
https://github.com/sass/embedded-host-node/pull/367

### DIFF
--- a/js-api-spec/value/function.test.ts
+++ b/js-api-spec/value/function.test.ts
@@ -80,3 +80,45 @@ describe('rejects a function signature that', () => {
   it('has no closing parenthesis', () => rejectsSignature('foo('));
   it('has a non-identifier name', () => rejectsSignature('$foo()'));
 });
+
+it('rejects a compiler function from a different compilation', () => {
+  let plusOne: Value | undefined;
+  compileString(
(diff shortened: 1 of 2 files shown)


js-api-spec/value/function.test.ts
```diff
@@ -80,3 +80,45 @@ describe('rejects a function signature that', () => {
...
+
+it('rejects a compiler function from a different compilation', () => {
+  let plusOne: Value | undefined;
+  compileString(
+    `
+    @use 'sass:meta';
+
+    @function plusOne($n) {@return $n + 1}
+    a {b: meta.call(foo(meta.get-function('plusOne')), 2)}
+  `,
```
nex3, line 92: Style nit: Indent multiline strings 2 spaces beneath the opening quote, like so:
```suggestion
      @use 'sass:meta';

      @function plusOne($n) {@return $n + 1}
      a {b: meta.call(foo(meta.get-function('plusOne')), 2)}
    `,
```

nex3 (approved):

PR #2054, opened 2025-03-31
Update specs to treat "@elseif" as an unknown at-rule

nex3:
See https://github.com/sass/sass/pull/4057

[skip dart-sass]

### DIFF
--- a/spec/libsass-closed-issues/issue_1060.hrx
+++ b/spec/libsass-closed-issues/issue_1060.hrx
@@ -17,19 +17,10 @@ foo {
   }
 }
 
-<===> output.css
-foo {
-  foo: true;
-  bar: true;
-}
-
-<===> warning
-DEPRECATION WARNING [elseif]: @elseif is deprecated and will not be supported in future Sass versions.
-
-Recommendation: @else if
-
+<===> error
+Error: This at-rule is not allowed here.
   ,
-4 |   } @elseif true {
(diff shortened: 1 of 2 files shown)


spec/libsass-closed-issues/issue_1060.hrx
```diff
@@ -17,19 +17,10 @@ foo {
...
-DEPRECATION WARNING [elseif]: @elseif is deprecated and will not be supported in future Sass versions.
-
-Recommendation: @else if
-
+<===> error
+Error: This at-rule is not allowed here.
   ,
-4 |   } @elseif true {
-  |     ^^^^^^^
+6 |   } @else {
```
Goodwine, line 23: I find it strange that the error is pinned to the `@else` rather than the `@elseif` 🤔

Goodwine: I guess this is no different as having:
```scss
@if {}
@foo {}
@else {}
```

Goodwine (approved):

spec/libsass-closed-issues/issue_1060.hrx
nex3: With this change, `@elseif` isn't actually an error—it's just treated like any other unknown at-rule. So the only reason this fails to compile is that `@else` can only come after an `@if` rule, which it doesn't here.

PR #2069, opened 2025-08-13
Fix `ts-node` with `--no-experimental-strip-types`

ntkme:
`ts-node` is broken out of box on node >=22.18.0 due to `--experimental-strip-types` being enabled by default.

As we still need to test older node version without native typescript support, this PR sets `NODE_OPTIONS=--no-experimental-strip-types` if necessary.


ntkme:
@nex3 I spent a bit time reimplementing this as a wrapper script so that we don't have to update CI pipelines everywhere, and hopefully this will provide a more consistent experience for local users as well.

ntkme:
Looks like this only works on ubuntu/mac runners, but still have issue on windows runners when node version >=22.18.0. Let me check.

PR #2070, opened 2025-08-14
Fix `ts-node` wrapper on windows

ntkme:
@nex3 Here is the fix for the wrapper. The issue turns out to be the `shell: true` is stripping the last empty argument which is actually necessary...


ntkme:
The PR checks again ran on 22.17.1 rather than 22.18.0 on the main branch, but I tested this new patch on both version. It should be okay now.

PR #2087, opened 2025-09-20
Add tests for plain-CSS `@function`

nex3:
This also updates tests for `@mixin`s with leading `--`s to be errors.

[skip dart-sass]

### DIFF
--- a/spec/css/mixin.hrx
+++ b/spec/css/mixin.hrx
@@ -1,0 +1,12 @@
+<===> error/css/mixin/input.scss
+@mixin --a {}
+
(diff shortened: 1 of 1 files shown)


Goodwine (approved):

spec/css/function.hrx
```diff
@@ -0,0 +1,309 @@
+<===> lowercase/parameter/input.scss
+@function --a(--b <color>) {result: c}
```
Goodwine: Shouldn't there be a similar set of files for `@mixin`? 🤔

PR #2097, opened 2025-12-27
Enhance sourceMapIncludeSources option

ntkme:
https://github.com/sass/sass/pull/4170
https://github.com/sass/dart-sass/pull/2713
https://github.com/sass/embedded-host-node/pull/411


nex3 (changes requested):
This should test explicit `'auto'`, as well as testing the combination of each option with whether the importer returns an explicit `result.sourceMapUrl`.

PR #2100, opened 2026-01-01
Add sass-spec tests for deprecated case-insensitive function names

kantnero:
Tests for case-insensitive function name deprecation:

- Dart Sass emits warning: https://github.com/sass/dart-sass/pull/2716
- Deprecation entry: https://github.com/sass/sass/pull/4175
- Marked [skip dart-sass] so it only runs in Dart Sass, because only Dart Sass emits this deprecation warning.

### DIFF
--- a/spec/directives/function/name.hrx
+++ b/spec/directives/function/name.hrx
@@ -176,3 +176,36 @@ Error: This name is reserved for the plain-CSS function.
   |           ^^^^
   '
   input.scss 1:11  root stylesheet
+
+
+<===>
+================================================================================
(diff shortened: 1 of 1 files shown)


kantnero:
Tests for case-insensitive function name deprecation:

- Dart Sass emits warning: https://github.com/sass/dart-sass/pull/2716
- Deprecation entry: https://github.com/sass/sass/pull/4175
- Marked [skip dart-sass] so it only runs in Dart Sass, because only Dart Sass emits this deprecation warning.

PR #2111, opened 2026-02-10
Add regression test for https://github.com/sass/dart-sass/pull/2735

ntkme:
https://github.com/sass/dart-sass/pull/2735


js-api-spec/value/calculation.test.ts
```diff
@@ -227,6 +227,23 @@ describe('SassCalculation', () => {
       ).toBe('a {\n  b: 3;\n}');
     });
 
+    it('calc() with incompatible units', () => {
```
nex3: Can you add a comment linking to the issue (or comment in this case) this is a deprecation test for?

PR #2170, opened 2026-09-14
[meta.load] Add tests

nex3:
See https://github.com/sass/sass/issues/739
See https://github.com/sass/sass/pull/4247

[skip dart-sass]
[skip sass-embedded]


spec/values/modules.hrx
```diff
@@ -0,0 +1,109 @@
...
+Error: Undefined operation "get-module("meta") * get-module("meta")".
+  ,
+2 | $_: meta.get-module(meta) * meta.get-module(meta);
+  |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+  '
+  input.scss 2:5  root stylesheet
+
+<===>
+================================================================================
+<===> error/division/input.scss
```
jathak, line 43: Do we still want to bother with this test since slash division is deprecated and will be removed soon?

jathak (approved):

spec/values/modules.hrx
nex3: It'll be easy enough to remove when the time comes.

ntkme:
@nex3 I see this is merged as a non`-dev` version bump, but it's not released. Are we waiting on something?

nex3:
I just wanted to wait until after the Yom Kippur long weekend to get it out, so I'd have plenty of availability if anything went wrong.

